PR #270, opened 2014-11-28
give apple OpenAL include better default location

Sharnw:
Hey mate,

Had to make a fix to build with OpenAL on osx. No probs if this is unnecessary but would love to contribute.

You're doing amazing work. Cheers!

### DIFF
--- a/src/audio/sinks/openal_sink.c
+++ b/src/audio/sinks/openal_sink.c
@@ -2,8 +2,8 @@
 
 // On Apple OS (Mac and IOS), al.h and alc.h are in nonstandard locations
 #ifdef __APPLE__
-#include <al.h>
-#include <alc.h>
+#include <OpenAL/al.h>
+#include <OpenAL/alc.h>
 #else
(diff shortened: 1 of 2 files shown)


ericmhalvorsen:
Saw the same issue, this fixed it for me. Compiled OpenAL from source downloaded from this link: http://kcat.strangesoft.net/openal.html

@Vagabond Any recommended method for MacOSX to avoid this issue?

Vagabond:
I think it depends on how you installed openal, I think I installed it from macports or from homebrew.

Vagabond:
I'm not sure how to solve this properly, maybe we can use cmake to look for the header files?

katajakasa:
I wouldn't go changing the default seek location; i'd rather modify the cmake script to look from different places, indeed. Or, I'd rather accept a patch that does that :P

katajakasa:
Possibly fixed in master with commit 23e5aa96bed9be006fac66e95bc2bb6268759213
Please test.

PR #272, opened 2015-04-06
Enhancements

exander77:
We can now use eight keys for control as in original OMF 2097.
The menu for keyboard binding looks the same as in OMF 2097.
Both players can control the menu (corresponds to OMF 2097).
Sliders go to 0 indifferent of has_off, the only difference is in rendering (corresponds to OMF 2097).
Slider for speed has now has_off set to 0  (corresponds to OMF 2097).


katajakasa:
Hey! Looking good with a quick glance. Looks like you got along with the gui system -- it's currently a bit of an undocumented mess. I'll take a deeper look tomorrow when I'm more awake :)

exander77:
Yeah, the GUI system is a little messy so It felt more like hacking. :) Also my code will surely need some improvements in the future.

exander77:
If you see something wrong on want something do in other way, just inform me and I modify the pull request.

exander77:
Comparison OMF 2097 (top) vs OPENOMF (bottom).
screenshot_2015-04-06_14-32-03

katajakasa:
Bug: In custom keyboard menu, pressing DONE doesn't leave the menu
Bug: In Input device selection menu, pressing left or right keyboard plays the selection sound many times at the same time, instead of just once. This causes an off reverb effect and louder sound than it should be.

exander77:
Will look into those bugs and prepare fixes.

exander77:
Could you please explain second bug more thoroughly?

katajakasa:
Sure. When you press the Left Keyboard or Right Keyboard buttons, for some reason the system sets several button press sound effect copies to playback. There are at probably 3-5 copies of the same sound effect playing at the same time. Because the same sound plays at the same time (with some millisecond variations), the output sound sounds louder and more rumbly. Try comparing the sound between the keyboard selection buttons and any other menu button. Best I can do :)

exander77:
Could you check the latest commit for those bugs?

exander77:
For now i moved default color in the label_create.
Also tried something new with the dynamically changing color - toying with the idea to be able dynamically change color using callback and modifer (int value), even per character.

katajakasa:
Render functions are called as many times per second as possible. Tick functions are always called every n milliseconds. Rendering can therefore happen as fast (or slowly) as necessary, and it doesn't affect the game speed, because the game logic itself is bound to tick. Therefore anything that depends on constant speed should be handled in tick functions.

Note that there should be nothing except font/text settings in text_settings struct. If you need a component specific local variable, then you should put it to the components local struct. Eg. there is already a tick variable in textrender, which bounces between 0 and 120 (see textbutton_tick function).

katajakasa:
Tested the DONE button, it seems to work correctly now.

exander77:
And the sound problem?

katajakasa:
No change.

exander77:
Should be fixed now.

Also fixed bug, that when one player chooses left/right, the other player is set to the opposite.
Previous implementation created colliding key bindings.

katajakasa:
Btw; please try to keep a single matter in a single pull request. If there are features which differ a lot from the one you are currently working on, open up a new branch and pull request. That way it is easier for me to review and pull. For example the latest commit could easily be a new pull request, since it has nothing to do with the input menu :/

exander77:
Not a problem, I can split it to several branches.
I try to clean it up a little bit today and I will create several smaller pull requests.

exander77:
Closing this, created new branches for appropriated commits:
- controls
- look_and_feel
- bugfixes

PR #276, opened 2015-04-06
Lower CMake version

pmjdebruijn:
Lots of distro's are still shipping CMake 2.8, so it would be rather nice to keep CMake 2.8 compatibility for another year or so.

Also, a minor tweak to the window title.


katajakasa:
I won't merge the title change, because the game isn't one must fall 2097. Granted, we are using its resources, but the engine is not even close to original OMF2097 in functionality. Therefore it would be a bad idea to advertise this game as the original.

Also, there should only ever be a single complete feature in one pull request. For example now I can't pull the cmake changes easily even though they are good, because another feature is blocking the request :/

pmjdebruijn:
Fair enough.

katajakasa:
Looks good, merging.

PR #278, opened 2015-04-08
Controls

exander77:
Fixing various controls related things.


exander77:
The keyboard binding menu is in new position and it has its own background and if I read the code correctly background is bound to guiframe. Also components are rendered inside their guiframe and the keyboard binding menu is on completely different position.

If I missed something, let me know and I will redo in a better way.

katajakasa:
Okay, that sounds like a valid reason. I kind of dislike the use of guiframe there, since it was kind of meant to be the root node of all gui stuff; kind of like a single window. I don't think that there are better solutions for this currently, however.

A request: Add a couple of TODO comments to the beginning of the menu source file, and explain there the use of guiframe quickly.

exander77:
Yes, I hate it too :(, I have some idea about relative/absolute/fixed positioning (line in css), but still not sure what to do with background. Will add TODOs.

katajakasa:
Review fixes seem to be applied and code is sufficiently explained. Merging.

katajakasa:
Btw, if you still feel like working on more stuff, feel free to pop in on irc, at #omf on freenode. We usually respond slightly faster there :P (also available via web interface at http://www.omf2097.com/chat/)

exander77:
Yes, I might drop by on irc. OMF was one of the best games of my youth. I still have my original tournament file. :)

PR #282, opened 2015-05-27
gamecontrollerdb: add RetroLink Saturn Classic Controller

pmjdebruijn:
The RetroLink Saturn Classic Controller has a (digital) D-pad,
six regular buttons, left/right shoulder buttons and a start button.


katajakasa:
As was discussed in IRC, we'd be happier to see this go to upstream gamecontrollerdb repository (seems like there is a PR ongoing), and then pull from there. So closing this.

katajakasa:
This is now done in a3274a85aece44d959348c4813b6a95cf3ee4186

Issue #285, opened 2015-09-07
Better resources paths


katajakasa:
Instead of using /usr/share/openomf, we should move resources to /usr/share/games/openomf.

We should also use subdirectories, so that distros can linkfarm if they need to. Eg. 
- /usr/share/games/openomf/resources (For old/original resources)
- /usr/share/games/openomf/new (For new resources. Better name ?)
- /usr/share/games/openomf/plugins (Maybe ? Or some other place ? These will be dll/so's).

Vagabond:
I am fine with this, although we should ideally make this configurable, or maintain it as a relative path to the binary, because the Debian way is not the only way.

katajakasa:
Yeah, let's keep it relative. So "../share/games/openomf/..." etc. We could also add a commandline argument for overriding it I guess, so that distros could make a wrapper shell script if necessary.

animehunter:
I like the idea of overriding the path on the commandline.

katajakasa:
So after taking a look at our path management code, I think we don't absolutely need a separate path for new and old resources. It would just be a pile of work and I don't feel like doing it. However, commits/patches welcome, if somebody WANTS to take care of it :D

Vagabond:
Should we notify the debian people?

Rondom:
Would you be open to a pull request making the paths configurable during build?
The Debian-way is not the only way and wrapper scripts sound a bit overkill for such a simple thing.

Vagabond:
Yes, please.

katajakasa:
@Rondom PR's welcome. Btw, I took a look at the debian ITP's. It's probably not a good idea to separately package libshadowdive (please see libsd readme: https://github.com/omf2097/libShadowDive).

Rondom:
I read the Readme and I have actually created test-packages using both static linking and dynamic linking. Both is possible, but static linking is frowned upon in Debian, so I plan to ask others for feedback.

katajakasa:
Well, just note then that the library will be broken and API will change pretty much all the time. Afaik that isn't very well received in debian either :P

katajakasa:
Oh, and. Current debian libdumb package will not work with openomf; it's too old. So you either need to update the debian package or statically link the kode54 library. libxmp may work, but the playback quality isn't as good.

Vagabond:
[...]

PR #287, opened 2015-09-09
Change unix default resource path, add environment variable overides

Vagabond:
The default resource path on linux, in release mode, is now
../share/games/openomf/, relative to the binary. This can be overridden
by the environment variable OPENOMF_RESOURCE_DIR, which should be an
absolute path with a trailing slash.

A similar environment variable override is provided for the plugin path:
OPENOMF_PLUGIN_DIR.

cc @katajakasa @animehunter

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -424,6 +424,6 @@ INSTALL(TARGETS openomf
     COMPONENT Binaries
 )
 INSTALL(FILES resources/openomf.bk resources/openomf_icon.png
-    DESTINATION share/openomf/
+    DESTINATION share/games/openomf/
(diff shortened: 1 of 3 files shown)


katajakasa:
This looks good to me.

Rondom:
I think it would make sense to separate the original OMF2097-data from the OMF data. What do you think?

Vagabond:
We only have 2 openomf specific files, isn't that a little overkill?

katajakasa:
This seems to work and I think the critical features are now covered. We can talk more about the separation of new and old resources in the bug ticket #285. Merging to master.

Issue #292, opened 2015-09-14
Fight vs Kreissack should be at Stadium


gdeda:
In original OMF, the final fight vs Kreissack's Nova always happens at Stadium.
This won't happen in OpenOMF right now

PR #293, opened 2015-09-14
Force Arena 0 when fighting Kreissack in 1 player mode

gdeda:
just to be useful in a way, or to try to be :)

Related to issue #292


src/game/scenes/vs.c
```diff
@@ -358,6 +358,9 @@ int vs_create(scene *scene) {
     if (player2->selectable) {
         // player1 gets to choose, start at arena 0
         local->arena = 0;
+    } else if (player2->pilot_id == 10) {
```
Vagabond: You can use PILOT_KREISSACK here rather than a magic number.

Vagabond:
Thanks so much, this looks great! If you can fix the one comment I had, I'll be happy to merge it.

Issue #290, opened 2015-09-14
Defeat status should be assessed on hit, not when HAR touches floor


gdeda:
All of throwing or moves which won't consist in a direct impact, rather in an opponent being throwed away, wont' trigger KO status: it continues to fight even if it reaches 0 health.
I can confirm this at least on Jaguar Overhead Throw and the Thorn throw.
In original OMF the KO is triggered just as the throwing move is performed by the player.

gdeda:
Corrected the title due to the following findings:
After some testings, looks like the bug is not related to throwing, but about the KO being triggered only when the opponent touches floor (as evidence, you can still move if you give a deadly punch/kick). I have smashed a enemy HAR with a Katana throw and the KO is triggered only if the opponent touches floor.
Same with Jaguar overhead throw: only if the opponent touches the floor first, the KO is triggered. Otherwise the match just continues.

IIRC, the KO should be triggered at the succesful attack, not when the opponent touches floor...

gdeda:
Ok, I think I have pinpointed the issue. Or at least I believe, as my tiny coding experience is about C# :)

The issue should be because, according to har.c, every time the HAR goes down, there is a series of checks about if the match should end, or not...

har.c, line 470:
            if(h->state != STATE_DEFEAT
                && h->state != STATE_FALLEN
                && h->health <= 0
                && h->endurance <= 0
                && player_is_last_frame(obj)) {

                h->state = STATE_DEFEAT;
                har_set_ani(obj, ANIM_DEFEAT, 0);
                har_event_defeat(h);
            } else if(pos.y >= (ARENA_FLOOR-5) &&
                      IS_ZERO(vel.x) &&
                      player_is_last_frame(obj)) {
                if (h->state == STATE_FALLEN) {
                    if (h->health <= 0 && h->endurance <= 0) {
                        // fallen, but done bouncing
                        h->state = STATE_DEFEAT;
                        har_set_ani(obj, ANIM_DEFEAT, 0);
                        har_event_defeat(h);
                    } else {
                        h->state = STATE_STANDING_UP;
                        har_set_ani(obj, ANIM_STANDUP, 0);
                        har_event_land(h);
                    }
                } else {
                    har_finished(obj);

What I understand here is, if the HAR goes down AND the health AND endurance is 0, ONLY THEN the defeat is called.

The HAR, anyway, won't die if health goes to 0 but there is some endurance, what instead happens is the next shot should bring endurance to 0.

har.c, line 519
[...]

PR #296, opened 2015-09-17
HARs should go down at 0 health no matter endurance level

gdeda:
Partially related to #290 

In OMF 2.1, if a HAR has little health, a shot which bring its health to 0 will end the match.
In OpenOMF, in same situation, the HAR is still standing because there is some endurance left.

This PR should make things much in line with OMF behaviour


gdeda:
Please wait before merging this, I realized I have pushed also merge commits and other stuff meant to keep my fork updated. I need some time to familiarize with git and come up with a clean PR...

PR #298, opened 2015-09-18
HARs have to go down if a hit bring them at 0 health, no matter endurance

gdeda:
Partially related to #290

In OMF 2.1, if a HAR has little health, a shot which bring its health to 0 will end the match.
In OpenOMF, in same situation, the HAR is still standing because there is some endurance left.

This PR should make things much in line with OMF behaviour

(I hope a pair of reset and force push on my fork were succesful to produce a clean PR :) )

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -514,23 +514,21 @@ void har_take_damage(object *obj, str* string, float damage) {
     }
 
     // Handle health changes
-    int oldhealth = h->health;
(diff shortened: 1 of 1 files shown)


katajakasa:
PR looks good. Testing commit.

katajakasa:
Tested against the original, feels close to identical. Also cleans up the code nicely.

Good to go, merging.

PR #343, opened 2017-08-03
Attempt to fix flails scrap move

katajakasa:
We don't really _know_ what CW tag is supposed to do, but since only flails scrap + destruction uses it, we can piggyback on it to hack other stuff. 

This is kind of an **ugly hack** and doesn't yet fix everything.


katajakasa:
Nope, this needs a rethink.

PR #344, opened 2017-08-07
Use safer string handling functions

katajakasa:


Vagabond:
Compile problems on travis.

katajakasa:
Yeah, it looks like the previous libpng fix doesn't work on travis, because travises ubuntu machines have an ancient libpng version :/

katajakasa:
Now it passes travis. Merging.

PR #356, opened 2019-02-10
Move libSD and tools over here

katajakasa:


Rondom:
Are you interested in keeping the history? I can give it a try, if you want.

katajakasa:
@Rondom Feel free! This is mostly an experimentation branch on how things *could* be done -- if you can manage this and keep the git history, then that would be awesome :)

Rondom:
I submitted #361, which supersedes this PR. It should contain all changes of this PR.

PR #364, opened 2019-03-01
Libshadowdive integration work

katajakasa:
WIP

### DIFF
--- a/src/formats/internal/writer.c
+++ b/src/formats/internal/writer.c
@@ -62,24 +62,24 @@ int sd_write_buf(sd_writer *writer, const char *buf, int len) {
     return 1;
 }
 
-int sd_write_buf(sd_reader *reader, const str *buf) {
+int sd_write_str(sd_writer *writer, const str *buf) {
     if(str == NULL) {
-        sd_write_uword(w, 0);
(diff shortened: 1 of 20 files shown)


src/formats/internal/writer.c
```diff
@@ -62,6 +62,26 @@ int sd_write_buf(sd_writer *writer, const char *buf, int len) {
...
+}
+
+int sd_write_cstr(sd_writer *writer, const str *buf) {
+    if(str == NULL) {
+        sd_write_uword(writer, 0);
+        return;
+    }
+    uint16_t len = strlen(str) + 1;
+    sd_write_uword(writer, len);
+    sd_write_buf(writer, str, len);
```
mrannanj, line 82: this and the function above have return type int, but never return any int

mrannanj (changes requested):

src/formats/internal/writer.c
katajakasa: Ya, it should return the length written. But yeah, this whole PR is pretty much a "let's stab in some experiments" -- never going to be merged. Instead, we should do other, smaller PRs.

PR #370, opened 2020-07-26
formats: Fix buffer overflow in sd_script_encoded_length()

mrannanj:
sizeof(tmp)-s can easily be less than zero there, and since tmp is not
used elsewhere, we can just remove the -s.


katajakasa:
Looks good to me :)

PR #374, opened 2020-07-26
Rec stuff

mrannanj:


CMakeLists.txt
```diff
@@ -233,6 +233,9 @@ endif()
 target_link_libraries(openomf ${CORELIBS})
 set_property(TARGET openomf PROPERTY C_STANDARD 11)
 
+# This is to make loading test resources possible
+add_compile_definitions(OPENOMF_ROOT_DIR="${CMAKE_SOURCE_DIR}")
```
katajakasa, line 237: Probably best move this to the "testing stuff" block -- afaiks not something that should be needed for the main binary.

mrannanj: Moved it to the testing stuff block and renamed it TESTS_ROOT_DIR

PR #380, opened 2020-09-12
testing: Return nonzero exit value when a test fails

mrannanj:


testing/test_main.c
```diff
@@ -74,7 +74,14 @@ int main(int argc, char **argv) {
     CU_basic_set_mode(CU_BRM_VERBOSE);
     CU_basic_run_tests();
 
-end:
-    CU_cleanup_registry();
-    return CU_get_error();
+end:;
```
katajakasa, line 77: ;

mrannanj: Empty statement is needed there unless the int ret declaration is moved elsewhere. I can move it to the top of the function though.

mrannanj: Moved int ret declaration to the start of the function.

PR #389, opened 2021-07-02
Fgets

mrannanj:


tools/stringparser/main.c
```diff
@@ -59,7 +59,10 @@ int main(int argc, char* argv[]) {
     if (strcmp("-", *astr->sval) == 0) {
         from_stdin = 1;
         char *tmp_str = malloc(512);
-        fgets(tmp_str, 512, stdin);
+        if (fgets(tmp_str, 512, stdin) == NULL) {
+            printf("fgets(512, stdin) failed\n");
```
katajakasa, line 63: This should really use the PERROR macro instead

mrannanj: Yeah, I'll make a proper fix instead ..

tools/stringparser/main.c
```diff
@@ -57,9 +57,16 @@ int main(int argc, char* argv[]) {
...
-        from_stdin = 1;
-        char *tmp_str = malloc(512);
-        fgets(tmp_str, 512, stdin);
+        if (fgets(tmp_str, 512, stdin) == NULL) {
+            if (feof(stdin)) {
+                printf("Unexpected end of stdin\n");
+            } else if (ferror(stdin)) {
+                printf("Error reading stdin\n");
+            } else {
+                printf("What case is this ..?\n");
```
mrannanj, line 66: This is a mystery, is it possible that no error happened or this was not EOF?

tools/stringparser/main.c
```diff
@@ -57,9 +57,16 @@ int main(int argc, char* argv[]) {
...
 
     if (strcmp("-", *astr->sval) == 0) {
-        from_stdin = 1;
-        char *tmp_str = malloc(512);
-        fgets(tmp_str, 512, stdin);
+        if (fgets(tmp_str, 512, stdin) == NULL) {
+            if (feof(stdin)) {
+                printf("Unexpected end of stdin\n");
+            } else if (ferror(stdin)) {
+                printf("Error reading stdin\n");
```
mrannanj, line 64: Here it is unclear how to find out what the error was, since man page explicitly says these functions do not set errno..

PR #410, opened 2021-10-29
Extend string lib

katajakasa:
WIP! This is currently a design plan, not actually working code. Needs more thought... possibly inplace versions of the functions too ? What would be the easiest ?

### DIFF
--- a/include/utils/str.h
+++ b/include/utils/str.h
@@ -3,44 +3,196 @@
 
 #include <stddef.h>
 #include <stdarg.h>
+#include <stdbool.h>
 
-typedef struct str_t {
+typedef struct {
     size_t len;
     char* data;
 } str;
 
-void str_create(str *string);
-void str_create_from_cstr(str *string, const char *cstr);
-void str_create_from_data(str *string, const char *data, size_t len);
+/**
+ * @brief Create an empty string object.
+ * @details Creates a new, empty string object. After creating the object, it must be freed
+ *          using the str_free. 
+ * @param dst Target string object
+ */
+void str_create(str *dst);
+
+/**
+ * @brief Create a new string from existing string.
+ * @details Source string contents will be copied.
+ * @param dst Target string object
+ * @param src Source string to copy
+ */
+void str_from(str *dst, const str *src);
+
+/**
+ * @brief Create a string object from C string (null terminated buffer).
+ * @details Source string contents will be copied.
+ * @param dst Target string object
+ * @param src Source C string to copy
+ */
+void str_from_c(str *dst, const char *src);
+
+/**
+ * @brief Create a string object from memory buffer.
+ * @details Source buffer content will be copied. Please make sure that the source
+ *          buffer contains no null characters, and is actual printable text string.
+ * @param dst Target string object
+ * @param buf Source buffer object
+ * @param len Source buffer object length (bytes)
+ */
+void str_from_buf(str *dst, const char *buf, size_t len);
+
+/**
+ * @brief Create a string object by format string
+ * @details This uses snprintf internally, so usual printf format strings are valid.
+ * @param dst Target string object
+ * @param format Format string
+ * @param ... Arguments list
+ */
+void str_from_format(str *dst, const char *format, ...);
+
+/**
+ * @brief Create a string object from range slice
+ * @details This will copy a text range from another string object. Note that source string
+ *          must not be same as target string!
+ * @param dst Target string object
+ * @param src Source string object
+ * @param start Slice start position
+ * @param end Slice end position
+ */
+void str_from_slice(str *dst, const str *src, size_t start, size_t end);
+
+/**
+ * @brief Free string object
+ * @details Frees up any memory used by the string object. Usage of the string after freeing it
+ *          is undefined behaviour.
+ * 
+ * @param string String to free
+ */
 void str_free(str *string);
 
-size_t str_size(const str *string);
-void str_slice(str *dst, const str *src, size_t start, size_t end);
-void str_copy(str *dst, const str *src);
+// String modification
+
+/**
+ * @brief Change all letters to uppercase.
+ */
+void str_toupper(str *dst);
+
+/**
+ * @brief Change all lettes to lowercase.
+ */
+void str_tolower(str *dst);
+
+/**
+ * @brief Strip whitespace characters on the right side of a string.
+ */
+void str_rstrip(str *dst);
+
+/**
+ * @brief Strip whitespace characters on the left side of a string.
+ */
+void str_lstrip(str *dst);
+
+/**
+ * @brief Strip whitespace characters on both sides of a string.
+ */
+void str_strip(str *dst);
+
+/**
+ * @brief Append a string to another string.
+ * @details Source string content will be copied.
+ */
 void str_append(str *dst, const str *src);
+
+/**
+ * @brief Append a C string (null terminated) to a string object.
+ * @details Source string content will be copied.
+ */
 void str_append_c(str *dst, const char *src);
-void str_prepend(str *dst, const str *src);
-void str_prepend_c(str *dst, const char *src);
-void str_remove_at(str *string, size_t pos);
 
-void str_printf(str *dst, const char *format, ...);
+/**
+ * @brief Append a C buffer (NOT null terminated) to a string object.
+ * @details Source buffer content will be copied. Please make sure that the source
+ *          buffer contains no null characters, and is actual printable text string.
+ */
+void str_append_buf(str *dst, const char *buf, size_t len);
 
-int str_first_of(const str *string, char find, size_t *pos);
-int str_next_of(const str *string, char find, size_t *pos);
-int str_last_of(const str *string, char find, size_t *pos);
+/**
+ * @brief Replace content in string with something else.
+ * @details Replace occurrences of the search string with the replacement string.
+ *          Limit parameter can be used to limit replacement to a certain number
+ *          of occurences if so wanted (or -1 for unlimited).
+ * @param dst Target string to modify
+ * @param seek String to search and replace
+ * @param replacement The replacement value
+ * @param limit Number of replacements performed (-1 for unlimited).
+ */
+void str_replace(str *dst, const char *seek, const char *replacement, int limit);
 
-int str_equal(const str *string, const str *string_b);
-int str_equal_c(const str *string, char *char_compare);
+/**
+ * @brief Get string length.
+ */
+size_t str_size(const str *string);
 
-void str_toupper(str *string);
-void str_tolower(str *string);
+/**
+ * @brief Find first occcurence of a character in a given string.
+ */
+bool str_first_of(const str *string, char find, size_t *pos);
 
+/**
+ * @brief Find last occcurence of a character in a given string.
+ */
+bool str_last_of(const str *string, char find, size_t *pos);
+
+/**
+ * @brief Check if strings match
+ */
+bool str_equal(const str *a, const str *b);
+
+/**
+ * @brief Check if strings match (match with C null terminated string)
+ */
+bool str_equal_c(const str *a, const char *b);
+
+/**
+ * @brief Check if strings match (match with C buffer)
+ */
+bool str_equal_buf(const str *a, const char *buf, size_t len);
+
+/**
+ * @brief Get a character at given index.
+ */
 char str_at(const str *string, size_t pos);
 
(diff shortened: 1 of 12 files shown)


turol:
Easiest? Upgrade to C++

katajakasa:
@turol If I was interested in switching languages, I would go for Rust ... or Python :P

katajakasa:
@turol Also, don't you mean "downgrade to C++" ? :P

turol:
Second easiest way, use someone else's library: https://github.com/maxim2266/str

src/utils/str.c
```diff
@@ -56,110 +56,198 @@ void str_printf(str *dst, const char *format, ...) {
...
-
-    dst->len += size;
-    dst->data[dst->len] = 0;
 }
 
-void str_slice(str *dst, const str *src, size_t start, size_t end) {
+void str_from_slice(str *dst, const str *src, size_t start, size_t end) {
+    assert(dst != src);
+    assert(start >= 0);
+    assert(end >= 0);
```
mrannanj, line 77: start and end are size_t, so these asserts cannot be false

mrannanj (changes requested):

src/utils/str.c
```diff
@@ -56,110 +56,198 @@ void str_printf(str *dst, const char *format, ...) {
...
+}
+
+void str_append_buf(str *dst, const char *buf, size_t len) {
+    size_t offset = dst->len;
+    STR_REALLOC(dst, dst->len + len);
+    memcpy(dst->data + offset, buf, len);
+    STR_ZERO(dst);
+}
+
+static bool _find_next(const str *string, char find, size_t *pos) {
```
mrannanj, line 157: so i wonder why this underscore start on the static functions? They are static, no one is going to be using the functions outside this file anyway

src/utils/str.c
```diff
@@ -56,110 +56,198 @@ void str_printf(str *dst, const char *format, ...) {
...
-int str_next_of(const str *string, char find, size_t *pos) {
-    for(size_t i = *pos; i < string->len; i++) {
+void str_replace(str *dst, const char *seek, const char *replacement, int limit) {
+    size_t seek_len = strlen(seek);
+    size_t replacement_len = strlen(replacement);
+    assert(seek_len > 0);
+    int found = 0;
+    size_t diff = replacement_len - seek_len;
+    size_t current_pos = 0;
+    while(_find_next(dst, seek[0], &current_pos) && (found < limit || limit < 0)) {
```
mrannanj, line 174: what's the meaning of limit less than 0?

src/utils/str.c
```diff
@@ -56,110 +56,198 @@ void str_printf(str *dst, const char *format, ...) {
...
         }
     }
-    return 0;
+    return false;
 }
 
-int str_last_of(const str *string, char find, size_t *pos) {
-    for(size_t i = string->len; i > 0; i--) {
+bool str_last_of(const str *string, char find, size_t *pos) {
+    for(size_t i = string->len; i >= 0; i--) {
```
mrannanj, line 214: this looks like i can underflow

src/utils/str.c
```diff
@@ -169,31 +257,33 @@ char str_at(const str *string, size_t pos) {
...
+    long out = strtol(string->data, &end, 10);
+    if (string->data == end) {
+        *result = 0;
+        return false;
+    }
+    if (out > INT_MAX || out < INT_MIN) {
+        *result = 0;
+        return false;
+    }
+    *result = out;
```
mrannanj, line 285: i guess explicit cast would be ok to add here

mrannanj (changes requested):

src/utils/str.c
```diff
@@ -4,44 +4,44 @@
...
+#include <limits.h>
 #include <ctype.h>
 #include <stdio.h>
 #include <assert.h>
 
-void str_create(str *string) {
-    string->len = 0;
-    string->data = omf_calloc(1, 1);
-}
+#define STR_ALLOC(var, size) do { \
```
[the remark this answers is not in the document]
katajakasa: Sure, renamed.

src/utils/str.c
katajakasa: Kinda sorta reminder that "this function is static" without needing to check the declaration.

src/utils/str.c
katajakasa: Yes, it can. Fixed.

src/utils/str.c
katajakasa: Actually, I think I will just drop the str_to_int function, since it doesn't seem to be used at all anymore. I'll just leave the str_to_float and str_to_long since they are simple.

src/utils/str.c
katajakasa: "replace all found instances, no limit"

src/utils/str.c
```diff
@@ -56,110 +56,198 @@ void str_printf(str *dst, const char *format, ...) {
...
 
-void str_prepend_c(str *dst, const char *src) {
-    size_t srclen = strlen(src);
-    dst->data = omf_realloc(dst->data, dst->len + srclen + 1);
-    memmove(dst->data + srclen, dst->data, dst->len);
-    memcpy(dst->data, src, srclen);
-    dst->len += srclen;
-    dst->data[dst->len] = 0;
+void str_rstrip(str *dst) {
+    // This is simple, just reduze sice and set ending 0.
```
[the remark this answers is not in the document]
katajakasa: Typo fixed

src/utils/str.c
```diff
@@ -56,110 +56,198 @@ void str_printf(str *dst, const char *format, ...) {
...
+    STR_ZERO(dst);
 }
 
-void str_copy(str *dst, const str *src) {
-    dst->data = omf_realloc(dst->data, src->len + 1);
-    dst->len = src->len;
-    memcpy(dst->data, src->data, dst->len);
-    dst->data[dst->len] = 0;
+void str_free(str *dst) {
+    if(dst == NULL) { return; }
```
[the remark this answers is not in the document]
katajakasa: Yes, but we are not freeing dst, we are freeing dst->data :)

src/utils/str.c
katajakasa: Righto, fixed.

include/utils/str.h
```diff
@@ -3,44 +3,62 @@
...
-int str_to_float(const str *string, float *result);
-const char* str_c(const str *string);
-const char* str_c_alloc(const str *string);
+// Data type conversion
+bool str_to_float(const str *string, float *m);
+bool str_to_long(const str *string, long *m);
+bool str_to_int(const str *string, int *m);
+
+// Compatibility with C strings
+const char* str_c(const str *src);
```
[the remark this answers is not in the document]
katajakasa: I slightly prefer the first one, though I don't think we have a rule about this one. I would be inclined not to care ... perhaps something that could be handled by autolinter in the future :/

src/utils/str.c
katajakasa: Dropped.

include/utils/str.h
```diff
@@ -3,44 +3,62 @@
 
 #include <stddef.h>
 #include <stdarg.h>
+#include <stdbool.h>
 
-typedef struct str_t {
+typedef struct {
     size_t len;
     char* data;
```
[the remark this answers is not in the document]
katajakasa: This is another one where I slightly prefer the "char* data" one when I think about it, but when I type stuff, I don't really care. I've probably used both interchangeably earlier anyways :/

include/utils/str.h
```diff
@@ -3,44 +3,62 @@
...
 
-void str_create(str *string);
-void str_create_from_cstr(str *string, const char *cstr);
-void str_create_from_data(str *string, const char *data, size_t len);
+/**
+ * @brief Create an empty string object.
+ * @details Creates a new, empty string object. After creating the object, it must be freed
+ *          using the str_free. 
+ * 
+ * @param string Target string to initialize
```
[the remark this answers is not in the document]
katajakasa: I just left periods out from all the type declarations. I don't know, maybe this is in my head, but feels wrong to use periods as they might not always be valid sentences :P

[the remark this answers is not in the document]
mrannanj: All right, so I'll point out the one that does have ending period.

include/utils/str.h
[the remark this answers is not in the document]
mrannanj: Okay let's ignore the style issues for now.

include/utils/str.h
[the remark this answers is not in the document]
mrannanj: Ignoring this too.

src/utils/str.c
mrannanj: Okay guess that's fine

src/utils/str.c
```diff
@@ -56,110 +56,198 @@ void str_printf(str *dst, const char *format, ...) {
...
+    }
 }
 
-void str_prepend(str *dst, const str *src) {
-    dst->data = omf_realloc(dst->data, dst->len + src->len + 1);
-    memmove(dst->data + src->len, dst->data, dst->len);
-    memcpy(dst->data, src->data, src->len);
-    dst->len += src->len;
-    dst->data[dst->len] = 0;
+static size_t _strip_size(const str *src, bool left) {
```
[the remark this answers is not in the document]
mrannanj: Well, don't mind this too much, so let's leave it there..

src/utils/str.c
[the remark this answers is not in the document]
mrannanj: oh okay

include/utils/str.h
```diff
@@ -3,44 +3,196 @@
...
+ * @param end Slice end position
+ */
+void str_from_slice(str *dst, const str *src, size_t start, size_t end);
+
+/**
+ * @brief Free string object
+ * @details Frees up any memory used by the string object. Usage of the string after freeing it
+ *          is undefined behaviour.
+ * 
+ * @param string String to free.
```
mrannanj, line 72: This now had period at the end, but others don't

src/utils/str.c
```diff
@@ -209,9 +209,9 @@ bool str_first_of(const str *string, char find, size_t *pos) {
 }
 
 bool str_last_of(const str *string, char find, size_t *pos) {
-    for(size_t i = string->len; i >= 0; i--) {
-        if(string->data[i-1] == find) {
-            *pos = i - 1;
+    for(size_t i = 0; i < string->len; i++) {
+        if(string->data[string->len - i] == find) {
```
mrannanj, line 212: Is the intent that this can find the last ```\0``` character too? I kind of would not expect that this function can access ```string->data``` at the ```string->len``` position since ...

mrannanj: i was thinking maybe it could be changed to
```
for(size_t i = 1; i <= string->len; i++) {
    ...
```

katajakasa: string->len does not account for the \0. Allocated length is always string->len + 1 :)

mrannanj: Yeah, I know, was just wondering why is it that you can use this to find the ```\0``` char, but I guess this is fine.

include/utils/str.h
katajakasa: asdf.

include/utils/str.h
```diff
@@ -3,44 +3,196 @@
...
+ */
+void str_from(str *dst, const str *src);
+
+/**
+ * @brief Create a string object from C string (null terminated buffer).
+ * @details Source string contents will be copied.
+ * @param dst Target string object
+ * @param src Source C string to copy
+ */
+void str_from_c(str *dst, const char *src);
```
mrannanj, line 35: The old name ```str_from_cstr``` was better, since this ```str_from_c``` makes you think string from one character

src/utils/str.c
katajakasa: Ah, right, I misread your comment. Right you are, it is missing -1 on the seek. Added now and bumped in an additional test.

mrannanj (changes requested):

include/utils/str.h
katajakasa: I like the str_from_c better, because:
- Faster to type
- Means we are "importing" from pure C side, so we can use this suffix in other places too if we need to
- I kept misreading cstr as just str for some reason

mrannanj: Yeah, i guess it's ok

mrannanj (approved):

Issue #398, opened 2021-10-18
Refactor reader and writer functions


katajakasa:
Libshadowdive had its internal memreader/reader/memwriter/writer functions in internal/ subdirectory (formats/internal/). We should move those to the openomf projects general utility directory (utils/) and refactor them to match openomf coding style.

* Remove "sd_" prefix from functions
* Rename mwriter to memwriter in function and struct names (etc).
* Rename mreader to memreader in functio nand struct names (etc).
* Drop the pointless struct names in all of these readers and writers -- just keep the typedeffed name.
* Remove the "SD_" prefix from guards, and make sure the guards match the filenames
* Add consts where possible
* Check that docs are in order, and add some new ones if possible.

Verify afterwards:
* Everything compiles
* Unit-tests
* Game still starts :)
* Utilities (aftool, bktool, ... ) should still work

PR #414, opened 2021-11-07
Memwriter

mrannanj:
#398

### DIFF
--- a/include/formats/internal/memreader.h
+++ b/include/formats/internal/memreader.h
@@ -1,31 +1,31 @@
-#ifndef _SD_MEMREADER_H
-#define _SD_MEMREADER_H
+#ifndef MEMREADER_H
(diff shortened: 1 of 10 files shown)


katajakasa:
Seems fine to me. I will read this through in more detail tomorrow, but if nothing shows up, I think I will just go ahead and merge.

mrannanj:
> Seems fine to me. I will read this through in more detail tomorrow, but if nothing shows up, I think I will just go ahead and merge.

The ticket said to also look for places where const could be added, but these commits only do the rename.

katajakasa:
> The ticket said to also look for places where const could be added, but these commits only do the rename.

Yup, that's fine. I like when people make things in small, manageable pieces :)

Also, went through it, nothing jumped at my face immediately.

PR #422, opened 2021-12-17
global: Rename sd_altpal_file to altpal_file

mrannanj:


include/formats/altpal.h
```diff
@@ -9,25 +9,25 @@
 
 #include "formats/palette.h"
 
-#ifndef _SD_ALTPAL_H
-#define _SD_ALTPAL_H
+#ifndef _ALTPAL_H
```
turol, line 12: This is illegal. Identifiers starting with _ and a capital letter are reserved for the C implementation. Use `ALTPAL_H`. If there are other such guards you should fix them too.

mrannanj: Starting with underscore is the convention in this codebase for include guards, need to ask @katajakasa

katajakasa: Well, to be fair, using them is totally legan, just discouraged. Compiler would complain loudly if it was illegal. But sure, I don't mind, might as well change it.

turol: With the right command line option clang does complain about them.

katajakasa: That is not complaining /loudly/ :) But I'm not going to argue further when I essentially agree with you :P

mrannanj: Added a commit to remove the underscores.

PR #439, opened 2022-01-05
Add string handling to file reader and writer

katajakasa:
These will allow us to easily read and write str objects in parsers.


mrannanj (approved):
Ok, just need to squash

Issue #444, opened 2022-01-16
Clang-Format


katajakasa:
Clang-format improvements

* [x] Run clang-format for the entire codebase
* [x] Make sure formatter output seems good, make any config changes needed
* [x] After codebase is formatted, enable check in Actions to ansure new code is also properly formatted. Raise error if it's not.

katajakasa:
Closing as fixed.

PR #445, opened 2022-01-16
global: Run make clangformat

mrannanj:
#444


katajakasa:
Hum. To reduce the delta, I wonder if clang-format allows configuring these:
- Allow if() and for(), instead of forhing "if ()" and "for ()" (with space)
- In lists, each item on its own line
- Require braces on even single-line ifs

katajakasa:
Hom. In addition, it seems like clang-format wants to handle switch-case indent differently from our standard indent. I wonder if that could be changed with an option to match our style for reduced delta ?

katajakasa (approved):
Seems like an acceptable delta :)

PR #446, opened 2022-01-16
Run formatting check in GA CI using clang-format

katajakasa:
CMake integration was not handling some things correctly, so I swapped the integration to just pure GLOB search of C and H files and then running clang-format against them. This way it does not care about what targets are enabled.

### DIFF
--- a/include/controller/controller.h
+++ b/include/controller/controller.h
@@ -1,12 +1,13 @@
 #ifndef CONTROLLER_H
 #define CONTROLLER_H
 
-#include "game/protos/object.h"
 #include "game/objects/har.h"
+#include "game/protos/object.h"
 #include "game/utils/serial.h"
 #include "utils/list.h"
 
-enum {
+enum
+{
     ACT_STOP = 0x01,
     ACT_KICK = 0x02,
     ACT_PUNCH = 0x04,
@@ -18,7 +19,8 @@ enum {
     ACT_FLUSH = 0x100
 };
 
-enum {
+enum
+{
     CTRL_TYPE_KEYBOARD,
     CTRL_TYPE_GAMEPAD,
     CTRL_TYPE_NETWORK,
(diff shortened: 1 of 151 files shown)


include/controller/controller.h
```diff
@@ -18,20 +18,9 @@ enum {
...
-};
+enum { CTRL_TYPE_KEYBOARD, CTRL_TYPE_GAMEPAD, CTRL_TYPE_NETWORK, CTRL_TYPE_AI, CTRL_TYPE_REC };
 
-enum {
-    EVENT_TYPE_ACTION,
-    EVENT_TYPE_SYNC,
-    EVENT_TYPE_HB,
-    EVENT_TYPE_CLOSE
-};
+enum { EVENT_TYPE_ACTION, EVENT_TYPE_SYNC, EVENT_TYPE_HB, EVENT_TYPE_CLOSE };
```
mrannanj, line 23: let's try to find a setting where enums are on separate lines

mrannanj: maybe `AllowShortEnumsOnASingleLine: false` could work

mrannanj: Looks like `AllowShortEnumsOnASingleLine: false` works, but it puts the starting brace on a new line, and doesn't look like
```
BraceWrapping:
  AfterEnum:       false
```
is enough to put the brace on same as the enum ..

katajakasa: Probably this: https://reviews.llvm.org/D93938'

Let's just leave it as true.

mrannanj: So the current configuration will now fix this issue when clang-format gets fixed.

katajakasa: No, I didn't want to leave any bombs to the config that would cause odd behaviour on version update. Let's just manually set it to false and update the clang-format when new one is released.

PR #448, opened 2022-01-18
Fix problems found by clang-tidy

katajakasa:
Fix problems found by clang-tidy

### DIFF
--- a/src/controller/ai_controller.c
+++ b/src/controller/ai_controller.c
@@ -203,8 +203,9 @@ int char_to_act(int ch, int direction) {
  *
  * \return Void.
  */
-void chain_controller_cmd(controller *ctrl, int commands[], ctrl_event **ev) {
-    for(int i = 0; i < sizeof *commands; i++) {
+void chain_controller_cmd(controller *ctrl, const int commands[], ctrl_event **ev) {
+    size_t n = sizeof(*commands) / sizeof(int);
+    for(int i = 0; i < n; i++) {
         controller_cmd(ctrl, commands[i], ev);
     }
 }
@@ -348,7 +349,7 @@ int get_enemy_range(const controller *ctrl) {
     har *h = object_get_userdata(o);
     object *o_enemy = game_state_get_player(o->gs, h->player_id == 1 ? 0 : 1)->har;
 
-    int range_units = abs(o_enemy->pos.x - o->pos.x) / 30;
+    int range_units = fabs(o_enemy->pos.x - o->pos.x) / 30;
     switch(range_units) {
         case 0:
         case 1:
@@ -1224,8 +1225,8 @@ int ai_har_event(controller *ctrl, har_event event) {
                 if(learning_moment(a) && a->thrown >= MAX_TIMES_THROWN) {
                     DEBUG("\e[33mAI adjusting in response to repeated throws.\e[0m");
                     // avoid defensive tactics
-                    if(pilot->att_def > -90)
-                        pilot->att_def = -10;
+                    if(pilot->att_def > 90)
+                        pilot->att_def = 10;
                     // favor sniper tactics
                     if(pilot->att_sniper < 90)
                         pilot->att_sniper += 10;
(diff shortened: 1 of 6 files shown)


src/controller/ai_controller.c
```diff
@@ -1224,8 +1225,8 @@ int ai_har_event(controller *ctrl, har_event event) {
                 if(learning_moment(a) && a->thrown >= MAX_TIMES_THROWN) {
                     DEBUG("\e[33mAI adjusting in response to repeated throws.\e[0m");
                     // avoid defensive tactics
-                    if(pilot->att_def > -90)
```
katajakasa, line 1227: all the att_* variables are uint, so they cannot be negative

src/formats/language.c
```diff
@@ -40,9 +40,8 @@ int sd_language_load(sd_language *language, const char *filename) {
 
     // Find out how many strings there are in the file
     unsigned int string_count = 0;
-    unsigned int offset = 0;
     long file_size = sd_reader_filesize(r);
-    while((offset = sd_read_udword(r)) < file_size) {
```
katajakasa, line 45: offset was not used here

src/game/utils/settings.c
```diff
@@ -25,29 +25,29 @@
...
-typedef union field_default_u {
+typedef union {
     char *s;
     int i;
     int b;
     double f;
 } field_default;
 
-typedef struct field_t {
+typedef struct {
```
katajakasa, line 43: struct variables sorted for reduced padding (and memory use). Not a big thing, but might as well.

src/controller/ai_controller.c
```diff
@@ -203,8 +203,9 @@ int char_to_act(int ch, int direction) {
  *
  * \return Void.
  */
-void chain_controller_cmd(controller *ctrl, int commands[], ctrl_event **ev) {
-    for(int i = 0; i < sizeof *commands; i++) {
+void chain_controller_cmd(controller *ctrl, const int commands[], ctrl_event **ev) {
+    size_t n = sizeof(*commands) / sizeof(int);
```
mrannanj, line 207: I don't see how this can know the size of the array, does this really work?

mrannanj (changes requested):

src/controller/ai_controller.c
katajakasa: Oh yeah, I forgot to check. So this should technically work due to Compiler Magic, but I'm not sure if it works through function calls. I know that the old sizeof at least did not work correctly :/

Probably best make a test case.

katajakasa: Or actually, probably best make a macro and pass a size_t size parameter.

mrannanj: Kind of wonder if that ever worked, but definitely there is no way to know the size of that array from those parameters.

katajakasa: It probably worked before because it was reading the pointer size, which was either 4 or 8. That would then have caused the loop to read all the content plus some extra :/

mrannanj: I pushed something that should work to this branch, but also `chain_consider_tactics()` is broken the same way

katajakasa: Added fixes for that.

src/controller/ai_controller.c
```diff
@@ -1128,16 +1128,18 @@ int ai_har_event(controller *ctrl, har_event event) {
...
             } else {
                 // we hit with a HAR attack
-                chain_consider_tactics(ctrl, (int[]){TACTIC_QUICK, TACTIC_TRIP, TACTIC_GRAB, TACTIC_PUSH, TACTIC_CLOSE,
-                                                     TACTIC_SHOOT, TACTIC_TURTLE, TACTIC_SPAM});
+                int tacs[] = {TACTIC_QUICK, TACTIC_TRIP,  TACTIC_GRAB,   TACTIC_PUSH,
+                              TACTIC_CLOSE, TACTIC_SHOOT, TACTIC_TURTLE, TACTIC_SPAM};
+                chain_consider_tactics(ctrl, tacs, N_ELEMENTS(tacs));
             }
-            break;
+        } break;
```
mrannanj, line 1139: These added ending braces are after the break in the previous such commit, should put them the same way on both. I don't care which way.

katajakasa: Fix'd them all, I think.

mrannanj (approved):

mrannanj (approved):

PR #456, opened 2022-03-12
Enhancement/omf 433 capitalize on stun

Snafulator:
Managed to extract just the parts that make AI capitalize on stun better.

They will continue with their current tactic after enemy stun so long as the current tactic is an aggressive tactic.

Additionally they will also apply the same behavior when the enemy hits a hazard.

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -78,6 +78,7 @@ void har_event_air_turn(har *h) {
 void har_event_walk(har *h, int direction) {
     // direction is -1, 1, for backwards and forwards
     har_event event;
+    memset(&event, 0, sizeof(event));
     event.type = HAR_EVENT_WALK;
     event.player_id = h->player_id;
     event.direction = direction;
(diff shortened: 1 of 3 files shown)


src/game/objects/har.c
```diff
@@ -148,6 +148,14 @@ void har_event_hazard_hit(har *h, bk_info *info) {
     fire_hooks(h, event);
 }
 
+void har_event_enemy_hazard_hit(har *h) {
+    har_event event;
```
mrannanj, line 152: hmmh, this is a bit unsafe memory use .. should memset the sctruct to zero always before use. However, it looks this is always used like this in the file. Could a commit be added that memsets all such har_events to zero before passing it on to `fire_hooks()`

Snafulator: Hmm. Could we create an issue to specifically address this in all of the events? That way we can keep the scope of the changes in this ticket down.

PR #457, opened 2022-05-29
Str

mrannanj:
Resurrecting and finishing https://github.com/omf2097/openomf/pull/364 - incomplete

### DIFF
--- a/include/utils/str.h
+++ b/include/utils/str.h
@@ -142,6 +142,11 @@ void str_copy_c(str *dst, const char *src);
 void str_replace(str *dst, const char *seek, const char *replacement, int limit);
 
 /**
+ * @brief Get and remove last character of the string.
+ */
+char str_pop_ch(str *src);
+
+/**
  * @brief Get string length.
  */
 size_t str_size(const str *string);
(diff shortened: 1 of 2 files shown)


include/utils/str.h
```diff
@@ -119,30 +124,40 @@ void str_append_c(str *dst, const char *src);
  */
 void str_append_buf(str *dst, const char *buf, size_t len);
 
+/**
+ * @brief Replace the string object data with a copy of the C string (null terminated).
+ */
+void str_copy_c(str *dst, const char *src);
```
katajakasa, line 130: I specifically dropped this, as it tends to make things weird by reusing variable for different things. str_from_c() is there instead to create a new string from C string. Is there some specific place where you need this ?

include/utils/str.h
```diff
@@ -119,30 +124,40 @@ void str_append_c(str *dst, const char *src);
...
  * @param seek String to search and replace
  * @param replacement The replacement value
  * @param limit Number of replacements performed (-1 for unlimited).
  */
 void str_replace(str *dst, const char *seek, const char *replacement, int limit);
 
+/**
+ * @brief Get and remove last character of the string.
+ */
+char str_pop_ch(str *src);
```
katajakasa, line 147: Afaiks the need for this got dropped as there is now the rstrip function. Is there some specific need for this ?

PR #464, opened 2023-04-14
Move everything from include dir to src

katajakasa:
Easier to find the headers when they are next to the .c files.

Opinions ?


Vagabond:
I guess if you want to make a 'devel' package, if anyone links against the headers for some reason, it'd be slightly more annoying?

mrannanj:
This is changing files from \n to \r\n line break. Seems not deliberate?

katajakasa:
> This is changing files from \n to \r\n line break. Seems not deliberate?

Now it should fix all \r\n to \n instead (some files still had \r\n from ye olden times).

PR #468, opened 2023-04-22
OpenGL

katajakasa:
WIP, do not merge.

Plan:
1. Dump sprites to an RG8 texture atlas (R = color, G = alpha)
2. Dump vertices and texture coords to a VBO
3. Dump palettes and remappings to UBO(s) (or possibly consider 1D textures later for gotta go faster ?)
4. Create an R8 texture and bind it to FBO, set FBO as target.
5. Draw all objects. Shader should do any palette trickery now.
6. Switch back to default FBO, and bind the previous render target texture.
7. Ram the textuer through RGBA conversion with current palette.
8. Run any upscaler shaders we want also (scalehqx, vhs, scanline, etc).
9. Profit!

Notes:
* Currently FBO size is 320x200, and we change the viewport to match that on fbo rendering phase. If/when we do sprite packs for larger sprites, we should bump the size. Perhaps some config option ?
* When everything else works, consider using https://github.com/libretro/glsl-shaders for upscaling the result in rgba phase.

It might make sense to split the rendering into three, as the news caster guy wants images from arenas without the text and UI crap. So perhaps first render all game objects, then snatch any screenshots, and then draw the UI crap.

Cleanup TODO:
* Consider getting rid of discard in glsl shaders
* Maybe texture1d for palette + remappings instead of ubo ?
* https://www.reddit.com/r/gamedev/comments/2j17wk/a_slightly_faster_bufferless_vertex_shader_trick/
* Look into the fread() bug. Something related to windows or mingw or msys2 ?


src/video/video.c
```diff
@@ -75,6 +71,116 @@ bool create_gl_context() {
...
+    glAttachShader(program_id, shader);
+    return true;
+}
+
+bool create_program(GLuint *program_id, const char *vertex_shader, const char *fragment_shader) {
+    GLuint id = glCreateProgram();
+    if(!load_shader(id, GL_VERTEX_SHADER, vertex_shader))
+        goto error_0;
+    if(!load_shader(id, GL_FRAGMENT_SHADER, fragment_shader))
+        goto error_0;
```
turol, line 169: This leaks vertex shader if fragment shader fails.

katajakasa:
> Prefer libepoxy over libglew. It supports GL ES too and starts up faster.

Interesting, I will take a look. Glew is a total pain to maintain.

src/video/video.c
katajakasa: pfft, as if it would ever fail.


... right ?

src/video/texture_atlas.c
```diff
@@ -0,0 +1,65 @@
...
+static GLuint create_atlas_texture(uint16_t width, uint16_t height) {
+    GLuint texture_id = 0;
+    glGenTextures(1, &texture_id);
+    glBindTexture(GL_TEXTURE_2D, texture_id);
+    glPixelStorei(GL_UNPACK_ALIGNMENT, 1);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+    glTexImage2D(GL_TEXTURE_2D, 0, GL_RG_INTEGER, width, height, 0, GL_RG_INTEGER, GL_UNSIGNED_BYTE, NULL);
```
turol, line 32: Prefer `TexStorage` functions over `TexImage`. Need to check extension is supported. Though there are no modern platforms with supported OpenGL where it isn't. Maybe WebGL

Either generate and set mipmaps or set max miplevel to 0.

turol: Not sure if `GL_RG_INTEGER` is a guaranteed supported type. Might have to check that first.

katajakasa: Yeah, none of this is likely to work. I just copied my ancient GL2.1 tex atlas for a starting point and threw in some ideas. I may make a separate stencil for the sprites to handle the alpha too, we shall see. Also, it might be better to copy the sprite bytecodes to a buffer and then just render the sprites from that. We shall see.

src/video/video.c
katajakasa: (fixed)

src/video/texture_atlas.c
katajakasa: Righto, the bytecode was too complicated for a shader to do easily, maybe just a texture atlas then. TexStorage* stuff seems to be GL 4.2... I suppose I could use those regarless as you said, but I just don't want to deal with it now :D Maybe an optional future path for it, if it is supported.

For texture data, I need to dump the shaders paletted textures so 1 byte of image data, and then some sort of 1 bit alpha value (on/off). Might be better to hold two textures, one with GL_RED for images and then something else for the alpha (I wonder if GL has something for black/white stuff that could be used for that alpha).

src/video/opengl/shaders.c
```diff
@@ -128,6 +128,11 @@ void bind_uniform_4fv(GLuint program_id, const char *name, GLfloat *data) {
     glUniformMatrix4fv(ref, 1, GL_FALSE, data);
 }
 
+void bind_uniform_block(GLuint program_id, const char *name, GLuint binding) {
+    GLuint ref = glGetUniformBlockIndex(program_id, name);
+    glUniformBlockBinding(program_id, ref, binding);
```
turol: Check `ref` before use. If something went wrong and the block is not there you will use -1 as unfirom block index and unpredictable stuff might happen.

turol:
Reading from UBO with a varying offset might be a slow path. Use a texture instead. Either `GL_TEXTURE_1D` or a 2d texture with width or height of 1.

Using `discard` in shader is potentially a slow path. Prefer outputting fully transparent color instead.

src/video/opengl/shaders.c
katajakasa: Good to know. Checks added and one type fixed.

katajakasa:
> Reading from UBO with a varying offset might be a slow path. Use a texture instead. Either `GL_TEXTURE_1D` or a 2d texture with width or height of 1.
> 
> Using `discard` in shader is potentially a slow path. Prefer outputting fully transparent color instead.

Interesting. I added these on my future to-do list.

turol:
You probably want this: https://www.reddit.com/r/gamedev/comments/2j17wk/a_slightly_faster_bufferless_vertex_shader_trick/

katajakasa:
> You probably want this: https://www.reddit.com/r/gamedev/comments/2j17wk/a_slightly_faster_bufferless_vertex_shader_trick/

Lol. On to the list it goes.

src/utils/io.c
```diff
@@ -0,0 +1,55 @@
...
+    file_seek(handle, 0, SEEK_END);
+    file_size = ftell(handle);
+    file_seek(handle, pos, SEEK_SET);
+    return file_size;
+}
+
+void file_read(FILE *handle, char *buffer, long size) {
+    size_t ptr = 0;
+    size_t read_size;
+    while(1) {
```
turol, line 38: Why this complicated crap? Just `fread` once with the correct size and error out if it fails.

katajakasa: Because if I read with anything over 1024, fread fails to EINVAL (invalid argument).

katajakasa: Also, technically it is possible for fread to return less then <count> amount of elements, while returning more on subsequent reads. It's valid behaviour for fread.

turol: I have never seen that. What kind of a fucked up platform are you using?

katajakasa: Windows 10 + MSYS2 (mingw) triggered this one. I don't think you can get much more fucked up than that ?

katajakasa: And yes, works fine on ubuntu before and now.

turol: At least 1. put some comments on it and 2. use an optimistic strategy so non-broken platforms aren't slowed down.

katajakasa: Not going to touch this anymore otherwise. I added the rest of the work on my todo list for cleanup.

shaders/palette.frag
```diff
@@ -1,17 +1,36 @@
...
 };
 
-uniform sampler2D image;
-
 void main() {
-    vec4 texel = texture(image, tex_coord);
-    if (texel.g == 0) {
-        discard;
+    // Get rid of the branching later! We can split this into two or three shaders pretty easily
+    // TODO: FIX LATER.
```
turol, line 15: Also sampling textures with implicit gradients in divergent control flow is UB. See https://themaister.net/blog/2019/09/12/the-weird-world-of-shader-divergence-and-lod/

You might get away with this in a fullscreen pass but generally it can result in really funky artifacts

shaders/palette.frag
```diff
@@ -1,17 +1,36 @@
 #version 330 core
 
 in vec2 tex_coord;
+in vec2 fbo_coord;
+in float blend_mode;
```
turol, line 5: You can pass integers to shaders with `glVertexAttribIPointer`.

katajakasa: The current pushed stuff is heavily WIP and getting hacked at as we figure out the actual tags, so none of this is likely to survive that.  This pr is going to get messy for a while. This stuff should eventually get fixed, but not now :P

shaders/palette.frag
katajakasa: Yeppers. In any case this particular shader is going to go through a ton of new stuff while we try to get the original games crap sorted out. It will probably end up in a couple of different vertex shaders and a larger bunch of fragment stuff.

katajakasa: Shaders make it pretty easy to test out stuff :)

src/engine.c
```diff
@@ -142,6 +142,12 @@ void engine_run(engine_init_flags *init_flags) {
                     if(e.key.keysym.sym == SDLK_F1) {
                         take_screenshot = 1;
                     }
+                    if(e.key.keysym.sym == SDLK_F9) {
```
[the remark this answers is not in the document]
katajakasa: Yes. Not doing it in this PR though, otherwise I will be refactoring everything and this PR becomes the new main branch :D

src/video/surface.c
```diff
@@ -288,6 +288,53 @@ void surface_convert_to_rgba(surface *sur, screen_palette *pal, int pal_offset)
     create_hash(sur);
 }
 
+static uint8_t find_closest_gray(screen_palette *pal, int range_start, int range_end, int ref) {
+    uint8_t closest = 0, current;
+    int closest_dist = 256, dist;
+
+    for(int i = range_start; i <= range_end; i++) {
+        current = pal->data[i][0]; // Grayscale, r = g = b so pick any.
+        dist = current - ref;
```
turol, line 297: You probably want to use squared distance.

katajakasa:
Further work items are now here: https://github.com/omf2097/openomf/issues/612

PR #471, opened 2023-04-29
Tournament fixes

Vagabond:
* Fix winnings to use right field
* Fix pilot names in tournament mode newsroom
* Show proper size gauges for HAR upgrades

thanks to @joneirik for data entry


Vagabond:
Gotta check the maxes when greying out the buy buttons, because some HARs have more leg/speed power than arm speed/power

PR #494, opened 2023-07-20
Rank loss in tournament

gdeda:
This makes the player to lose a rank position in the tournament when a match is lost.
A check is added to avoid going out of the array list of ranked participants if the player loses the match while being already at the bottom of the ladder.


gdeda:
NOTE: will work only in OMF tournaments, not in custom tournaments

katajakasa:
I'm not close to my dev pc so cannot test this now, but it seems good so I'm just going to trust you and merge it ;)

Vagabond:
does seem reasonable

PR #512, opened 2024-06-26
Don't wipe existing flags

fjtrujy:
## Description
I have faced this error during the PS2 port.

The current `cmake` was always overriding the flags we were passing, it should instead just append the needed ones.

Cheers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,11 +19,11 @@ OPTION(USE_FORMAT "Use clang-format for checks" OFF)
 
 # These flags are used for all builds
 set(CMAKE_C_STANDARD 11)
-set(CMAKE_C_FLAGS "-Wall")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -18,12 +18,11 @@ OPTION(USE_TIDY "Use clang-tidy for checks" OFF)
 OPTION(USE_FORMAT "Use clang-format for checks" OFF)
 
 # These flags are used for all builds
-set(CMAKE_C_STANDARD 11)
```
katajakasa, line 21: did you mean to drop CMAKE_C_STANDARD ?

fjtrujy: That was a mistake!

fjtrujy: Fixed!

Issue #509, opened 2024-06-25
Bundle OMF2097 assets with releases


mrannanj:
Create a release job that bundles OMF2097 assets with the release package for ease of install.

PR #513, opened 2024-06-29
Asset bundling

mrannanj:
Goal is to implement #509 - omf assets bundling on release packages


.gitignore
```diff
@@ -15,3 +15,23 @@ mingw_build/
 .vscode
 cmake-build-debug/
 cmake-build-release/
+
+resources/*.AF
```
katajakasa, line 19: What does this .gitignore change do ?

mrannanj: This is where debug build expects resources to be found, but guess it should not be part of this pull request.

Issue #516, opened 2024-06-30
Initialize cmake build directory with openomf.bk


mrannanj:
Copy openomf.bk to cmake build resources directory to ease new developer experience setting up.

PR #518, opened 2024-07-01
CMakeLists.txt: Copy openomf.bk to destination resources

mrannanj:
#516

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,9 @@ else()
     message(STATUS "Development: Unit-tests are disabled")
 endif()
 
+# Copy openomf.bk to destination resources directory to ease development setup.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -262,6 +262,9 @@ else()
     message(STATUS "Development: Unit-tests are disabled")
 endif()
 
+# Copy openomf.bk to destination resources directory to ease development setup.
+file(COPY resources/openomf.bk DESTINATION resources)
```
katajakasa, line 266: Will this work if someone builds to the project root ? It will try to copy in-place ?

mrannanj: I tested this, and it did not break, so cmake . worked.

PR #524, opened 2024-07-13
Use argtable2 if found, but fall back to argtable3 if needed.

cynic:
Some distributions (like mine, NixOS) don't package argtable2; on the other hand, I note that some package managers (like brew) don't support argtable3.  The main game and the tools can compile with either.  This patch allows compilation with argtable2 or argtable3.


mrannanj:
Thank you.

PR #567, opened 2024-09-10
Trnlist leak

mrannanj:


src/utils/list.c
```diff
@@ -16,6 +17,9 @@ void list_free(list *list) {
     now = list->first;
     while(now != NULL) {
         next = now->next;
+        if(list->free) {
```
katajakasa, line 20: Should this also do the free on list_delete ?

mrannanj: oh, yeah, didn't notice that existed.

PR #575, opened 2024-09-13
Fix a use-after-free in xmp context handling

Vagabond:


src/audio/audio.c
```diff
@@ -254,6 +263,7 @@ void audio_close() {
             audio->xmp_context = NULL;
         }
         omf_free(audio);
+        audio = NULL;
```
mrannanj, line 266: `omf_free()` nulls this already

PR #578, opened 2024-09-16
game: Add arena state hash

mrannanj:


src/utils/log.c
```diff
@@ -32,15 +33,16 @@ void log_hide(char mode, const char *fn, const char *fmt, ...) {
...
-        fprintf(handle, "[%7u][%c] ", _log_tick, mode);
+        len = snprintf(log_buf, sizeof(log_buf), "[%7u][%c] ", _log_tick, mode);
     }
     va_list args;
     va_start(args, fmt);
-    vfprintf(handle, fmt, args);
+    vsnprintf(&log_buf[len], sizeof(log_buf) - len, fmt, args);
     va_end(args);
-    fprintf(handle, "\n");
+    fprintf(handle, "%s\n", log_buf);
```
turol, line 46: Why not `fputs`?

mrannanj: Well, the previous code was using this stream handle thing .. dunno

mrannanj: In theory, the log could be going to a file that is not stdout.

mrannanj: Oh right, fputs does have `FILE*` argument .. not sure it's worth changing ..

mrannanj: Okay, so how to add the ending newline nicely if we are using fputs?

turol: `fputc`

mrannanj: Nah, would make for 2 writes.

turol: That should not matter, `FILE *` functions are buffered.

PR #589, opened 2024-09-19
Exchange state hashes and try to improve consistency

Vagabond:


src/controller/net_controller.c
```diff
@@ -40,14 +40,18 @@ typedef struct {
     int last_acked_tick;
     int last_har_state;
     int last_traced_tick;
+    int peer_last_hash;
+    int peer_last_hash_tick;
+    int last_hash;
```
mrannanj, line 45: Better change `arena_state_hash()` to return int if using it as int here.

mrannanj (approved):

PR #594, opened 2024-09-24
Lobby view

mrannanj:


src/formats/bk.c
```diff
@@ -95,6 +96,21 @@ int sd_bk_load(sd_bk_file *bk, const char *filename) {
     sd_reader *r;
     int ret = SD_SUCCESS;
 
+    size_t fn_len = strlen(filename);
```
Vagabond, line 99: Might be better to have a 'bk_from_pcx' function or something

Vagabond: oh, I see, you're just using the existing BK loader per scene, I guess this is fine then

PR #598, opened 2024-09-26
Implement horizontal menus and start converting the lobby to use them

Vagabond:


src/game/gui/textbutton.c
```diff
@@ -119,6 +119,8 @@ component *textbutton_create(const text_settings *tconf, const char *text, int d
 
     textbutton *tb = omf_calloc(1, sizeof(textbutton));
     tb->text = strdup(text);
+    printf("width of %s is %d\n", text, text_width(tconf, text));
```
mrannanj, line 122: Better use `DEBUG()` to control where the output is going

PR #600, opened 2024-09-28
Hashmap improvements

katajakasa:


src/utils/hashmap.c
```diff
@@ -160,8 +73,25 @@ int hashmap_resize(hashmap *hm, int n_size) {
     // Free old bucket list and assign new list and size of the hashmap
     omf_free(hm->buckets);
     hm->buckets = new_buckets;
-    hm->buckets_x = n_size;
-    return 0;
+    hm->capacity = new_size;
+}
+
+static void hashmap_enlarge_check(hashmap *hm) {
```
Vagabond, line 79: Some comments on what these are actually checking would be nice

katajakasa: Comments added.

src/utils/hashmap.c
```diff
@@ -51,33 +51,32 @@ void hashmap_create_cb(hashmap *hm, hashmap_free_cb free_cb) {
...
-    if(new_size == hm->capacity || new_size < 1)
+    if(new_size <= hm->capacity)
         return;
 
     // Allocate and zero out a new memory blocks for the resized bucket list
-    hashmap_node **new_buckets = omf_calloc(new_size, sizeof(hashmap_node *));
+    size_t old_bytes = hm->capacity * sizeof(hashmap_node *);
+    size_t new_bytes = new_size * sizeof(hashmap_node *);
+    hm->buckets = omf_realloc(hm->buckets, new_bytes);
+    memset(((void *)hm->buckets) + old_bytes, 0, new_bytes - old_bytes);
```
mrannanj, line 61: Hmmh. I think pointer arithmetic on void pointer is undefined behaviour. Is the buckets pointer void? Should change it to unsigned char.

mrannanj: Okay, checked that it has the type `hashmap_node **`, so could something like `memset(&hm->buckets[hm->capacity], 0, new_bytes - old_bytes);` work here?

mrannanj: Checked that pointer arithmetic on void * is supported by gcc, so maybe don't need to change it.

PR #609, opened 2024-09-30
Add a text render mode to indicate active/inactive/disabled colors

Vagabond:


src/game/gui/text_render.c
```diff
@@ -8,10 +8,13 @@
 
 void text_defaults(text_settings *settings) {
```
katajakasa, line 9: I'm not sure if the color indexes should be set here. Do they apply for all scenes ?

src/game/gui/text_render.c
```diff
@@ -8,10 +8,13 @@
...
     memset(settings, 0, sizeof(text_settings));
-    settings->cforeground = 0xFF;
+    settings->cforeground = 0xFD;
+    settings->cselected = 0xFF;
+    settings->cinactive = 0xFE;
+    settings->cdisabled = 0xC0;
 }
 
-int text_render_char(const text_settings *settings, int x, int y, char ch) {
+int text_render_char(const text_settings *settings, text_mode state, int x, int y, char ch) {
```
katajakasa, line 17: Why do we need the mode ? Wouldn't it be enough to just change the test->cforeground ?

katajakasa (dismissed):

PR #632, opened 2024-10-07
Less temporary memory allocations & share AF data when possible

Vagabond:
This needs somewhat careful review, especially the string parser changes.

### DIFF
--- a/src/formats/script.c
+++ b/src/formats/script.c
@@ -203,20 +203,16 @@ static int read_int(const str *src, int *original) {
         return 0; // If we found no number, just return default 0.
     }
 
-    str test;
-    str_from_slice(&test, src, start, end);
-    if(!str_to_int(&test, &value)) {
+    if(!str_to_int(src, start, &value)) {
         value = 0; // This should never really happen.
     }
-    str_free(&test);
     *original = end;
     return value;
 }
 
-static bool test_tag_slice(const str *test, sd_script_tag *new, str *src, int *now) {
-    const int len = str_size(test);
-    const int jmp = *now + len;
-    if(sd_tag_info(str_c(test), &new->has_param, &new->key, &new->desc) == 0) {
+static bool test_tag_slice(const char *test, int size, sd_script_tag *new, str *src, int *now) {
+    if(sd_tag_info(test + *now, size, &new->has_param, &new->key, &new->desc) == 0) {
+        const int jmp = *now + strlen(new->key);
         // Ensure that tag has no value, if value is not desired.
         if(!new->has_param && find_numeric_span(src, jmp) > jmp) {
             return false;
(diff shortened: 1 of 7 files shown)


src/game/gui/textslider.c
```diff
@@ -16,35 +16,37 @@ typedef struct {
...
 
-static void textslider_render(component *c) {
+static void render_text(component *c) {
     textslider *tb = widget_get_obj(c);
-    str txt;
-    str_from_format(&txt, "%s ", tb->text);
+    int off = snprintf(tb->txt, sizeof(tb->txt), "%s ", tb->text);
     if(tb->has_off && *tb->pos == 0) {
-        str_append_c(&txt, "OFF");
+        snprintf(tb->txt + off, sizeof(tb->txt) - off, "%s", "OFF");
```
katajakasa, line 29: I dislike this -- the idea with str type is that we can later convert it to utf-8. This is taking steps backwards. How about we cache the str to textslider instead ?

katajakasa:
I think we can close this now ?

Vagabond:
no, I think there's still some valid changes here, and maybe a broader discussion about when we should be using `str`

katajakasa:
> no, I think there's still some valid changes here, and maybe a broader discussion about when we should be using `str`

Okay. I'll start then :) I'd rather use the str whenever it comes to text, since it abstracts away the buffer length management and makes the code look a bit easier to read. snprintf and friends especially make it really hard to read (or properly understand!) the code :/

src/formats/script.c
```diff
@@ -238,19 +234,16 @@ static bool parse_tag(sd_script_tag *new, str *src, int *now) {
     }
 
     // Check if tag is legit.
-    str test;
     for(int m = 3; m > 0; m--) {
-        str_from_slice(&test, src, *now, *now + m);
-        if(str_equal_c(&test, "usw") && find_numeric_span(src, *now + m) > *now + m) {
+        // str_from_slice(&test, src, *now, *now + m);
+        if(strncmp(src->data + *now, "usw", m) == 0 && find_numeric_span(src, *now + m) > *now + m) {
```
Nopey, line 239:
```suggestion
        if(strncmp(str_c(src) + *now, "usw", 3) == 0 && find_numeric_span(src, *now + m) > *now + m) {
```
The strncmp call should compare all 3 characters like the str_equal_c did before, not just the first m.

src/utils/str.h
```diff
@@ -215,7 +215,7 @@ bool str_to_long(const str *string, long *m);
  * @return true If success
  * @return false If failure (result value is invalid)
  */
-bool str_to_int(const str *string, int *m);
+bool str_to_int(const str *string, size_t offset, int *m);
```
Nopey, line 218: Please document the new `offset` param.

src/utils/str.c
```diff
@@ -286,13 +286,14 @@ bool str_to_long(const str *string, long *result) {
...
 }
 
-bool str_to_int(const str *string, int *result) {
-    long value;
-    bool got = str_to_long(string, &value);
-    if(got) {
-        *result = clamp_long_to_int(value);
+bool str_to_int(const str *string, size_t pos, int *result) {
+    char *end;
+    *result = strtol(string->data + pos, &end, 10);
```
Nopey, line 291: Please bounds-check `pos` (aka `offset`) param, at least with an assert()

src/game/gui/textslider.c
```diff
@@ -16,35 +16,37 @@ typedef struct {
...
     int has_off;
     int positions;
+    char txt[20];
 
     void *userdata;
     textslider_slide_cb slide;
 } textslider;
 
-static void textslider_render(component *c) {
+static void render_text(component *c) {
```
Nopey, line 25: Please rename this method to something that does not contain the word "render."

textslider_format? textslider_updatestr?

Nopey (changes requested):
scene.c fighter de-duplication is a neat shortcut, seems worthwhile.

script.c  changes reduce number of extremely small stack based strings we create from slices, but no longer effects allocator pressure. Unsure if text slider changes are still relevant, but maybe?

replacing str usage with snprintf is icky, especially after katajakasa added the small string optimization.. *checks commit log* a few hours after this PR was opened? x)

katajakasa:
Probably deprecated now ? textslider stuff is fixed with the text objects, and srt should use a small stack buffer for the small text objects.

Vagabond:
We can kill this.

PR #633, opened 2024-10-07
GL remapping/blending stuff (WIP)

katajakasa:


shaders/rgba.frag
```diff
@@ -5,33 +5,30 @@ in vec2 tex_coord;
...
-    int real_index = color_index + add_index;
-    if (color_index <= 0) {
-        color = colors[0].rgba;
+    int add_index = clamp(int(texel.g * 255.0), 0, 19);
+    int sub_index = clamp(int(texel.b * 255.0), 0, 19);
+    vec4 tex_add = texture(remaps, vec2(texel.r, (add_index + offset) / 19.0));
+    vec4 tex_sub = texture(remaps, vec2(texel.r, (sub_index + offset) / 19.0));
+
+    if (add_index > 0) {
+        color = colors[int(tex_add.r * 255.0)].rgba;
```
turol, line 25: Loading  from UBO in divergent control flow is suboptimal. You want to calculate the index and then load after reconvergence. Something like
```
int index = tex_add.r * 255.0;
if (...) {
  index = ...;
}

color = colors[index];
```

katajakasa: Good to know. Something to fix after I have a better idea on how the remappings are actually supposed to work; there are some exceptions to their use which I have to dig from the original game and slap onto the shader :/

katajakasa: I threw in a quick fix for this.

PR #634, opened 2024-10-07
Stacked STR

katajakasa:


Vagabond:
This doesn't trigger any asan warnings and doesn't seem to make all those thousands of micro allocations anymore.

mrannanj (approved):

PR #650, opened 2024-10-11
BUILD.md: include libepoxy in apt-get & homebrew package lists

rjoonas:
Hi, tried to build the project for the first time using the instructions in BUILD.md and noticed they were missing one recently added dependency.

Also upgraded the assets pack URL to HTTPS to avoid an unnecessary redirect from the insecure URL.


katajakasa:
Seems good, thank you!

PR #653, opened 2024-10-12
Rework dialog boxes to use modern components and sizing

Vagabond:
Dialog boxes were previously very hardcoded. This PR makes them use the more modern container widgets and text rendering so they are more flexible with regard to their contents.

This also improves the menu widget layout behavior and adds support for centered horizontal menus, which are used for the dialog buttons.

Dialog boxes are now made up of a menu containing a textlabel, for the dialog text, then a horizontal, centered menu that holds the dialog buttons.


Vagabond:
Don't merge this yet, it breaks the mainmenu somewhat.

Vagabond:
Ok, this should be good now.

src/game/scenes/vs.c
```diff
@@ -617,7 +618,7 @@ int vs_create(scene *scene) {
     scene_set_render_cb(scene, vs_render);
     scene_set_render_overlay_cb(scene, vs_render_overlay);
     scene_set_input_poll_cb(scene, vs_input_tick);
-    scene_set_dynamic_tick_cb(scene, vs_dynamic_tick);
+    // scene_set_dynamic_tick_cb(scene, vs_dynamic_tick);
```
katajakasa, line 621: No longer needed ?

Vagabond: it never seemed necessary

katajakasa: Ok. Just remove the line and the tick function completely ? We can find it in git history if it's needed.

Vagabond: Fixed, I just forgot to remove it.

katajakasa (approved):

PR #662, opened 2024-10-17
Work on making the textinput more faithful and more useful

Vagabond:


src/game/gui/textinput.c
```diff
@@ -21,56 +23,107 @@ typedef struct {
...
+            }
+            return 0;
+            break;
+        case ACT_DOWN:
+            tb->buf[tb->pos] = tb->buf[tb->pos] - 1;
+            if(tb->buf[tb->pos] < 32) {
+                tb->buf[tb->pos] = 126;
+            }
+            return 0;
+            break;
```
mrannanj, line 96: Seems a bit strange for all these cases to have both return and break, the break would never be reached.

src/game/gui/text_render.c
```diff
@@ -242,11 +243,11 @@ void text_render(const text_settings *settings, text_mode mode, int x, int y, in
 
         // Skip spaces
         int k = 0;
-        for(; k < line_len; k++) {
+        /*for(; k < line_len; k++) {
             if(text[ptr + k] != ' ')
                 break;
             real_len--;
-        }
+        }*/
```
mrannanj, line 250: Could the commented out code be removed?

src/game/gui/text_render.c
```diff
@@ -72,17 +73,17 @@ int text_render_char(const text_settings *settings, text_mode state, int x, int
...
+    /*for(i = 0; i < len; i++) {
         if(ptr[i] != ' ')
             break;
-    }
-    if(i == len - 1) {
+    }*/
+    /*if(i == len) {
         return i;
-    }
+    }*/
```
mrannanj, line 86: Could the commented out code be removed?

mrannanj (changes requested):

Vagabond:
I've updated this to restore whitespace stripping (as an option), to fix the broken test.

src/game/gui/textinput.c
Vagabond: I didn't want the code to behave unexpectedly in case someone ever refactored it to remove the return. I think C switch statements should ALWAYS have breaks unless fallthrough is deliberate.

mrannanj (approved):

PR #677, opened 2024-10-23
Remove `str` from textslider and use a fixed allocation

Vagabond:
There is no need for all the string dynamism in here, and its just slow and wasteful.


katajakasa:
Probably deprecated now ?

PR #683, opened 2024-10-25
`str` String fixes

Nopey:
- Strings of length 15 were being missed by the small string optimization,
- str_replace would always resize the string before replacing (potentially truncating the contents), and
- str_replace no longer falls into an infinite loop when replacement contains seek, and can find consecutive hits even with an empty replacement.

~~Also .gitignore'd the test artifacts, clangd compile_commands.json & .cache, and any untracked files added to /resources (typically the OMF 2097 assets).~~

### DIFF
--- a/testing/test_str.c
+++ b/testing/test_str.c
@@ -351,6 +351,21 @@ void test_str_replace_sm(void) {
     str_free(&d);
 }
 
+void test_str_replace_sm_regression(void) {
+    str d;
+    char long_str[STR_STACK_SIZE + 1];
+    memset(long_str, 'A', sizeof long_str);
+    long_str[sizeof long_str - 2] = 'B';
+    long_str[sizeof long_str - 1] = '\0';
+    str_from_c(&d, long_str);
+    CU_ASSERT_PTR_NOT_NULL(d.data);
+
+    str_replace(&d, "AA", "C", 1);
+    CU_ASSERT(str_c(&d)[d.len] == '\0');
+    CU_ASSERT(str_c(&d)[d.len - 1] == 'B');
+    str_free(&d);
+}
+
 void test_str_replace_multi(void) {
     str d;
     str_from_c(&d, "test $1 string $1");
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -15,3 +15,13 @@ mingw_build/
...
+.cache
+compile_commands.json
+resources
+Testing
+test.af
+test.bk
+test.gpl
+test.rec
+test.trn
+test2.gpl
```
katajakasa, line 27: It should not be required to ignore these all. Building and running should happen in a separate build directory.

Nopey: un-added test artifacts from gitignore :+1:

.gitignore
```diff
@@ -15,3 +15,6 @@ mingw_build/
 .vscode
 cmake-build-debug/
 cmake-build-release/
+.cache
+compile_commands.json
+/resources
```
katajakasa, line 20: The resources directory in the root should not be ignored either; it contains our own generated resources and their project files (and will probably receive more files as time goes on).

I'm not sure where .cache and compile_commands.json come from.

Nopey: .cache and compile_commands.json are for clangd
clangd expects compile_commands.json to be in a parent directory of your sourcefile (I create a symlink to the generated on in `build`).

will put the gitignore changes in a seperate PR

testing/test_str.c
```diff
@@ -26,6 +26,24 @@ void test_str_from(void) {
...
+void test_str_from_threshold(void) {
+    str d;
+
+    char short_str[STR_STACK_SIZE];
+    memset(short_str, 'A', sizeof short_str);
+    short_str[sizeof short_str - 1] = '\0';
+    str_from_c(&d, short_str);
+    CU_ASSERT_PTR_NULL(d.data);
+    str_free(&d);
+
```
katajakasa, line 38: Looks like a good place to split the test into two ?

PR #684, opened 2024-10-25
Gitignore clangd files

Nopey:
.cache and compile_commands.json are for use with clangd
https://clangd.llvm.org/design/compile-commands
https://stackoverflow.com/questions/51402740/visual-studio-code-clangd-extension-configuration#51692215

the build instructions BUILD.md states:
> On Linux in debug mode, the OMF resource files should be put in resources/ subdirectory.

To prevent these from being checked in accidentally (and generally ruining my git status), each file from omf2097-assets.zip is now gitignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ mingw_build/
 .vscode
 cmake-build-debug/
 cmake-build-release/
+.cache
+compile_commands.json


.gitignore
```diff
@@ -15,3 +15,61 @@ mingw_build/
 .vscode
 cmake-build-debug/
 cmake-build-release/
+.cache
+compile_commands.json
+
+resources/ALTPALS.DAT
```
katajakasa, line 21: Note that resources/ subdirectory in root is not the same directory that is discussed in the 
README. When compiling, you are supposed to do this:

```
mkdir build/
cd build/
cmake ..
make
```

This will build the openomf binary and tools and everything else into a fresh build directory. It will /also/ copy the openomf.bk to build/resources/, which is also where the assets should be put (NOT the root /resources). Building and running should never ever happen from the project root.

PR #685, opened 2024-10-25
Support vcpkg

Nopey:
to build with vcpkg, set the toolchain to vcpkg when configuring:
```
cmake -S. -Bbuild --toolchain=PATH_TO/vcpkg/scripts/buildsystems/vcpkg.cmake
```

FindXYZ cmake scripts are modified to use find_package(CONFIG) when vcpkg is detected, which will use each installed `XYZConfig.cmake` file.

I turned sdl2's default features off to make my life easier when developing on Void Linux, as sdl2's dbus feature then in turn default-depends on libsystemd, which pans out poorly on Void.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,7 +69,7 @@ if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
 endif()
 
 # Find OpenOMF core sources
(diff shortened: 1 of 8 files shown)


Nopey:
rolled into https://github.com/omf2097/openomf/pull/681

PR #686, opened 2024-10-25
Fix mismatched allocs/frees

Nopey:


src/utils/ringbuffer.c
```diff
@@ -24,8 +24,8 @@ void rb_create(ring_buffer *rb, size_t size) {
 void rb_free(ring_buffer *rb) {
     if(rb == NULL)
         return;
-    free(rb->data);
-    free(rb);
+    omf_free(rb->data);
+    // omf_free(rb); // we didn't allocate rb, it's probably on a stack somewhere.
```
katajakasa, line 28: Yep. This ringbuffer.c/.h was brought in from another project and I did a hasty convert to opneomf style. It's not yet used anywhere, so the free was forgotten.

Please remove the comments and the omf_free(rb) for cleanliness :)

PR #694, opened 2024-10-27
Bugfixes

Nopey:
Misc bugs that have been fixed:
- Player 2's Pilot name was hidden, rendering behind its background
- Newsroom was using the wrong font 
- Newsroom was using hardcoded pronouns and pilot names, which had incorrect case
- Newsroom wrapping didn't match original game. Unsure if this is entirely correct, but I think this is better.
 This PR wraps newsroom text at 31 characters (250px), and never wraps the fourth line:
```
Whoa, this challenger meant
business tonight. Shirro
could show the old pros a
thing or two about that Jaguar.
```

### DIFF
--- a/src/game/scenes/newsroom.c
+++ b/src/game/scenes/newsroom.c
@@ -68,9 +68,9 @@ void newsroom_fixup_str(newsroom_local *local) {
 
        1= Player1 - Crystal
        2= Player2 - Steffan
-       3= HAR1 - jaguar
-       4= HAR2 - shadow
-       5= Arena - Stadium. (DOS OMF always refers to the arena as Stadium)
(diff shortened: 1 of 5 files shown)


src/game/scenes/newsroom.c
```diff
@@ -20,14 +20,16 @@
 #define NEWSROOM_TEXT 87
 // there are 3*2 pronouns in total
 #define NEWSROOM_PRONOUN 81
+//
```
mrannanj, line 23: Did you mean to say something here? Add comment or remove it

mrannanj (changes requested):

mrannanj (approved):

PR #698, opened 2024-10-29
Buildsystem fixes, mostly for MSVC builds

Nopey:
vcpkg `x64-windows-static` triplet is now used to build the msvc release artifact, which statically links all dependencies.

argtable3's DllMain has been removed, because it shouldn't have one.
libepoxy's threadlocal windows GL dispatch logic has been disabled, as openomf only creates a single GL context-- and it relied on DllMain, which would require shipping libepoxy as a seperate DLL.


vcpkg-ports/argtable3/portfile.cmake
```diff
@@ -33,6 +33,15 @@ elseif(EXISTS "${CURRENT_PACKAGES_DIR}/lib/cmake/${PORT}")
     vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/${PORT})
 endif()
 
+if(NOT ARGTABLE3_REPLACE_GETOPT)
+    # fix dependency unofficial-getopt-win32
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/share/${PORT}/Argtable3Config.cmake" "# Generated CMake target import file."
+[[
```
mrannanj, line 39: This `[[` seems suspicious, does it not break anything?

vcpkg-configuration.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg-configuration.schema.json",
+  "overlay-ports": [
+    "./vcpkg-ports"
+   ]
+}
```
mrannanj, line 6: Could add newline to end of this file so it looks nice in git diffs.

mrannanj (approved):

PR #709, opened 2024-10-31
Fix libnatpmp vcpkg port

Nopey:
The link issue was because the MINIUPNP_STATICLIB and NATPMP_STATICLIB defines weren't being propogated from the dependency into openomf_core, which caused the headers to misbehave (and think they were part of a DLL build)

By linking to an imported cmake target we can be certain that any PUBLIC target_add_definitions in the libraries get to the openomf target we are linking them to, here openomf_core.

Unfortunately, [ubuntu's libminiupnpc-dev](https://packages.ubuntu.com/focal/amd64/libminiupnpc-dev/filelist) does not ship an exported cmake target (they are probably using the Makefile to build their package); if they did ship an exported cmake target, we would be able to rid ourselves of the Findminiupnpc.cmake script entirely and rely on find_package(CONFIG).

### DIFF
--- a/cmake-scripts/Findminiupnpc.cmake
+++ b/cmake-scripts/Findminiupnpc.cmake
@@ -1,3 +1,16 @@
+if(VCPKG_TOOLCHAIN)
+  if(TARGET miniupnpc::miniupnpc)
+    return()
+  endif()
+  find_package(miniupnpc CONFIG)
+  if(TARGET miniupnpc::miniupnpc)
+    MESSAGE(STATUS "Found miniupnp in vcpkg")
+    set(MINIUPNPC_LIBRARY miniupnpc::miniupnpc)
+    SET(MINIUPNPC_FOUND TRUE)
+    set(MINIUPNPC_INCLUDE_DIR)
+    return()
+  endif()
+endif(VCPKG_TOOLCHAIN)
 
 SET(MINIUPNPC_SEARCH_PATHS
     /usr/local/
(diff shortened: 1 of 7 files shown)


cmake-scripts/Findminiupnpc.cmake
```diff
@@ -1,3 +1,16 @@
...
+  endif()
+  find_package(miniupnpc CONFIG)
+  if(TARGET miniupnpc::miniupnpc)
+    MESSAGE(STATUS "Found miniupnp in vcpkg")
+    set(MINIUPNPC_LIBRARY miniupnpc::miniupnpc)
+    SET(MINIUPNPC_FOUND TRUE)
+    set(MINIUPNPC_INCLUDE_DIR)
+    return()
+  endif()
+endif(VCPKG_TOOLCHAIN)
```
mrannanj, line 13: This has uneven indentation compared to rest of the file, and also strange that `set` is sometimes in capital letters, sometimes not?

Nopey: Perhaps we should invest in setting up [cmake-format](https://github.com/cheshirekow/cmake_format)?
For now, this shim matches the style of the other find_package(CONFIG) shims

I'll make a seperate PR to master for cmake-format

Nopey: Opened  #710

mrannanj (approved):

Nopey:
note that the target branch of this PR is @Vagabond's `adt/nat-vpkg` branch, and not `master`.

PR #710, opened 2024-11-01
Enforce standard cmake formatting with cmake-format

Nopey:
The `--literal-comment-pattern .` argument is used to disable reflowing of comments, as it makes a mess of FindSDL2.cmake for little benefit.


katajakasa:
Quick note to documentation on how to set this up locally would be nice!

Nopey:
Added instructions to CONTRIBUTING.md, and put the literal_comment_pattern setting into a `.cmake-format.py` config file in case we ever have to change the settings further.

PR #711, opened 2024-11-01
Precompiled Headers & Faster CI

Nopey:
small extra change: renamed bktool's function bk_info bk_getinfo because `bk_info` is already the name of the struct bk_info_t typedef

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -25,9 +25,9 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Install clang-format and fd-find
-      run: |
-        sudo apt-get update
-        sudo apt-get -y install clang-format-${{ env.clang-format-version }} fd-find
+      uses: Eeems-Org/apt-cache-action@v1.3
+      with:
+        packages: clang-format-${{ env.clang-format-version }} fd-find
     - name: Run clang-format style check
       run: |
         fdfind . -e .c -e .h -X clang-format-${{ env.clang-format-version }} --dry-run --Werror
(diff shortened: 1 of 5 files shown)


tools/bktool/main.c
```diff
@@ -882,7 +882,7 @@ int main(int argc, char *argv[]) {
         } else if(keylist->count > 0) {
             bk_keylist();
         } else {
-            bk_info(&bk);
+            bk_getinfo(&bk);
```
mrannanj, line 885: This change seems to be maybe part of some other commit?

mrannanj (changes requested):

tools/bktool/main.c
Vagabond: it was renamed to not conflict with the struct of the same name.

Vagabond (approved):
This all looks reasonable to me

.github/workflows/compilation.yml
```diff
@@ -44,18 +44,18 @@ jobs:
       fail-fast: false
       matrix:
         config:
-        - { name: "gcc 14",   cc: gcc-14,   tidy: "On"  }
```
Vagabond, line 47: GCC has tidy turned off now?

tools/bktool/main.c
mrannanj: I don't object to the change, but it doesn't seem to be in the correct commit.

.github/workflows/compilation.yml
mrannanj: Well, I guess this is about clang-tidy, so it would be a mistake to have it on for gcc.

katajakasa (approved):

.github/workflows/compilation.yml
Nopey: I've moved the clang-tidy run from gcc unittests to clang unittests

clang-tidy can only consume the precompiled header files generated by a matching version of clang
Enabling Tidy in a gcc build will disable PCH and emit a warning (I added a check in cmakelists.txt)
This is also why I bumped clang from 16 to 18, to match our clang-tidy version.

tools/bktool/main.c
Nopey: Moved it into its own commit :+1:

mrannanj (approved):

PR #721, opened 2024-11-01
Don't statically link dependencies, yet

Nopey:
Until we replace dependencies that are LGPL licensed (libxmp, getopt-win32)
we shouldn't statically link our dependencies.

libopenmpt looks like a good alternative to libxmp, and getopt-win32 is a transitive dependency via argtable.

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -106,7 +106,7 @@ jobs:
         --toolchain vcpkg/scripts/buildsystems/vcpkg.cmake
         -DCMAKE_BUILD_TYPE="Release"
         -DVCPKG_BUILD_TYPE="release"
-        -DVCPKG_TARGET_TRIPLET="x64-windows-static"
+        -DVCPKG_TARGET_TRIPLET="x64-windows"
     - name: Build openomf
       run: cmake --build build-msvc --target openomf
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.21)
 project(OpenOMF C)
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake-scripts)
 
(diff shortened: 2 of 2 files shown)


katajakasa:
Last we tried, libopenmpt could not play the psm files correctly.

Vagabond:
there's always libdumb, too...

Nopey:
I just used OpenMPT 1.31.12.00 to export the PSM files to FLAC and can't find any issue
here's my export:
https://larsenml.net/OMF/

looked at libdumb, and it has been marked as deprecated in favor of libopenmpt afaict

Nopey:
fixed merge conflicts

Vagabond:
Can we selectively statically link things? getopt doesn't look avoidable on windows?

Vagabond:
https://github.com/kubo/ya_getopt/ looks like a LGPL alternative to getopt-win32

Vagabond:
I'd love to see all (L)GPL deps avoided, personally.

Nopey:
We can selectively static link some libs, see https://github.com/microsoft/vcpkg/discussions/33256

Requires overlaying our own vcpkg triplet, something like `x64-windows-static-except-lgpl`, and then selectively linking xmp and getopt-arg32 dynamically in there.

I don't think the problem of "how are we gonna static link our non-lgpl deps or replace the lgpl deps" needs to be solved to merge this PR, as this PR solves the license compliance issue.

Vagabond:
Yes, true, we can just stop statically linking for now.

Nopey:
Alternatively, we could just say that our (MSVC) binaries that static link LGPL libraries are LGPL licensed
seems mildly unexpected for an MIT licensed program to produce an LGPL licensed binary, though

also unsure what the implications for the n64 port are, probably similar situation to the msvc binaries.

Nopey:
we're also not shipping any of the third party licenses, which is a problem even for the permissively licensed libraries like SDL2 (zlib license)

Vagabond:
isn;t there a convention for satisfying those kinds of credit requirements in some kind of text file nowadays?

Nopey:
yeah, usually the third party licenses get thrown in a text file or webpage (for web services)
https://discord.com/licenses
https://github.com/ValveSoftware/source-sdk-2013/blob/master/thirdpartylegalnotices.txt

carstene1ns:
Just chiming in to say that libxmp is MIT-licensed for a while now.
Should be no problem to statically link it.

Nopey:
Oh good, just getopt-win32 then. Thanks, carstene1ns.

I should be able to open a PR to microsofts vcpkg repo tomorrow, to update the license spec there

Nopey:
The latest release of libxmp 4.6.0 was before the MIT relicense, so I'm not gonna upstream this to the microsoft/vcpkg repo
For now, I've made openomf's vcpkg port build use latest git libxmp.

Closing this PR in favor of #738

PR #725, opened 2024-11-02
Fix vga_state memory bugs

Nopey:
Also fixed off-by-one error in the assert.


katajakasa (approved):
Lol, i made the same changes to my own branch. Yep, this is correct, feel free to just merge.

PR #726, opened 2024-11-02
Stdin as console input

mrannanj:


src/engine.c
```diff
@@ -26,7 +26,7 @@ static int start_timeout = 30;
 static int enable_screen_updates = 1;
 static int debug_palette_number = 0;
 
-int engine_init(void) {
+int engine_init(bool headless) {
```
Vagabond, line 29: This probably should have stdbool.h included?

mrannanj: Well, that's possible, but it works for me and for the CI with the current. I can put it there sure.

mrannanj: Added stdbool.h

PR #728, opened 2024-11-02
Fixes for text dialog, kreissack dialog

Nopey:
label_create_with_width was abusing `text_find_max_strlen` thinking it produces the length of the longest line in the text after wrapping, when it in fact only produces the next line's length.

Kreissack's difficulty dialog is now wider than other dialogs, which is good
compare OMF 2097:
image

and OpenOMF (with these changes)
image

### DIFF
--- a/src/game/gui/dialog.c
+++ b/src/game/gui/dialog.c
@@ -6,7 +6,7 @@
 #include "video/video.h"
 #include <string.h>
 
-#define MAX_WIDTH 160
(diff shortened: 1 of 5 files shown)


Nopey:
image
this breaks scoreboard at the moment

Nopey:
image

mrannanj (approved):

katajakasa (approved):

PR #729, opened 2024-11-02
More renderer work 2

katajakasa:


Vagabond (approved):
Seems reasonable, and credits render properly

PR #730, opened 2024-11-02
Add transparent index and optimize font rendering with shadows.

nopjne:
Reduces texture switches when rendering fonts with shadows.
Introduces transparent index as an internal value supported by the entre sprite implementation.
Also puts all different transparent indices in one spot.


src/formats/palette.h
```diff
@@ -22,6 +22,28 @@
 extern "C" {
 #endif
 
+// All sprites have holes, these holes need to be transparent.
+#define SPRITE_TRANSPARENT_INDEX 0
```
katajakasa, line 26: Instead of naming these per purpose, could we instead have reusable defined ?

#define DEFAULT_TRANSPARENCY 0
#define NO_TRANSPARENCY -1

and then the special, specific ones (90, D0, etc). That would potentially let us reuse the generic ones in other code, too.

src/formats/sprite.c
```diff
@@ -310,6 +313,9 @@ int sd_sprite_vga_decode(sd_vga_image *dst, const sd_sprite *src) {
                     uint8_t b = src->data[i];
                     int pos = ((y * src->width) + x);
                     dst->data[pos] = b;
+                    if (b == transparent) {
+                        assert(false);
+                    }
```
katajakasa, line 318: assert(b != transparent)

src/formats/vga_image.c
```diff
@@ -1,11 +1,19 @@
+#if !defined(N64_BUILD) && !defined(MIN_BUILD)
 #include <png.h>
+#endif
```
katajakasa, line 3: Hum. Could we move the png reading from this file to utils/png_reader.h ? Then we could import the png headers etc. there, and also do the png support enable/disable there.

Then the functions, e.g. png_write_paletted() could just return "OK" if there was no support :)

We could also support the png files completely in cmake, then ? Probably.

src/game/gui/gauge.c
```diff
@@ -4,6 +4,7 @@
 #include "utils/miscmath.h"
 #include "video/surface.h"
 #include "video/video.h"
+#include "formats/palette.h"
```
katajakasa, line 7: Is this used ?

src/formats/palette.h
```diff
@@ -22,6 +22,28 @@
...
+
+// Backgrounds have no transparent pixels and need all pixels to be rendered.
+#define BACKGROUND_TRANSPARENT_INDEX -1
+
+// Force an opaque menu for now.
+#define MENU_TRANSPARENT_INDEX -1
+
+// Portraits have a transparancy that needs to offset past other items in the global palette.
+#define PORTRAIT_TRANSPARENT_INDEX 0xD0
```
katajakasa, line 35: Hmmh. These specific ones should probably be in the files where they are used. They are not generic, after all.

src/video/surface.c
```diff
@@ -1,48 +1,54 @@
...
+#include <SDL.h>
 
 // Each surface is tagged with a unique key. This is then used for texture atlas.
 // This keeps track of the last index used.
 static unsigned int guid = 0;
 
-void surface_create(surface *sur, int w, int h) {
+void surface_create(surface *sur, int w, int h, int transparent) {
+#ifdef N64_BUILD
+    sur->data = malloc_uncached_aligned(64, w * h);
```
katajakasa, line 17: Instead of making different reservation for each platform, we should probably have our own omf_malloc_uncached_aligned() or somesuch. Then x86 could just map to alloc, while N64 could map to malloc_uncached_aligned. That would also satisfy the requirement to have memory allocation wrapped in our omf_prefixed namespace.

katajakasa (changes requested):

src/formats/palette.h
```diff
@@ -22,6 +22,28 @@
...
+
+// Portrait highlights need a special offset for transparency to not conflict with other entries in the global palette.
+#define HIGHLIGHT_TRANSPARENT_INDEX 90
```
katajakasa, line 38: Actually, why does HIGHLIGHT and PORTRAIT need other transparency from 0 ?

src/formats/palette.h
nopjne: It was like that originally and it was very painful to find all of them. I would really like them to be in 1 spot.

src/game/gui/gauge.c
nopjne: Palette has the definition for transparency indices.

src/formats/vga_image.c
nopjne: It is okay if that becomes a tracked issue for now?
I don't want to get into anything cmake for now.

src/formats/sprite.c
nopjne: That would add a function call for each iteration, not great for perf.

katajakasa: assert is automatically stripped out in release builds. It's safe to use performance-wise.

src/formats/palette.h
nopjne: I can define DEFAULT_TRANSPARENCY and NO_TRANSPARENCY and set the defines that are currently 0 and -1 to the new defines. But would like to keep them separately and purposefully named when they are used in the c files.
This helps me to differentiate them, and it's easier to debug them when they are defined individually and are all stored in one place.

src/formats/palette.h
nopjne: It breaks for me with different transparency indices. Something to look into later.

src/video/surface.c
nopjne: It's slight complicated and not sure if we need to deal with it. The N64 is a UMA (unified memory architecture) to some extent, so memory allocated can be used as a texture, but ofcourse there are some requirements on alignment because the hardware has less than 32 address bits.

So here it makes sense for surfaces, other platforms will be able to use malloc normally unless they want to allocate.
The problem is also that libdragon requires a matching aligned alloc to have a matching aligned free. Becasue it stores some header info somewhere.

Again very specific, I would wait until we have another platform that needs something different.

I also don't see x86 allocating GPU memory here because then all accesses would need to have a map/unmap call.

src/formats/sprite.c
nopjne: I don't want to impact the debug builds too much either.

katajakasa: assert is a macro. This will (in theory) leave a condition to all release builds, which will slow them down. If you put the condition inside the assert macro, it will get removed and the speed will not change to either direction in debug builds.

src/formats/vga_image.c
katajakasa: I don't think we need to do the cmake thing now, but I would definitely still move the ifdeffing to the separate png-related files.

src/video/surface.c
katajakasa: x86 would not allocate gpu memory. X86 would just use normal malloc(). The point is, we would have the omf_malloc_uncached_aligned(), which would only do something specific on platforms that need it to do something specific. Then we could use that omg_malloc_uncached_aligned() only on sprites and stuff that need it.

The issue with waiting until someone else needs it, is that then the response to this same query will be "But it's already done like this, and I don't know about n64, and it's too much work" :(

src/formats/palette.h
katajakasa: Ok. As discussed on discord, let's do tranparency.h or somesuch.

src/formats/pcx.c
```diff
@@ -141,19 +143,21 @@ int pcx_load_font(pcx_font *font, const char *filename) {
     return SD_SUCCESS;
 }
 
-int pcx_font_decode(const pcx_font *font, sd_vga_image *o, uint8_t ch, int8_t palette_offset) {
+int pcx_font_decode(const pcx_font *font, sd_vga_image *o, uint8_t ch, int8_t palette_offset, uint8_t transparent) {
     if(ch >= font->glyph_count || font == NULL || o == NULL) {
         return SD_INVALID_INPUT;
     }
 
+    assert((transparent >= 0) && (transparent < 256));
```
katajakasa, line 151: "transparent" is uint8_t. It cannot be anything else except 0-256.

src/video/surface.c
Vagabond: Right, you tag the allocation as potentially needing special handling, and each platform can decide if it needs to do anything special.

nopjne: I think what would make sense is surface_create, surface_free, sd_vga_image_create,sd_vga_image_free move to platform.c.

platform.c will have a different implementation based on the platform. 
I will clean that up when the N64 files start coming in.

Vagabond:
Merging #732 broke this PR a bit. I wonder if you'd be open, @nopjne to me working on some code to separate the text flowing from the text rendering. So flowing some text would result in an array of surface pointers and the offsets to render them at. Then you could do a 2 pass render to render the shadows first and get the same optimization as you had here, but with the benefit of not having to reflow text each frame?

src/video/surface.c
katajakasa: No, those functions should really stay where they are, as that is where the devs will look for them and that is what makes sense. There is no point in moving them anyway, because it's not the right abstraction level. Abstracting malloc is.

nopjne: Then this function will still need an ifdef N64_BUILD. Because the alignment is platform specific.

nopjne:
> Merging #732 broke this PR a bit. I wonder if you'd be open, @nopjne to me working on some code to separate the text flowing from the text rendering. So flowing some text would result in an array of surface pointers and the offsets to render them at. Then you could do a 2 pass render to render the shadows first and get the same optimization as you had here, but with the benefit of not having to reflow text each frame?

I was sort of hoping 732 would be merged after this PR was done, but it is what it is. 

I also have a change that buffers the flowing of text, please take a look at the text_object in the n64 if you plan on working on this.
It's a big change I don't really want to do much merging on. Most important parts are ability to prerender letters to a texture, and allow a way to invalidate the cache. Also my change allows declaring the text_object as frame_over_frame copy safe, and I make heavy use of it in the N64 branch where I copy the rectangle from the previous frame, without having to allocate memory for a text_object.

src/formats/fonts.c
```diff
@@ -57,18 +58,20 @@ int sd_font_save(const sd_font *font, const char *file) {
...
+    assert((transparent >= 0) && (transparent < 256));
+
     int t = 0;
     for(unsigned i = 0; i < font->h; i++) {
         for(int k = font->h - 1; k >= 0; k--) {
             if(font->chars[ch].data[i] & (1 << k)) {
                 o->data[t] = color;
             } else {
-                o->data[t] = 0;
+                o->data[t] = (char)transparent;
```
Nopey, line 74: shouldn't the transparent argument be a `uint8_t` like color? that would remove the need for the assert and the cast.

Nopey: hehe, just saw katajakasa's comment on some similar code below
unsure why it was marked as resolved, though

nopjne: Transparent can be -1 to indicate that the background is not transparent. For sd_font_decode that is not possible. Therefore the check is in place. To prevent this from breaking.

src/formats/transparent.h
```diff
@@ -0,0 +1,44 @@
...
+// Force an opaque menu for now.
+#define MENU_TRANSPARENT_INDEX DEFAULT_NOT_TRANSPARENT
+
+// Portraits have a transparancy index that needs to offset past other items in the global palette.
+// TODO: Should really become default.
+#define PORTRAIT_TRANSPARENT_INDEX 0xD0
+
+// Portrait highlights need a special offset for transparency to not conflict with other entries in the global palette.
+// TODO: Should really become default.
+#define HIGHLIGHT_TRANSPARENT_INDEX 90
```
Vagabond, line 34: This is not needed

src/game/scenes/melee.c
```diff
@@ -728,8 +730,7 @@ int melee_create(scene *scene) {
 
     // Create a black surface for the highlight box. We modify the palette in renderer.
     unsigned char *black = omf_calloc(1, 51 * 36);
-    surface_create_from_data(&local->select_hilight, 51, 36, black);
-    surface_set_transparency(&local->select_hilight, -1);
+    surface_create_from_data(&local->select_hilight, 51, 36, black, HIGHLIGHT_TRANSPARENT_INDEX);
```
Vagabond, line 733:
```suggestion
    surface_create_from_data(&local->select_hilight, 51, 36, black, DEFAULT_NOT_TRANSPARENT);
```

src/utils/allocator_default.h
```diff
@@ -1,10 +1,14 @@
 #ifndef ALLOCATOR_DEFAULT_H
 #define ALLOCATOR_DEFAULT_H
```
Vagabond, line 3: Should you not define allocator_n64 instead?

nopjne: Not sure, I was suggesting a platform.c but that was shutdown.

nopjne: Keeping the implementation in a header file, doesn't allow me to override this for just the N64.

katajakasa: 1. Copy allocator_default.h to allocator_n64.h
2. Make allocator_n64.h handle allocations in n64 style
3. Add ifdef to allocator.h to use allocator_n64.h instead (where there is comment "Add ifdefs here to include platform-specific allocators"). 
4. Import allocator.h

This should work fine ?

src/formats/transparent.h
katajakasa: Are any of these actually needed? All the sprites already get these values by default, and our current code already informs the sprite on what the default is. I don't understand why we need any of these ?

nopjne: We keep coming back to this discussion :) 
The transparency index was a bit of mess, and although it slightly improved, it is still is a hack.
This change consolidates the hack by disallowing "sidesets" and modification of the transparency outside of initialization.

I don't agree with "it's already set by default". It's not, there are issues with 0xD0 for portraits. There is a -1 and 0.
From the code before, it was also hard to track down all the locations where transparency was being populated. This change puts it all in 1 place.

I also maintain transparency should be able to move around, per platform.

katajakasa: How would you move the transparency around per platform ?

On the -1 value, that should be the renderers problem to decide what to do with -1. On GL renderer, it's just "no transparency", on other renderers it can mean whatever they want it to mean. 

Moving the value setting to the constructor is fine, I don't care either way.

src/utils/allocator_default.h
```diff
@@ -35,4 +39,16 @@ static inline void *omf_realloc_real(void *ptr, size_t size, const char *file, i
     abort();
 }
 
+static inline void *omf_alloc_with_options_real(size_t nmemb, size_t size, int options, const char *file, int line) {
```
katajakasa, line 42: Do we need the options ? Could this just be omF_alloc_surface_data() or somesuch ?

src/formats/transparent.h
nopjne: > How would you move the transparency around per platform ?
It would be an ifdef in the transparent.h

> On the -1 value, that should be the renderers problem to decide what to do with -1. On GL renderer, it's just "no transparency", on other renderers it can mean whatever they want it to mean.
When a sprite is currently "decompressed" it's filled with the transparent index (0 default), a platform could choose to select 8 as the transparent index or 255, as long as that value doesn't clash with any of the already present values in the sprite data.

src/utils/allocator_default.h
nopjne: It could, though keeping it more abstract allows us to reduce the number of functions in the API and changes don't need to touch the common code if adding new options. It also allows implementations to ignore new options when they are added, instead of having to implement the missing functions.

katajakasa: I have three issues with that:
1. We don't currently know of any cases that need more flags, making this an unnecessary complication
2. Free function also requires the same flags as the alloc, which us ugly
3. Why do we need a bitmask ? malloc cases tend to be pretty specific anyways, making separate functions clearer in the long run.

src/formats/transparent.h
katajakasa: Why would a platform require other transparency indexes ?

src/utils/allocator_default.h
nopjne: > 1. We don't currently know of any cases that need more flags, making this an unnecessary complication

Any allocation that could be used by hardware directly, will need to go through alloc with options.

> 2. Free function also requires the same flags as the alloc, which us ugly

This is more of a stylistic preference why is a set of flags and 1 free-function uglier than a set of free-functions names.

> 3. Why do we need a bitmask ? malloc cases tend to be pretty specific anyways, making separate functions clearer in the long run.

Allocation parameters may be different, cached/uncached, aligned/unaligned, size-extended/non-extended, different alginment requirements for different allocation types.
My gripe with separate functions is that you will have to rev all your platforms every time a new function is added and most platforms won't do anything special or different. So just increasing the merge conflict size, when multiple people are working on that code.

src/formats/transparent.h
nopjne: Optimization. Right now nuking index zero (for transparency) of the palette on the N64 actually requires a copy of the surrounding 3 palette entries. Putting these somewhere outside of the sprite's palette simplifies the transparency nuking, as I can nuke the surrounding palette entries with uninitialized memory and forego the read modify write.

src/utils/allocator_n64.h
```diff
@@ -0,0 +1,65 @@
+#ifndef ALLOCATOR_DEFAULT_H
+#define ALLOCATOR_DEFAULT_H
+
+#include <assert.h>
+#include <stddef.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+// Allocator options
+#define ALLOC_HINT_TEXTURE 0x00000001
```
Nopey, line 10: unrelated to the bitflags conversation: If `ALLOC_HINT_TEXTURE` is to be used in cross-platform code, it should not be defined here, but in allocator.h.

src/utils/allocator_n64.h
```diff
@@ -0,0 +1,65 @@
...
+        return ret;
+    fprintf(stderr, _text_realloc_error, ptr, size, file, line);
+    abort();
+}
+
+static inline void *omf_alloc_with_options_real(size_t nmemb, size_t size, int options, const char *file, int line) {
+    void *ret = NULL;
+    if((options & ALLOC_HINT_TEXTURE) != 0) {
+        assertf(size != 0, "texture size was 0");
+        ret = malloc_uncached_aligned(64, size);
```
Nopey, line 46: malloc(0) is generally valid, although not a great use of CPU time.
if libdragon's malloc_uncached_aligned explodes when size is 0, perhaps we should return NULL here?
if malloc(0) is not a bug, perhaps a warning is more fitting than an assert? both compile to nothing in release builds.

Nopey: just noticed allocator.h declares hardly anything
this is unfortunate and not this PR's fault

I'd prefer if allocator.h defined all the symbols that users of allocator.h are supposed to use: `omf_free`, etc..
`omf_realloc_real` and similar (omf_free_real) can continue to be only called in allocator.h and only defined in allocator_platform.h, as they're not public symbols.

src/utils/allocator_default.h
Nopey: bitflags are unsifficient for arbitrarily aligned allocations and allocate_at_least style allocations.

Arbitrarily (un)aligned allocations do not map particularly well to bitflags, as the caller needs to control to what the allocation is aligned-- 32-byte alignment for AVX, 16-byte for SSE, etc.. I'd expect any aligned allocator we provide to be similar to [aligned_alloc](https://en.cppreference.com/w/c/memory/aligned_alloc).

size-extended allocations.. sounds like when vector.h asks for an allocation of size 12 (for 3 4-byte elements), and the allocator decides to round up to some common size like 16-- the allocator must have a way of passing the size of the allocation to an interested caller (see C++'s [allocate_at_least](https://en.cppreference.com/w/cpp/memory/allocator_traits/allocate_at_least)), preferrably without the overhead of storing alloc size in every allocation (as glibc does for malloc_usable_size).

src/utils/allocator_n64.h
```diff
@@ -0,0 +1,65 @@
...
+
+static inline void *omf_realloc_real(void *ptr, size_t size, const char *file, int line) {
+    void *ret = realloc(ptr, size);
+    if(ret != NULL)
+        return ret;
+    fprintf(stderr, _text_realloc_error, ptr, size, file, line);
+    abort();
+}
+
+static inline void *omf_alloc_with_options_real(size_t nmemb, size_t size, int options, const char *file, int line) {
```
Nopey, line 42: Please document the requirement that the return value of omf_alloc_with_options_real not be passed to vanilla omf_free,
preferrably in allocator.h and not allocator_n64.h

CMakeLists.txt
```diff
@@ -38,6 +38,8 @@ if(MSVC)
     string(REGEX REPLACE "/W3" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
     # we would probably like to pass /Wall instead of /W4, but there's a lot
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /permissive- /W4")
+    #LibPNG is not linking on MSVC correctly, disable it.
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DMIN_BUILD")
```
Nopey, line 42: Please do not. see also https://github.com/omf2097/openomf/pull/769

nopjne: This was temporary you can take it out when 769 is merged. Or I'll fix it if 769 goes in first.

src/game/gui/text_render.c
```diff
@@ -318,6 +321,29 @@ static void text_render_len(const text_settings *settings, text_mode mode, int x
             }
 
             render_char_shadow_surface(settings, sur, mx + start_x, my + start_y);
```
Nopey, line 323: Please update the comment on line 317 that reads: `// Render character`
as this loop does not render the character, just the shadows.

src/utils/allocator_default.h
nopjne: > bitflags are unsifficient for arbitrarily aligned allocations and allocate_at_least style allocations.

I've thought about this too and initially thought it wouldn't be sufficient but warmed up to the idea of letting the platform decide on the allocation type based on the incoming flags. 
The flags are very high level but allow the platform to make a decision if it should be any of the above types. While alloc_aligned may make sense on one platform it doesn't have to make sense on another and can be unaligned somewhere else.

src/utils/allocator_n64.h
nopjne: > if libdragon's malloc_uncached_aligned explodes when size is 0, perhaps we should return NULL here?

It does blow up with size 0, but I am adamant that we should not be allocating textures of size 0 and would like those to be caught and fixed.

src/utils/allocator.h
```diff
@@ -5,11 +5,30 @@ extern const char *_text_malloc_error;
...
+// hardware directly. These allocations can be textures, palettes, audio buffers or pointers to read only memory.
+// The options flags are hints as to what type of allocation is necessary, it is up to the platform's implementation to
+// decide how to translate the flags: what alignment is necessary to be optimal and whether memory needs to be allocated
+// cached or uncached.
+// Using this function can allow the platform to save on copies and translations in the platform hardware implementation
+// layers.
+//
+// NOTE: An allocation with a specific set of options also requires a matching free with the same set of options.
+//
+#define omf_alloc_with_options(nmemb, size, options)                                                                   \
```
katajakasa, line 32: If the options is going to declare different types of allocations, then what happens if ALLOC_HINT_TEXTURE and e.g. ALLOC_HINT_SOUND are defined simultaneously ? This is a very dangerous thing for undefined behavirour. Perhaps the flags should be an enum instead.

src/formats/transparent.h
Vagabond: This seems like a valid point, and doesn't cost us all that much to support.

src/utils/allocator.h
Vagabond: It's probably worth thinking about what allocator flags we'd actually want to support. Obviously some combinations are nonsensical (like TEXTURE | SOUND) but we could also add a simple check, in debug mode, for invalid flag combinations, if we wished.

katajakasa: Srsly, it might be best to split this to functions by purpose, and then allow more detailed flags for those as required. So malloc_texture(amount, flags), malloc_sound(amount, flags) etc. That would prevent nonsensical combinations, and just reserve the options flag for fine-grained changes.

Vagabond: the other use, other than n64 support, that I can think of would be an arena style allocator, where we could indicate whether an allocation should live for the life of the scene (in the scene arena allocation), in a global/engine arena or should be manually managed (a more ephemeral object). Then we could use the hints to track the kind of allocation we're doing.

src/formats/transparent.h
katajakasa: Ok. I feel the reasoning is thought through. I'll resolve this one.

src/utils/allocator.h
katajakasa: I'd really like to use some arena allocator also. If someone has the interest, I would be all for integrating one.

PR #732, opened 2024-11-02
Split text renderer into more pieces

katajakasa:
Split text renderer a bit. Now we fetch the surface only once, and there is additional renderer function for rendering the text type (which already knows its own length).

### DIFF
--- a/src/game/gui/text_render.c
+++ b/src/game/gui/text_render.c
@@ -26,42 +26,45 @@ void text_defaults(text_settings *settings) {
     settings->strip_trailing_whitespace = false;
 }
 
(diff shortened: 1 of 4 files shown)


Vagabond:
No strong objection, although I don't know how much this saves us in efficency. I do think that separating flowing from rendering will be a much bigger improvement, and this does help with that, slightly.

katajakasa:
> No strong objection, although I don't know how much this saves us in efficency. I do think that separating flowing from rendering will be a much bigger improvement, and this does help with that, slightly.

This is more of a cleanliness thing. It felt like a good individual piece for one PR :)

PR #737, opened 2024-11-03
Towards Monolithic OpenOMF language files

Nopey:
Currently, this PR:
- Generates the LANG2_STR enum from a list of language strings known as `LanguageStrings.cmake`
- Copies strings listed in `LanguageStrings.cmake` from ENGLISH and GERMAN.DAT to their OpenOMF specific language file

The unobtained end goals of this PR:
- Replace all uses of lang_get + hardcoded value with lang2_get + an enum value
- Ship a single localization file per language (not counting tournament localization, that's another PR)
- Reduce translator burden by not shipping unused OMF 2097 language strings
- Reduce runtime memory usage by only retaining the useful language strings' `data`, and not their `description`.


katajakasa:
This seems messy and difficult. Would it be better to just create our own language files, add only text that is used and include them to the project? Then we could drop the original language files completely.

Nopey:
Added "Add language string for Text Encoding", so there's something signalling whether the .LNG is in utf-8 or cp437
Rebased on master to resolve merge conflict with stdbool-everywhere pr.

Intentionally not removing the ENGLISH.DAT + scripts we wrote -> ENGLISH.LNG pipeline, because despite its complexity it allows us to avoid commiting OMF assets into the repo.

Should I remove the Dansk Maskinoversottelse? its a derivative of ENGLISH.DAT

Nopey:
the `resources/translations` submodule is checked out to https://github.com/omf2097/openomf-translations/pull/1

src/game/scenes/arena.c
```diff
@@ -1539,7 +1540,9 @@ int arena_create(scene *scene) {
             local->rec->pilots[i].info.color_1 = player->pilot->color_1;
             local->rec->pilots[i].info.color_2 = player->pilot->color_2;
             local->rec->pilots[i].info.color_3 = player->pilot->color_3;
-            memcpy(local->rec->pilots[i].info.name, lang_get(player->pilot->pilot_id + 20), 18);
+            // XXX ugly strncpy, implicit truncation, no nul termination guarantee, why are we even copying the pilot's
+            // name in here?
+            strncpy(local->rec->pilots[i].info.name, lang_get_offset(LangPilot, player->pilot->pilot_id), 18);
```
mrannanj, line 1543: There is that `strncpy_or_truncate` that could used now.

src/game/scenes/mechlab/lab_menu_main.c
```diff
@@ -144,7 +144,8 @@ void lab_menu_focus_arena(component *c, bool focused, void *userdata) {
         sd_chr_enemy *enemy = mechlab_next_opponent(s);
         if(enemy) {
             char tmp[100];
-            snprintf(tmp, 100, lang_get(537), enemy->pilot.name);
+            // XXX lang string provided format string
```
mrannanj, line 147: What's the XXX for? Are there safety problems?

Nopey: Yes, and I'd like users to be free to install untrusted game data from the internet with less risk.
Untrusted format strings can do terrible things (stack corruption -> rop -> ace), and are inflexible in the order of arguments.
Later (not in this PR), i want all %s's to be replaced with newsroom-style substitutions.

src/game/scenes/arena.c
Nopey: Added and using strncpy_or_assert, which asserts on truncation.

mrannanj (approved):
Still undecided about submodules, but initially okay.

PR #738, opened 2024-11-03
Remove LGPL dependencies from vcpkg port

Nopey:
getopt-win32: removed
libxmp: updated to the latest git revision, which has been relicensed as MIT.
Latest release libxmp 4.6.0 was before the relicense, so I'm just gonna avoid that one.

Obsoletes PR #721


vcpkg-ports/argtable3/portfile.cmake
```diff
@@ -5,14 +5,13 @@ vcpkg_from_github(
...
         "Remove-DllMain.patch"
 )
 
 set(ARGTABLE3_REPLACE_GETOPT ON)
-if(VCPKG_TARGET_IS_WINDOWS AND NOT VCPKG_TARGET_IS_MINGW)
-   set(ARGTABLE3_REPLACE_GETOPT OFF)
-endif()
+#if(VCPKG_TARGET_IS_WINDOWS AND NOT VCPKG_TARGET_IS_MINGW)
+#   set(ARGTABLE3_REPLACE_GETOPT OFF)
+#endif()
```
mrannanj, line 14: Why comment out instead of removing these?

Nopey: no reason, removed them.

mrannanj (approved):

Vagabond:
Why don't we need getopt anymore?

Nopey:
We're using argtable3's own implementation by setting ARGTABLE3_REPLACE_GETOPT

PR #740, opened 2024-11-03
fix newsroom text area transparency

katajakasa:


mrannanj (approved):
Tested that this works.

PR #748, opened 2024-11-05
Impl video_reinit so video_menu takes effect right away

Nopey:


katajakasa (approved):
Nice! That's one thing off from my renderer todo list :)

Issue #702, opened 2024-10-30
Flail move deadlock when executing destruction move


nopjne:
When Flail executes a destruction move the match deadlocks. 

https://github.com/user-attachments/assets/2ddea512-73b1-42c4-b105-7612a74c43ac

Vagabond:
Fixed in #749

PR #749, opened 2024-11-05
Fix flail destruct deadlock #702

Vagabond:


src/game/objects/har.c
```diff
@@ -465,6 +465,11 @@ void har_move(object *obj) {
     // Check for wall hits
     if(obj->pos.x <= ARENA_LEFT_WALL || obj->pos.x >= ARENA_RIGHT_WALL) {
         h->is_wallhugging = 1;
+        if(player_frame_isset(obj, "cw") && player_frame_isset(obj, "d")) {
+            DEBUG("disabling d tag on animation because of wall hit");
+            obj->animation_state.disable_d = 1;
+        }
```
katajakasa, line 471: probably not correct, but better than before.

katajakasa (approved):

PR #757, opened 2024-11-06
Slightly better Newsroom Captures

Nopey:
New:
1. Newsroom shows robots in color
2. Newsroom "final blow" capture will zoom out to show both robots as needed
3. Arena overlay UI is now hidden in captures (the round token objects are unfortunately still present)

1 and 2 are for parity with the original game, 3 is extra.

future work:
- the original game's seamless transition from arena to newsroom
- improve camera positioning + sizing logic
- hide objects in captures, chiefly the round token objects.
- shadow's shadows currently all show up in color, haven't checked what the original does

image

screenshot_20241106_122744

### DIFF
--- a/src/game/utils/har_screencap.c
+++ b/src/game/utils/har_screencap.c
@@ -51,6 +51,8 @@ void har_screencaps_capture(har_screencaps *caps, object *obj, object *obj2, int
         caps->ok[id] = false;
     }
 
+    game_state *gs = obj->gs;
+
     // Position
     vec2i pos = camera_position_for(obj);
     vec2i size = {SCREENCAP_W, SCREENCAP_H};
@@ -62,6 +64,12 @@ void har_screencaps_capture(har_screencaps *caps, object *obj, object *obj2, int
         pos.x = min2(pos.x, pos2.x);
         pos.y -= (size.y - SCREENCAP_H) / 2;
     }
+
+    video_render_prepare();
+    gs->hide_ui = true;
+    game_state_render(gs);
+    gs->hide_ui = false;
+    video_render_finish_offscreen();
 
     // Capture
     video_area_capture(&caps->cap[id], pos.x, pos.y, size.x, size.y);
--- a/src/game/utils/har_screencap.c
+++ b/src/game/utils/har_screencap.c
@@ -31,25 +31,40 @@ int har_screencaps_clone(har_screencaps *src, har_screencaps *dst) {
     return 1;
 }
 
-void har_screencaps_capture(har_screencaps *caps, object *obj, int id) {
-    if(caps->ok[id]) {
-        surface_free(&caps->cap[id]);
-        caps->ok[id] = false;
-    }
-
-    // Position
+static vec2i camera_position_for(object *obj) {
     vec2i size = object_get_size(obj);
     vec2i pos = object_get_pos(obj);
     int x_margin = (SCREENCAP_W - size.x) / 2;
     int y_margin = (SCREENCAP_H - size.y) / 2;
     int x_center = pos.x - size.x / 2;
     int y_center = (NATIVE_H - pos.y);
-    int x = clamp(x_center - x_margin, 0, NATIVE_W - SCREENCAP_W);
-    int y = clamp(y_center + y_margin, 0, NATIVE_H);
+
+    vec2i res;
+    res.x = clamp(x_center - x_margin, 0, NATIVE_W - SCREENCAP_W);
+    res.y = clamp(y_center + y_margin, 0, NATIVE_H);
+    return res;
+}
+
+void har_screencaps_capture(har_screencaps *caps, object *obj, object *obj2, int id) {
+    if(caps->ok[id]) {
+        surface_free(&caps->cap[id]);
+        caps->ok[id] = false;
+    }
+
+    // Position
+    vec2i pos = camera_position_for(obj);
+    vec2i size = {SCREENCAP_W, SCREENCAP_H};
+    if(obj2) {
+        // pan out when bots are apart
+        vec2i pos2 = camera_position_for(obj2);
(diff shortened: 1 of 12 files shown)


src/video/video.c
```diff
@@ -169,12 +169,17 @@ static void video_set_blend_mode(object_array_blend_mode request_mode) {
 }
 
 void video_area_capture(surface *sur, int x, int y, int w, int h) {
+    // XXX hack: round up read width to nearest multiple of 4 for glReadPixels call
+    // could someone who knows GL better replace this magic number 4 with something from the API?
+    int bufw = (w + 3) / 4 * 4;
```
katajakasa, line 174: Why this ?

Nopey: On my computer (NVidia, Windows), calling glGetPixels with widths that aren't multiples of 4 produces output with weird stride:
image

katajakasa: I see. Try setting glPixelStorei(GL_PACK_ALIGNMENT, 1);

src/video/video.c
```diff
@@ -196,9 +201,29 @@ static inline void set_screen_viewport(void) {
...
+    activate_program(g_video_state.palette_prog_id);
+    render_target_activate(g_video_state.target);
+
+    object_array_blend_mode mode;
+    while(object_array_get_batch(g_video_state.objects, &batch, &mode)) {
+        video_set_blend_mode(mode);
+        object_array_draw(g_video_state.objects, &batch);
+    }
+
+    // Disable render target, and dump its contents as RGBA to the screen.
```
katajakasa, line 220: Comment is now half wrong

src/video/video.c
```diff
@@ -196,9 +201,29 @@ static inline void set_screen_viewport(void) {
     glViewport(vp_x, vp_y, g_video_state.viewport_w, g_video_state.viewport_h); // This is used for screen shakes.
 }
 
+void video_render_finish_offscreen(void) {
+    object_array_finish(g_video_state.objects);
+
```
katajakasa, line 206: This is now rendering with palette and remaps that is one frame late.

src/video/video.c
```diff
@@ -196,9 +201,29 @@ static inline void set_screen_viewport(void) {
...
+    render_target_activate(g_video_state.target);
+
+    object_array_blend_mode mode;
+    while(object_array_get_batch(g_video_state.objects, &batch, &mode)) {
+        video_set_blend_mode(mode);
+        object_array_draw(g_video_state.objects, &batch);
+    }
+
+    // Disable render target, and dump its contents as RGBA to the screen.
+    render_target_deactivate();
```
katajakasa, line 221: deactivate should be right before the RGBA phase, since it resets any and all FBO selection.

src/game/utils/har_screencap.c
```diff
@@ -31,27 +31,51 @@ int har_screencaps_clone(har_screencaps *src, har_screencaps *dst) {
...
+        size.y = SCREENCAP_H * size.x / SCREENCAP_W;
+        pos.x = min2(pos.x, pos2.x);
+        pos.y -= (size.y - SCREENCAP_H) / 2;
+    }
+
+    video_render_prepare();
+    gs->hide_ui = true;
+    game_state_render(gs);
+    gs->hide_ui = false;
+    video_render_finish_offscreen();
```
katajakasa, line 72: This is missing half the blending work, since it's done in the RGBA step. You may need to add another render phase for the remapping work, and reduce RGBA part to just palette search step.

Nopey: The har_screencap (via video_area_capture) captures the indexed colors from the offscreen framebuffer

src/video/video.c
```diff
@@ -169,12 +169,17 @@ static void video_set_blend_mode(object_array_blend_mode request_mode) {
...
     render_target_activate(g_video_state.target);
-    unsigned char *buffer = omf_calloc(1, w * h);
-    glReadPixels(x, y, w, h, GL_RED, GL_UNSIGNED_BYTE, buffer);
-    surface_create_from_data_flip(sur, w, h, buffer);
+    unsigned char *buffer = omf_calloc(1, bufw * h);
+    glReadPixels(x, y, bufw, h, GL_RED, GL_UNSIGNED_BYTE, buffer);
+    surface_create_from_data_flip_stride(sur, w, h, bufw, buffer);
     surface_set_transparency(sur, -1);
     omf_free(buffer);
+    render_target_deactivate();
```
katajakasa, line 182: render_target_deactivate() should only be called when we're preparing to render to the screen. Otherwise it's an unnecessary GL api call.

src/game/utils/har_screencap.c
katajakasa: No, it only captures the red channel with the current direct palette indexes. You will need to copy the whole RGBA color set for that. In that case though, the arguments in GBA channels may no longer match the required remapping data.

src/video/video.c
Nopey: that does it, thanks!

katajakasa:
One way to do this with final rendering results would be to:
1. Add another rendering step and FBO between the palette and rgba shaders, which takes care of the remaining remapping operations
2. Schedule the screenshot to happen after that phase -- no need to do offscreen screenshotting
3. Reserve enough space from texture atlas, and do a direct memory copy inside the gpu. This gets rid of the slow glReadPixels that stalls the gpu, and we don't need to move the picture back and forth.

src/video/video.c
Nopey: moved the palette and remap update above the call to video_render_finish_offscreen

src/game/utils/har_screencap.c
Nopey: This PR does not introduce this problem, nor attempt to solve it.
Should we block this PR on the lack of remapping support in video_area_capture?

katajakasa (approved):

PR #758, opened 2024-11-06
Do some planning on plugin interface.

katajakasa:
Switchable renderer plugin interface.


src/video/renderers/opengl3/gl3_renderer.c
```diff
@@ -0,0 +1,381 @@
...
+    gl3_context *ctx = userdata;
+    ctx->draw_atlas = toggle;
+}
+
+static void renderer_init(renderer *gl3_renderer) {
+    gl3_renderer->ctx = omf_calloc(1, sizeof(gl3_context));
+}
+
+static void renderer_close(renderer *gl3_renderer) {
+    omf_free(gl3_renderer);
```
Nopey, line 354: I think you meant `omf_free(gl3_renderer->ctx)`?

src/video/video.c
```diff
@@ -1,306 +1,174 @@
...
-    if(!create_program(&g_video_state.rgba_prog_id, "rgba.vert", "rgba.frag")) {
-        goto error_3;
+/**
+ * This is run at start to hunt the available renderers.
+ */
+void video_scan_renderers(void) {
+    renderer tmp;
+    bool available;
+    for(int i = 0; i < renderer_count; i++) {
+        renderer_statuses[i].init(&tmp);
```
Nopey, line 39: renderer init without close leaks memory

src/video/video.c
```diff
@@ -1,306 +1,174 @@
...
+            return true;
+        }
+    }
+    return false;
 }
 
-void video_render_prepare(void) {
-    object_array_prepare(g_video_state.objects);
+static bool video_find_renderer(const char *try_name) {
+    memset(&current_renderer, 0, sizeof(renderer));
```
Nopey, line 87: if video_find_renderer is to be called at most once per program run, it does not need to zero current_renderer-- C guarantees static objects are initialized.
If video_find_renderer can be called multiple times per program run (e.g. switching renderers in the settings menu), it should close the previous current_renderer before memsetting it to zero to prevent a leak.

same applies to find_best_renderer.

src/video/renderers/null/null_renderer.c
```diff
@@ -0,0 +1,98 @@
...
+static bool setup_context(void *userdata, int window_w, int window_h, bool fullscreen, bool vsync) {
+    INFO("NULL Renderer initialized!");
+    return true;
+}
+
+static void get_context_state(void *userdata, int *window_w, int *window_h, bool *fullscreen, bool *vsync) {
+    if(window_w != NULL)
+        *window_w = 0;
+    if(window_h != NULL)
+        *window_h = 0;
```
mrannanj, line 28: duplicate lines here

Nopey: no?
window_w and window_h are different pointers

mrannanj: yeah, I saw wrong.

src/video/video.c
katajakasa: There is no corresponding close for gl3_renderer_init(), since it does not reserve memory

src/video/video.c
katajakasa: None of these functions do any memory allocations; they just swap function pointers in the current_renderer struct.

But yeah, the memsets were debugging stuff; I will get rid of them.

src/video/renderers/opengl3/gl3_renderer.c
katajakasa: Yes, I did. Will fix.

src/video/video.c
katajakasa: Reorganized this part to be clearer.

src/video/video.c
katajakasa: Reorganized this part to be clearer.

src/video/renderers/opengl3/gl3_renderer.c
Nopey: Perhaps init and close should be called create and destroy instead?

katajakasa: Sounds good to me.

Nopey:
Without CMake changes, all .c files in src/video/renderer/ will be compiled, regardless of whether they are for the target platform build.
Can I push a change that allows render plugins to be disabled at compile-time via cmake?

src/game/scenes/mainmenu/menu_video.c
```diff
@@ -80,6 +80,11 @@ void resolution_toggled(component *c, void *userdata, int pos) {
     }
 }
 
+void renderer_toggled(component *c, void *userdata, int pos) {
+    settings_video *v = &settings_get()->video;
+    video_get_renderer_info(pos, &v->renderer, NULL);
```
Nopey, line 85: settings' F_STRING is for owned strings, not string constants (this currently crashes on exit if the settings have been opened).
v->renderer should be a `char*`, and renderer_toggled needs to free and allocate (or do both at once with realloc).

src/game/scenes/mainmenu/menu_video.c
```diff
@@ -128,6 +133,18 @@ component *menu_video_create(scene *s) {
...
+    component *renderer_selector =
+        textselector_create(&tconf, "API:", "Choose the video renderer API to use", renderer_toggled, local);
+    menu_attach(menu, renderer_selector);
+
+    // Add standard resolutions
+    const char *r_name;
+    for(int r = 0; r < video_get_renderer_count(); r++) {
+        video_get_renderer_info(r, &r_name, NULL);
+        textselector_add_option(renderer_selector, r_name);
+    }
```
Nopey, line 146: comment has been copy-pasted and is not correct.
pedantic: Why is r_name outside of the for-loop?

src/game/scenes/mainmenu/menu_video.c
```diff
@@ -128,6 +133,18 @@ component *menu_video_create(scene *s) {
     menu_attach(menu, label_create(&tconf, "VIDEO"));
     menu_attach(menu, filler_create());
 
+    // Renderer selector
+    component *renderer_selector =
+        textselector_create(&tconf, "API:", "Choose the video renderer API to use", renderer_toggled, local);
```
Nopey, line 138: this textselector will always return to the first item in its list (when leaving the video options & going back), as there's no `textselector_set_pos` call.

src/game/scenes/mainmenu/menu_video.c
katajakasa: https://github.com/omf2097/openomf/pull/774/files

src/game/scenes/mainmenu/menu_video.c
katajakasa: https://github.com/omf2097/openomf/pull/775

PR #756, opened 2024-11-06
Prevent palette transformers from staying stuck on

Nopey:
Palette transformers now damage both the current frame's dmg_current and the next frame's dmg_base

This ensures that the first frame a palette transformer is removed, the base colors will be written to the current palette & uploaded.

PR #759, opened 2024-11-07
Treat object palette transformers like other cb's

Nopey:
Objects now have a `palette_transform` field to store their palette transformer callback.

`vga_render` now calls into game_state and object to gather all the palette transformers for the render.

Much like PR 756, `vga_state_render` will now calculate the palette data on the first frame all transformers are off, to ensure the current palette is updated to match the base palette. This solves the bug where palette transformer's effects persist after removal.

Obsoletes PR #756

EDIT: removed vga_state_tick layer of laziness/change detection-- transformers are now gathered every render frame

### DIFF
--- a/src/video/vga_state.c
+++ b/src/video/vga_state.c
@@ -1,5 +1,6 @@
 #include "video/vga_state.h"
 
+#include "game/game_state.h"
 #include "utils/miscmath.h"
 #include "utils/png_writer.h"
 #include <assert.h>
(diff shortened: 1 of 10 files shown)


src/video/vga_state.c
```diff
@@ -47,9 +48,14 @@ void vga_state_pop_palette(void) {
     damage_set_range(&state.dmg_base, 0, 255);
 }
 
-void vga_state_render(void) {
+void vga_state_render(game_state *gs) {
+    unsigned int old_count = state.transformer_count;
+    // recalculate transformers stack
+    state.transformer_count = 0;
+    game_state_palette_transform(gs);
```
katajakasa, line 55: vga_state should not reach into other subsystems. This function should be called from engine.c instead.

Nopey:
fixed the scene crossfade timer being ticked in static_tick and checked in dynamic_state-- this allowed frames to render between the next_wait_ticks becoming 0 and the scene transitioning to the next, which was quite jarring.

katajakasa (approved):

PR #760, opened 2024-11-07
Add incremental BK loading

Vagabond:
On slow hardware, the blocking bk loading can cause the audio to stutter because it blocks the engine loop. This PR makes loading BK files incremental, so the game loop can keep ticking and loading the BK file incrementally during scene transitions.

### DIFF
--- a/src/formats/bk.c
+++ b/src/formats/bk.c
@@ -93,13 +93,14 @@ void sd_bk_postprocess(sd_bk_file *bk) {
 }
 
 int sd_bk_load_incremental(sd_bk_file *bk, sd_reader *r) {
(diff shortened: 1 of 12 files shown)


Vagabond:
One issue is, on scene transition without a "crossfade" it flickers a bit

Nopey:
On a transition with cross-fade, it plays the cross fade four times

https://github.com/user-attachments/assets/7811ae53-3d79-4806-a8ff-a5a1c6bf2ab2

Nopey (changes requested):
OMF logo to main menu crossfade repeats four times
Other cross fades seem unaffected

PR #761, opened 2024-11-07
Show Newsroom and VS in DEMO mode and Score fixes

Nopey:
Newsroom and VS are now part of the demo sequence.
This makes it possible to see the character-vs-themselves insults in openomf =)
image

Scores are now handled more like the original, hiding the score points for nonhuman players (here, two ai)
image
VITALITY and PERFECT ROUND score bonuses are now awarded every round (instead of just the final round), and only when a human player defeats an AI.

not changed:
- DEMO still does not visit the MELEE scene as Player 1, unlike the original game.
- Pilots will still use the Nova in demo mode (this is done in game_state_init_demo)
- Kreissack still does not appear in demo mode (but easily could, if we wanted him to. same function.)

### DIFF
--- a/src/game/common_defines.h
+++ b/src/game/common_defines.h
@@ -91,6 +91,8 @@ enum
     PILOT_RAVEN,
     PILOT_KREISSACK,
     NUMBER_OF_PILOT_TYPES,
+    // Major Kreissack is not a playable character
+    NUMBER_OF_PLAYABLE_PILOT_TYPES = NUMBER_OF_PILOT_TYPES - 1,
 };
 
 enum
--- a/src/game/utils/score.c
+++ b/src/game/utils/score.c
@@ -126,25 +126,26 @@ void chr_score_tick(chr_score *score) {
     }
 }
 
-void chr_score_render(chr_score *score) {
-    // Render all texts in list to right spot
-    char tmp[50];
-    score_format(score->score, tmp, 50);
-
+void chr_score_render(chr_score *score, bool render_total_points) {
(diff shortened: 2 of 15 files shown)


src/game/common_defines.h
```diff
@@ -89,6 +89,8 @@ enum
     PILOT_RAVEN,
     PILOT_KREISSACK,
     NUMBER_OF_PILOT_TYPES,
+    // Major Kreissack is not a playable character
+    NUMBER_OF_PLAYABLE_PILOT_TYPES = NUMBER_OF_PILOT_TYPES - 1
```
mrannanj, line 93: Best to end all enums with a comma, in case there will be new entries later, there will be less noise in the diff.

src/game/utils/score.c
```diff
@@ -126,25 +126,27 @@ void chr_score_tick(chr_score *score) {
...
+    if(render_total_points) {
+        char tmp[50];
+        score_format(score->score, tmp, 50);
 
-    } else {
-        int s2len = strlen(tmp) * font_small.w;
-        text_render(&tconf_score, TEXT_DEFAULT, score->x - s2len, score->y, 200, 6, tmp);
+        if(score->direction == OBJECT_FACE_RIGHT) {
+            text_render(&tconf_score, TEXT_DEFAULT, score->x, score->y, 200, 6, tmp);
+
```
mrannanj, line 142: Looks like here is an extra newline.

mrannanj (approved):

PR #762, opened 2024-11-07
Remove extraneous

katajakasa:
So I quickly tried the new code, and it broke Credits. Removing these seems to fix things. After looking at these, I don't quite understand what they are meant to do that the original code doesn't already do ?

### DIFF
--- a/src/engine.c
+++ b/src/engine.c
@@ -289,7 +289,6 @@ void engine_run(engine_init_flags *init_flags) {
 
         // Do the actual video rendering jobs
         if(enable_screen_updates) {
-            vga_state_clear_palette_transform();
             game_state_palette_transform(gs);
 
             vga_state_render();
--- a/src/video/damage_tracker.c
+++ b/src/video/damage_tracker.c
@@ -20,6 +20,7 @@ void damage_combine(damage_tracker *dst, const damage_tracker *src) {
     dst->dirty_range_start = min2(dst->dirty_range_start, src->dirty_range_start);
     dst->dirty_range_end = max2(dst->dirty_range_end, src->dirty_range_end);
 }
+
 void damage_set_range(damage_tracker *tracker, vga_index start, vga_index end) {
     tracker->dirty = true;
     tracker->dirty_range_start = min2(tracker->dirty_range_start, start);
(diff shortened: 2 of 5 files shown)


Nopey:
The render target is leaking image data from the previous frame into the current frame.
In branch [magnus/debug-credits](https://github.com/omf2097/openomf/tree/magnus/debug-credits) I wrote some debugging that prints how many objects and transformers are in each frame while stepping through frames with your debugger_proceed F5+Space.

For the credits scene, it outputs this when a credit finishes fading out:
```
[    400][D] engine_run(): render tick 400 {
[    400][D] engine_run(): render tick 400 } (3 objects rendered)
[    402][D] engine_run(): render tick 402 {
[    402][D] vga_state_render(): transformer_count went from 1 to 0
[    402][D] engine_run(): render tick 402 } (2 objects rendered)     
```
On tick 402, the text flashed red for a frame or two before vanishing-- it did not stick around for me to press space again.
This is because the text that was fading out is not in the list of objects to be rendered in frame 402, at all (it got deleted)

I tried adding a glClear above video_render_finish's offscreen render calls:
```c

    // Clearing the render target texture should have no effect, because we always render a scene background
    // However, uncommenting the following line causes every second frame to be blindingly green
    // glClear(GL_COLOR_BUFFER_BIT);

    object_array_blend_mode mode;
    while(object_array_get_batch(g_video_state.objects, &batch, &mode)) {
        video_set_blend_mode(mode);
        object_array_draw(g_video_state.objects, &batch);
    }
```
This makes the game flash green, please do not do this if you are sensitive.
However, I argue that it should not matter if we clear the render target before rendering to it as we always render the scene's background!

Nopey:
I expected inserting a glFinish() between the render to FBO and the RGBA pass to help, but the game's still flashing. What am I missing?
Would double buffering the render target FBO help?

It's late here now, so I'm gonna stop debugging this for now
I'll let you know if I pick this up again, but for now I'm out of leads

Nopey:
As for a quick example of things not working properly with `dmg_transform` tracking removed:
image
image
The top image is a screenshot of the credits background with dmg_transform tracking, and the second without.
The palette in the second is subtly dimmed from the crossfade palette transformer (that already finished, many frames ago!).

There's probably better examples of this damage tracking preventing transformers from staying stuck on, but I noticed this one while debugging the credits :")

Nopey (approved):

PR #765, opened 2024-11-08
FPS counter

katajakasa:
DO NOT MERGE! This uses GetPerformanceCounter, which is not very portable. We should wait for SDL3 and SDL_GetTicksNS() for this.

So, for dev testing only for now!

### DIFF
--- a/src/engine.c
+++ b/src/engine.c
@@ -1,12 +1,11 @@
 #include "engine.h"
 #include "audio/audio.h"
 #include "console/console.h"
-#include "controller/controller.h"
 #include "formats/altpal.h"
 #include "game/game_player.h"
 #include "game/game_state.h"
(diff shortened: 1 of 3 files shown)


Vagabond:
This seems silly to need SDL3 for, we could just use a platform dependent high resolution timing source, or use a library like https://github.com/nclack/tictoc

Nopey:
SDL3's SDL_GetTicksNS uses SDL_GetPerformanceCounter, and some values calculated from SDL_GetPerformanceFrequency.

https://github.com/libsdl-org/SDL/blob/5c8bed378040007d0e2e9267ff4bc76c016cee8f/src/timer/SDL_timer.c#L624

What about GetPerformanceCounter is not portable?
We could avoid the (double) arithmetic if we're careful with our Uint64's

PR #766, opened 2024-11-09
Fix Credits, make only object_palette_copy_transform sticky

Nopey:
Same goal as https://github.com/omf2097/openomf/pull/762, while only making `object_palette_copy_transform` sticky.

the `Shuffle palette repair a bit` commit didn't merge cleanly, so I didn't cherry-pick it.


Nopey:
Closing in favor of #762

PR #708, opened 2024-10-31
Attempt to add NAT libraries to vcpkg

Vagabond:

PR #770, opened 2024-11-18
CMake: Use targets for dependencies instead of variables

Nopey:
- Added a new script, cmake-scripts/Dependencies.cmake
- Dependencies.cmake creates cmake targets for each depenency, eg. openomf::argtable, that contain all the necessary link libraries, include paths, and compiler defines (including openomf-specific ones like ARGTABLE3_FOUND)
- Depenendencies.cmake knows to use find_package CONFIG mode in vcpkg builds, and find_package module mode in other builds.
- vcpkg shims have been removed from FindXYZ.cmake scripts, as they are no longer necessary.
- optional dependencies (CUnit, miniupnpc, natpmp) are now only built by vcpkg when a corresponding option is set (USE_TESTS, USE_MINIUPNPC, and USE_NATPMP).
- miniupnpc and natpmp are now only linked to when a corresponding option is set: USE_MINIUPNPC and USE_NATPMP
- Requesting an optional dep (eg USE_MINIUPNPC) will cause the build to fail if the dep cannot be found; this prevents miniupnpc and natpmp from silently regressing in CI.

Closes #708, as this PR contains it.


Nopey:
Rebased on latest master and squashed into two commits-- one for Vagabond's libnatpmp port (from PR 708) and one for the cmake changes.

cmake-scripts/Dependencies.cmake
```diff
@@ -0,0 +1,134 @@
...
+# this file locates all third party dependencies, and creates easy to use
+# targets like `openomf::argtable` for the openomf build scripts to link to.
+
+# set module path only temporarily, to discourage find_package being used outside of this script
+set(ORIGINAL_CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH})
+set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake-scripts)
+
+
+# SDL2, SDL2_mixer
+if(VCPKG_TOOLCHAIN)
```
katajakasa, line 11: Could these if-elses be minimized to just contain stuff that is different ? And move the identical lines outside of the if-else block ?

katajakasa (changes requested):

cmake-scripts/Dependencies.cmake
Nopey: Moved the add_library calls outside of the vcpkg_toolchain if/else's, thanks-- that helps make it clear what targets Dependencies.cmake is providing to the rest of the scripts.
Note that I kept the find_package calls intentionally within the if-else blocks.

Also fixed the mess of tabs and spaces that was the indentation in this file (oops)

katajakasa (approved):
LGTM as far as I know about vcpkg. Although I am kind of annoyed how much additional stuff it seems to be requiring.

Nopey:
rebased after cmakelists.txt conflict, gonna wait for CI to pass before a merge

Nopey:
borked it heh

Nopey:
removed platform.h.in and check_symbol_exists(strdup) stuff from CMakeLists.txt, since previous pr already pulled out it out.

katajakasa (approved):

Nopey:
CI passed and pre-rebase version had approval-- merging.

PR #771, opened 2024-11-18
Add big endian support

nopjne:
The big endian support assumes the GCC compiler and the presence of __builtin_bswapX.


src/formats/internal/memreader.c
```diff
@@ -5,6 +5,11 @@
 #include "formats/internal/memreader.h"
 #include "formats/internal/reader.h"
 #include "utils/allocator.h"
+#ifdef N64_BUILD
+#include <libdragon.h>
+#elif BIG_ENDIAN_BUILD
+#define BITCAST_I2F(x) (float)(x)
+#endif
```
Nopey, line 12: This is not a correct fallback implementation of [libdragon's BITCAST_I2F](https://libdragon.dev/ref/fmath_8h_source.html).
BITCAST_I2F is more similar to (C++) `bit_cast<float>(x)`.

Perhaps it would be easier to use memcpy directly and avoid the utility function?
```c

#ifdef BIG_ENDIAN_BUILD
    uint32_t fl;
    fl = *((uint32_t *)(reader->buf + reader->pos));
    fl = __builtin_bswap32(fl);
    memcpy(&r, &fl, sizeof(fl));
    reader->pos += sizeof(fl);
#else
```

src/formats/internal/memreader.c
```diff
@@ -83,22 +96,38 @@ int8_t memread_byte(memreader *reader) {
...
     reader->pos += sizeof(r);
     return r;
 }
 
 float memread_float(memreader *reader) {
     float r;
+#ifdef BIG_ENDIAN_BUILD
+    uint32_t fl;
+    fl = *((uint32_t *)(reader->buf + reader->pos));
```
Nopey, line 123: Please use a memcpy, as the buf+pos pointer is insufficiently aligned to be a `(uint32_t *)` .
```diff
-    fl = *((uint32_t *)(reader->buf + reader->pos));
+    memcpy(&fl, reader->buf + reader->pos, sizeof(fl));
```

src/formats/internal/writer.c
```diff
@@ -7,6 +7,9 @@
 
 #include "formats/internal/writer.h"
 #include "utils/allocator.h"
+#ifdef N64_BUILD
+#include <libdragon.h>
+#endif
```
Nopey, line 12: This include is unused.
Please remove all `#include <libdragon.h>`'s from this PR, none should be needed.

src/formats/internal/memwriter.c
```diff
@@ -57,27 +60,58 @@ void memwrite_ubyte(memwriter *writer, uint8_t value) {
...
+    int16_t t = value;
+    t = __builtin_bswap16(t);
+    memwrite_buf(writer, (char *)&t, 2);
+#else
     memwrite_buf(writer, (char *)&value, 2);
+#endif
 }
 
 void memwrite_dword(memwriter *writer, int32_t value) {
+#ifdef N64_BUILD
```
Nopey, line 97: Why are these checking N64_BUILD and not BIG_ENDIAN_BUILD?

Nopey:
__builtin_bswapX usage seems justified, and if we ever need to port to a big-endian compiler that doesn't support them it's easy enough to do the ugly bitwise integer operations or add support for that hypothetical compiler's equivalent builtin.

reviewers: please ignore the MSVC build failure, all MSVC CI pipelines are expected to fail until PR 769 or an alternative is merged.

Vagabond:
Why do you sometimes introduce a temporary variable to swap, but other times you swap the value in-place?

Vagabond:
> > Why do you sometimes introduce a temporary variable to swap, but other times you swap the value in-place?
> 
> I think this is an N64_BUILD quirk specific to floats, but I didn't want to introduce the N64_BUILD in here. The problem is that doing a bswap on a float causes a floating point exception, the exception could be silenced but would need a libdragon change and would also silence any other float exceptions.

This is a valid reason, maybe a comment to that effect would be a good idea to prevent overzealous optimization in future?

src/formats/internal/reader.c
```diff
@@ -170,18 +195,27 @@ int8_t sd_peek_byte(sd_reader *reader) {
...
+#endif
     return d;
 }
 
 float sd_peek_float(sd_reader *reader) {
     float f = 0;
     sd_peek_buf(reader, (char *)&f, 4);
+#ifdef BIG_ENDIAN_BUILD
+    f = __builtin_bswap16(f);
+#endif
```
Nopey, line 218: [__builtin_bswap16](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) does not take an argument of type float, it takes a uint16_t.
The `__builtin_bswap16(f)` call is equivalent to `__builtin_bswap16((uint16_t)f)`, which is why the N64 build was encountering floating point errors.

Also: bswap16 is the wrong number of bits, should be bswap32.

(same as other review comment: please implement sd_peek_float in terms of sd_peek_dword)

src/formats/internal/reader.c
```diff
@@ -213,6 +247,16 @@ int sd_read_line(const sd_reader *reader, char *buffer, int maxlen) {
     if(fgets(buffer, maxlen, reader->handle) == NULL) {
         return 1;
     }
+#ifdef BIG_ENDIAN_BUILD
+    int len = maxlen;
+    for(int i = 0; i < len / 4; i += 1) {
+        ((int *)buffer)[i] = __builtin_bswap32(((int *)buffer)[i]);
```
Nopey, line 253: `char *buffer` is insufficiently aligned to be a `int*`
same comment applies in sd_read_str.

please memcpy to a uint32_t local variable (and uint16_t local var for the final-2-bytes bytes-swap)

src/formats/internal/memreader.c
```diff
@@ -83,22 +91,37 @@ int8_t memread_byte(memreader *reader) {
...
+    uint32_t fl;
+    memcpy(&fl, reader->buf + reader->pos, sizeof(fl));
+    fl = __builtin_bswap32(fl);
+    reader->pos += sizeof(fl);
+#else
     memcpy(&r, reader->buf + reader->pos, sizeof(r));
     reader->pos += sizeof(r);
+#endif
     return r;
 }
```
Nopey, line 126: I think memread_float (and similar: memwrite_float, sd_read_float, etc..)
should be implemented in terms of their respective integer read-write functions.

for memread_float, this would look like:
```c
   uint32_t u = memread_dword(reader);
   float f;
   memcpy(&f, &u, 4);
   return f;
```

this way, they don't need to perform any byte swapping and can be focused on just the uint-to-float bit cast.

src/formats/internal/memreader.c
```diff
@@ -62,14 +62,22 @@ uint8_t memread_ubyte(memreader *reader) {
 
 uint16_t memread_uword(memreader *reader) {
     uint16_t r;
+#ifdef BIG_ENDIAN_BUILD
+    r = __builtin_bswap16(*((uint16_t *)(reader->buf + reader->pos)));
+#else
     memcpy(&r, reader->buf + reader->pos, sizeof(r));
+#endif
     reader->pos += sizeof(r);
     return r;
```
Nopey, line 71: buf+pos is not insufficiently aligned to be a (uint16_t *)

please use the memcpy from the existing code, and keep your #ifdef block as small as possible (your sd_peek_dword is a good example of how it should look).

Nopey (changes requested):

src/formats/internal/reader.c
nopjne: Well, the exceptions were definitely from doing float f = __buildin_bswap32((float)x);
The only reason this function is wrong is because nothing in the engine calls it, and I just didn't hit the issue.

src/formats/internal/memreader.c
nopjne: I mean this code is tested and working. dropping the bswap would make it not work.

src/formats/internal/memreader.c
nopjne: I've thought about this, and it hasn't been an issue with the current game assets. I don't want to slow it down more than needed.

src/formats/internal/memreader.c
Nopey: memread_dword performs a byteswap.

memread_float is currently only used to read some tournament mode pilot variables we ignore, unsure what you mean by tested and working

Nopey:
The byteswapping in sd_read_str is pure magic to me, as I'm not familiar with big endian systems.
I assume you've tested with and without the final 2-bytes bswap16 and found it necessary

PR #772, opened 2024-11-18
 Document allocator.h API, fail on omf_malloc(0)

Nopey:
allocator.h now has doc strings for the various methods it provides.

Notably, omf_malloc(0) and similar calls now have well-specified behavior (rather than libc implementation-specified behavior): it aborts the program with an error message

### DIFF
--- a/src/formats/sprite.c
+++ b/src/formats/sprite.c
@@ -59,8 +59,8 @@ int sd_sprite_load(sd_reader *r, sd_sprite *sprite) {
     sprite->missing = sd_read_ubyte(r);
 
     // Copy sprite data, if there is any.
-    if(sprite->missing == 0) {
-        sprite->data = omf_calloc(sprite->len, 1);
+    if(sprite->missing == 0 && sprite->len != 0) {
+        sprite->data = omf_calloc(1, sprite->len);
         sd_read_buf(r, sprite->data, sprite->len);
     } else {
         sprite->data = NULL;
--- a/src/resources/sprite.c
+++ b/src/resources/sprite.c
@@ -13,10 +13,16 @@ void sprite_create(sprite *sp, void *src, int id) {
     sd_sprite *sdsprite = (sd_sprite *)src;
     sp->id = id;
     sp->pos = vec2i_create(sdsprite->pos_x, sdsprite->pos_y);
-    sp->data = omf_calloc(1, sizeof(surface));
-    sp->owned = true;
+
+    if(sdsprite->width == 0 || sdsprite->height == 0) {
+        sp->data = NULL;
+        sp->owned = false;
+        return;
+    }
 
     // Load data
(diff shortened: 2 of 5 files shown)


src/formats/sprite.c
```diff
@@ -59,7 +59,7 @@ int sd_sprite_load(sd_reader *r, sd_sprite *sprite) {
     sprite->missing = sd_read_ubyte(r);
 
     // Copy sprite data, if there is any.
-    if(sprite->missing == 0) {
+    if(sprite->missing == 0 && sprite->len != 0) {
         sprite->data = omf_calloc(sprite->len, 1);
```
katajakasa, line 63: While you're at it, could tou turn this omf_calloc around ? omf_calloc(1, sprite->len) seems more correct.

src/resources/sprite.c
```diff
@@ -13,10 +13,14 @@ void sprite_create(sprite *sp, void *src, int id) {
     sd_sprite *sdsprite = (sd_sprite *)src;
     sp->id = id;
     sp->pos = vec2i_create(sdsprite->pos_x, sdsprite->pos_y);
-    sp->data = omf_calloc(1, sizeof(surface));
-    sp->owned = true;
+
+    if(sdsprite->width == 0 || sdsprite->height == 0) {
+        return;
```
katajakasa, line 18: Should this also set data, owned to null, false ?

src/utils/allocator.c
```diff
@@ -3,3 +3,4 @@
 const char *_text_malloc_error = "malloc(%zu) failed on %s:%d\n";
 const char *_text_calloc_error = "calloc(%zu, %zu) failed on %s:%d\n";
 const char *_text_realloc_error = "realloc(%p, %zu) failed on %s:%d\n";
+const char *_text_alloc_zero_error = "attempted to allocate 0 bytes on %s:%d\n";
```
katajakasa, line 6: No longer used

katajakasa (changes requested):

src/utils/allocator.h
```diff
@@ -1,15 +1,67 @@
...
+#define omf_realloc(ptr, size) omf_realloc_real((ptr), (size), __FILE__, __LINE__);
+
+/**
+ * @brief Free an allocation by pointer, and reassigns the pointer to NULL.
+ * @details Free an allocation from one of: omf_malloc, omf_calloc, or omf_realloc.
+ * If the pointer is already NULL, does nothing (well-behaved).
+ *
+ * Do not pass pointers from other allocators (strdup, SDL_malloc, etc..), doing so triggers undefined behavior!
+ * @param ptr the pointer to free & NULL.
+ */
```
katajakasa, line 60: Documentation! Very nice!

katajakasa (approved):

PR #773, opened 2024-11-19
PROTO: Create text_objects and use them for rendering to improve text rendering performance

Vagabond:
This is a cherry-pick/rework of the commit from the n64 branch. It builds and works but it needs more work/cleanups.

EDIT(Nopey): Closes #677

### DIFF
--- a/src/engine.c
+++ b/src/engine.c
@@ -154,6 +154,7 @@ void engine_run(engine_init_flags *init_flags) {
             // Handle other events
             switch(e.type) {
                 case SDL_QUIT:
+                    assert(false);
                     run = 0;
                     break;
                 case SDL_KEYDOWN:
--- a/src/engine.c
+++ b/src/engine.c
@@ -154,7 +154,6 @@ void engine_run(engine_init_flags *init_flags) {
             // Handle other events
             switch(e.type) {
                 case SDL_QUIT:
-                    assert(false);
                     run = 0;
                     break;
                 case SDL_KEYDOWN:
--- a/src/game/gui/label.c
+++ b/src/game/gui/label.c
@@ -6,15 +6,17 @@
 typedef struct {
     char *text;
     text_settings tconf;
+    text_object text_cache[1];
 } label;
 
 static void label_render(component *c) {
     label *local = widget_get_obj(c);
-    text_render(&local->tconf, TEXT_DEFAULT, c->x, c->y, c->w, c->h, local->text);
+    text_render(&(local->text_cache[0]), &local->tconf, TEXT_DEFAULT, c->x, c->y, c->w, c->h, local->text);
 }
 
 static void label_free(component *c) {
     label *local = widget_get_obj(c);
+    text_objects_free(local->text_cache, 1);
     omf_free(local->text);
     omf_free(local);
 }
@@ -45,7 +47,7 @@ component *label_create_with_width(const text_settings *tconf, const char *text,
 
     int tsize = text_char_width(tconf);
     int longest = 0;
-    int h = text_find_line_count(tconf, max_width / tsize, 0, strlen(text), text, &longest);
+    int h = text_find_line_count(tconf, max_width / tsize, 0, text, &longest);
 
     // fonts are all 8 high?
     component_set_size_hints(c, longest * tsize, h * 8);
@@ -55,7 +57,7 @@ component *label_create_with_width(const text_settings *tconf, const char *text,
     widget_set_obj(c, local);
     widget_set_render_cb(c, label_render);
     widget_set_free_cb(c, label_free);
-
+    //local->text_cache[0].dynamic = true;
     return c;
 }
 
--- a/src/game/gui/label.c
+++ b/src/game/gui/label.c
@@ -57,7 +57,7 @@ component *label_create_with_width(const text_settings *tconf, const char *text,
     widget_set_obj(c, local);
     widget_set_render_cb(c, label_render);
     widget_set_free_cb(c, label_free);
-    //local->text_cache[0].dynamic = true;
+    // local->text_cache[0].dynamic = true;
     return c;
 }
 
--- a/src/game/gui/menu.c
+++ b/src/game/gui/menu.c
@@ -111,9 +111,10 @@ static void menu_render(component *c) {
                 video_draw(m->help_bg2, m->help_x - 8, m->help_y - 8);
             }
             // TODO: The text_cache should move into inner component. In this case the one of the s->objs.
-            // Right now it will get invalidated on every render. For help text it usually doesn't matter because there is
-            // usually only 1 help text on the screen per time.
-            text_render(&(m->text_cache[0]), &m->help_text_conf, TEXT_DEFAULT, m->help_x, m->help_y, m->help_w, m->help_h, (*tmp)->help);
+            // Right now it will get invalidated on every render. For help text it usually doesn't matter because there
+            // is usually only 1 help text on the screen per time.
+            text_render(&(m->text_cache[0]), &m->help_text_conf, TEXT_DEFAULT, m->help_x, m->help_y, m->help_w,
+                        m->help_h, (*tmp)->help);
         }
         i++;
     }
--- a/src/game/gui/menu.c
+++ b/src/game/gui/menu.c
@@ -110,7 +110,10 @@ static void menu_render(component *c) {
             if(m->help_bg2) {
                 video_draw(m->help_bg2, m->help_x - 8, m->help_y - 8);
             }
-            text_render(&m->help_text_conf, TEXT_DEFAULT, m->help_x, m->help_y, m->help_w, m->help_h, (*tmp)->help);
+            // TODO: The text_cache should move into inner component. In this case the one of the s->objs.
+            // Right now it will get invalidated on every render. For help text it usually doesn't matter because there is
+            // usually only 1 help text on the screen per time.
+            text_render(&(m->text_cache[0]), &m->help_text_conf, TEXT_DEFAULT, m->help_x, m->help_y, m->help_w, m->help_h, (*tmp)->help);
         }
         i++;
     }
@@ -368,6 +371,7 @@ void menu_set_help_text_settings(component *c, text_settings *settings) {
 
 static void menu_free(component *c) {
     menu *m = sizer_get_obj(c);
+    text_objects_free(m->text_cache, 1);
     if(m->bg1) {
         surface_free(m->bg1);
         omf_free(m->bg1);
@@ -409,6 +413,11 @@ void menu_set_margin_top(component *c, int margin) {
 void menu_set_padding(component *c, int padding) {
     menu *m = sizer_get_obj(c);
     m->padding = padding;
+}
+
+void menu_invalidate_help_text_cache(component *c) {
+    menu *m = sizer_get_obj(c);
+    m->text_cache->dirty = true;
 }
 
 component *menu_create(int obj_h) {
--- a/src/game/gui/text_render.c
+++ b/src/game/gui/text_render.c
@@ -6,6 +6,7 @@
 #include "utils/miscmath.h"
 #include "utils/vector.h"
 #include "video/video.h"
+#include "utils/allocator.h"
 
 static unsigned char FIRST_PRINTABLE_CHAR = (unsigned char)' ';
 
@@ -54,7 +55,48 @@ static inline surface *get_font_surface(const text_settings *settings, char ch) 
     return (ret == NULL) ? NULL : *ret;
 }
 
-static inline void render_char_shadow_surface(const text_settings *settings, const surface *sur, int x, int y) {
+
+int text_render_char_uncached(const text_settings *settings, text_mode state, int x, int y, char ch, bool shadow) {
+    surface *sur = get_font_surface(settings, ch);
+
+    if(sur == NULL) {
+        return 0;
+    }
+    // Handle shadows if necessary
+    if (shadow != false) {
+        if(settings->shadow & TEXT_SHADOW_RIGHT)
+            video_draw_offset(sur, x + 1, y, settings->cshadow, 255);
+        if(settings->shadow & TEXT_SHADOW_LEFT)
+            video_draw_offset(sur, x - 1, y, settings->cshadow, 255);
+        if(settings->shadow & TEXT_SHADOW_BOTTOM)
+            video_draw_offset(sur, x, y + 1, settings->cshadow, 255);
+        if(settings->shadow & TEXT_SHADOW_TOP)
+            video_draw_offset(sur, x, y - 1, settings->cshadow, 255);
+
+        return sur->w;
+    }
+
+    int color;
+    switch(state) {
+        case TEXT_SELECTED:
+            color = settings->cselected;
+            break;
+        case TEXT_UNSELECTED:
+            color = settings->cinactive;
+            break;
+        case TEXT_DISABLED:
+            color = settings->cdisabled;
+            break;
+        default:
+            color = settings->cforeground;
+    }
+
+    // Draw the actual font
+    video_draw_offset(sur, x, y, color - 1, 255);
+    return sur->w;
+}
+
+/*static inline void render_char_shadow_surface(const text_settings *settings, const surface *sur, int x, int y) {
     if(settings->shadow & TEXT_SHADOW_RIGHT)
         video_draw_offset(sur, x + 1, y, settings->cshadow, 255);
     if(settings->shadow & TEXT_SHADOW_LEFT)
@@ -63,9 +105,9 @@ static inline void render_char_shadow_surface(const text_settings *settings, con
         video_draw_offset(sur, x, y + 1, settings->cshadow, 255);
     if(settings->shadow & TEXT_SHADOW_TOP)
         video_draw_offset(sur, x, y - 1, settings->cshadow, 255);
-}
+}*/
 
-static inline void render_char_surface(const text_settings *settings, text_mode state, surface *sur, int x, int y) {
+/*static inline void render_char_surface(const text_settings *settings, text_mode state, surface *sur, int x, int y) {
     int color;
     switch(state) {
         case TEXT_SELECTED:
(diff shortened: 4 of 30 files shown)


src/game/gui/menu.c
```diff
@@ -411,6 +416,11 @@ void menu_set_padding(component *c, int padding) {
     m->padding = padding;
 }
 
+void menu_invalidate_help_text_cache(component *c) {
```
Vagabond, line 419: Does anything use this?

src/game/gui/textinput.c
```diff
@@ -212,6 +213,7 @@ static void textinput_free(component *c) {
     textinput *tb = widget_get_obj(c);
     surface_free(&tb->sur);
     str_free(&tb->text);
+    text_objects_free(tb->text_cache, (sizeof(tb->text_cache) / sizeof(tb->text_cache[0])));
```
Vagabond, line 216: Why isn't the second parameter an integer?

src/game/gui/textinput.c
```diff
@@ -52,16 +53,16 @@ static void textinput_render(component *c) {
             int xspace = c->w - tb->tconf.padding.left - tb->tconf.padding.right;
             start_x += ceilf((xspace - tmp_s) / 2.0f);
             tb->tconf.halign = TEXT_LEFT;
-            text_render(&tb->tconf, TEXT_DEFAULT, start_x + offset, c->y, c->w, c->h, "\x7F");
+            text_render(&tb->text_cache[0], &tb->tconf, TEXT_DEFAULT, start_x + offset, c->y, c->w, c->h, "\x7F");
             tb->tconf.halign = TEXT_CENTER;
         } else {
-            text_render(&tb->tconf, TEXT_DEFAULT, start_x + offset, c->y, c->w, c->h, "\x7F");
+            text_render(&tb->text_cache[1], &tb->tconf, TEXT_DEFAULT, start_x + offset, c->y, c->w, c->h, "\x7F");
```
Vagabond, line 59: This can also use `text_cache[0]`, as it is an else.

src/game/gui/textinput.c
```diff
@@ -27,6 +27,7 @@ typedef struct {
 
     textinput_done_cb done_cb;
     void *userdata;
+    text_object text_cache[3];
```
Vagabond, line 30: Can be 2 not 3 here

src/game/gui/textinput.c
```diff
@@ -262,5 +264,8 @@ component *textinput_create(const text_settings *tconf, int max_chars, const cha
     widget_set_action_cb(c, textinput_action);
     widget_set_tick_cb(c, textinput_tick);
     widget_set_free_cb(c, textinput_free);
+    tb->text_cache[0].dynamic = true;
```
Vagabond, line 267: We can invalidate on keystroke, we don't need it to be fully dynamic

src/game/scenes/arena.c
```diff
@@ -1225,6 +1231,10 @@ void arena_startup(scene *scene, int id, int *m_load, int *m_repeat) {
 int arena_create(scene *scene) {
     settings *setting;
     arena_local *local;
+    memset(text_cache, 0, sizeof(text_cache));
+    for(size_t i = 0; i < (sizeof(text_cache) / sizeof(text_cache[0])); i++) {
+        text_cache[i].dynamic = true;
```
Vagabond, line 1236: Check this is needed, it may not be.

src/game/scenes/cutscene.c
```diff
@@ -77,11 +78,13 @@ static void cutscene_input_tick(scene *scene) {
...
 static void cutscene_render_overlay(scene *scene) {
     cutscene_local *local = scene_get_userdata(scene);
-    text_render(&local->text_conf, TEXT_DEFAULT, local->text_x, local->text_y, local->text_width, 200, local->current);
+    text_render(&local->text_cache[0], &local->text_conf, TEXT_DEFAULT, local->text_x, local->text_y, local->text_width,
+                200, local->current);
 }
 
 static void cutscene_free(scene *scene) {
     cutscene_local *local = scene_get_userdata(scene);
+    text_objects_free(local->text_cache, (sizeof(local->text_cache) / sizeof(local->text_cache[0])));
```
Vagabond, line 87: just use 1 as second parameter here?

src/game/scenes/mainmenu/menu_main.c
```diff
@@ -113,6 +113,10 @@ component *menu_main_create(scene *s) {
     menu_attach(menu, textbutton_create(&tconf, "HELP", NULL, COM_DISABLED, NULL, NULL));
     menu_attach(menu, textbutton_create(&tconf, "DEMO", NULL, COM_ENABLED, mainmenu_demo, s));
     menu_attach(menu, textbutton_create(&tconf, "SCOREBOARD", NULL, COM_ENABLED, mainmenu_soreboard, s));
+#ifdef N64_BUILD
```
Vagabond, line 116: This probably doesn't belong in this PR, remove.

src/game/scenes/melee.c
```diff
@@ -52,6 +52,10 @@ typedef struct {
 #define CURSORS_DONE(local) (local->cursor[0].done && local->cursor[1].done)
 #define CURSOR_NOVA_SELECT(local, player) (local->cursor[player].row == 1 && local->cursor[player].column == 2)
 
+static const char *p1_bio_prev = NULL;
```
Vagabond, line 55: Unclear why these can't be part of the `local` struct?

src/game/scenes/newsroom.c
```diff
@@ -182,7 +184,8 @@ void newsroom_overlay_render(scene *scene) {
         tconf_yellow.strip_leading_whitespace = false;
         tconf_yellow.strip_trailing_whitespace = true;
         tconf_yellow.max_lines = 9;
-        text_render(&tconf_yellow, TEXT_DEFAULT, 34, 155, 250, 6, str_c(&local->news_str));
+        local->text_cache->dirty = true;
```
Vagabond, line 187: We can probably dirty this on input, not every frame.

src/game/scenes/scoreboard.c
```diff
@@ -26,6 +26,7 @@ typedef struct scoreboard_local_t {
     int page;
     component *ti;
     guiframe *frame;
+    text_object text_cache[22];
```
Vagabond, line 29: Missing frees?

src/game/scenes/vs.c
```diff
@@ -312,73 +319,83 @@ void vs_render(scene *scene) {
         char text[256];
         char money[16];
         fight_stats *fight_stats = &scene->gs->fight_stats;
+        static int force_redraw = 1;
+        if(force_redraw) {
+            // TODO: Find out why this caching doesn't work. The cache is built before the stats are correct.
```
Vagabond, line 324: TODO: investigate this

src/game/scenes/vs.c
```diff
@@ -129,6 +130,10 @@ void cb_vs_destroy_object(object *parent, int id, void *userdata) {
 
 void vs_free(scene *scene) {
     vs_local *local = scene_get_userdata(scene);
+    size_t cache_size = sizeof(local->text_cache) / sizeof(local->text_cache[0]);
```
Vagabond, line 133: Shouldn't we be freeing here?

src/game/utils/score.c
```diff
@@ -38,6 +39,8 @@ vec2i interpolate(vec2i start, vec2i end, float fraction) {
 }
 
 void chr_score_create(chr_score *score) {
+    score->text_cache[0].dynamic = true;
```
Vagabond, line 42: Unclear why this needs to be dynamic. Only the x/y position changes, maybe those should not be cached?

src/video/video.c
```diff
@@ -246,3 +248,10 @@ void video_draw_remap(const surface *src_surface, int x, int y, int remap_offset
     dst.y = y;
     draw_args(src_surface, &dst, remap_offset, remap_rounds, 0, 255, 255, 0, options);
 }
+
+void video_render_text_block(text_object *text) {
```
Vagabond, line 252: I'm not sure this belongs in video.c

nopjne: The idea here was that the video plugin can decide what to do with a text object.
For instance the N64 side doesn't traverse the letters if the object is not marked dirty. The code in here for video.c (PC) is unchanged from the original call pattern. If there is a better place that would allow the renderer to decide how to handle the text objects, I'm open to suggestions. This really seemed like the correct place. For instance the NULL renderer will just do nothing here too.

src/game/scenes/vs.c
nopjne: This problem can be seen when bringing up the in-game menu the second time through, it will show an empty spot where the "explanation" is on the N64 branch indicating that the text object wasn't correctly dirtied.

src/game/scenes/vs.c
nopjne: It looks like we free(local) so it's definitely necessary to destroy the text objects here. I agree there should be a text_objects_free here.

src/video/video.c
katajakasa: This is why we discussed the possible pluggable text renderer backend earlier. The plugin backend would decide how to call the video renderer instead.

src/game/gui/textinput.c
katajakasa: These text_caches should seriously be named variables.

src/game/utils/score.c
nopjne: Dynamic means something special, it means that the frame-over-frame rendering of the text object cannot be guaranteed to be the same. As the object may get occluded and require a re-render. Unfortunately I couldn't figure out a way to track this, so I resorted to hand picking objects that were NOT-SAFE to copy from the previous frame.

nopjne: This only really matters on the N64, where we can copy from the previous frame buffer, instead of allocating an intermediate texture.
The fallback would be to allocate the intermediate texture on the N64.
And can be completely ignored on the PC.

src/game/gui/textinput.c
Vagabond: I don't disagree, or at least used named indices

src/engine.c
```diff
@@ -150,6 +150,7 @@ void engine_run(engine_init_flags *init_flags) {
             // Handle other events
             switch(e.type) {
                 case SDL_QUIT:
+                    assert(false);
```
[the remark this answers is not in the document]
nopjne: Should be removed. Was left after debugging I was tring to find where the quit message was coming from on the N64 as that should never happen and somehow regressed after a merge.

src/game/utils/score.c
Vagabond: ah, ok, maybe we can have a better name than dynamic, or have a comment that explains it.

src/game/gui/textinput.c
nopjne: size_t is an unsigned int, but will be 64bit or 32bit based on your architecture. And it's not a signed int because I never expect to have a negative amount of objects to destroy.

src/game/gui/textinput.c
nopjne: I agree. it looks like the default settings are never changed per object.

src/game/gui/textinput.c
nopjne: Sounds good.

src/game/scenes/scoreboard.c
nopjne: Agreed.

src/game/gui/menu.c
nopjne: game_menu_return should be calling it.

src/video/video.c
Vagabond: I agree, but maybe we can do something intermediate for now, before we go to the "solves all the problems" pluggable text renderer.

katajakasa: I WANT IT NOW! We should throw in kubernetes too somehow ;)

src/game/gui/menu.c
Vagabond: Maybe I goofed the rebase?

src/video/video.c
nopjne: @katajakasa Would this be something you could rework later, after this change goes in?

katajakasa: Yes, it is. Let's get this part merged first.

src/game/scenes/vs.c
Vagabond: There's a game menu on the 'vs' screen?

src/game/gui/textinput.c
Vagabond: I mean why not just pass the number of text objects

Nopey: size_t is fine, but the arraysize magic should probably be replaced with a constant (same as the [0]'s, [1]'s, [2]'s and [3]'s, as pointed out in a different conversation)

src/game/gui/label.c
```diff
@@ -55,7 +57,7 @@ component *label_create_with_width(const text_settings *tconf, const char *text,
     widget_set_obj(c, local);
     widget_set_render_cb(c, label_render);
     widget_set_free_cb(c, label_free);
-
+    // local->text_cache[0].dynamic = true;
```
Vagabond, line 60: Remove

src/game/gui/menu.c
```diff
@@ -110,7 +110,11 @@ static void menu_render(component *c) {
             if(m->help_bg2) {
                 video_draw(m->help_bg2, m->help_x - 8, m->help_y - 8);
             }
-            text_render(&m->help_text_conf, TEXT_DEFAULT, m->help_x, m->help_y, m->help_w, m->help_h, (*tmp)->help);
+            // TODO: The text_cache should move into inner component. In this case the one of the s->objs.
```
Vagabond, line 113: I believe this comment is wrong and it's not invalidated every frame

src/game/gui/textinput.c
nopjne: This should become a macro called ARRAY_SIZE(tb->text_cache).
That makes it forward compatible, when someone expands or reduces the array, they won't need to change this instance.

src/game/gui/menu.c
nopjne: Could also be I missed it in the original PROTO submit.

src/game/scenes/vs.c
nopjne: Uhhh, that's interesting. I didn't realize I left this comment in the VS, it probably needs to be in arena somewhere.

src/game/gui/menu.c
nopjne: I think what I ran into was that the menu didn't animate, so I called all these dynamic.

src/game/gui/text_render.h
```diff
@@ -77,16 +78,46 @@ typedef struct {
     uint8_t max_lines;
 } text_settings;
 
+typedef struct letter {
+    int x;
+    int y;
+    int offset;
+    int limit;
```
Nopey, line 85:
```suggestion
    int palette_offset;
    int palette_limit;
```
Might also be possible to make these a smaller (8-bit) integer type, but that sounds like work for another PR.

src/game/gui/text_render.h
```diff
@@ -77,16 +78,46 @@ typedef struct {
...
+    surface *sur;
```
Nopey, line 86:
```suggestion
    surface const *sur;
```
Since this is not an owned surface, and we do not plan to modify the surface via this pointer.

src/game/gui/text_render.c
```diff
@@ -304,39 +375,162 @@ static void text_render_len(const text_settings *settings, text_mode mode, int x
...
+    dirty |= (memcmp(&cached_text->settings, settings, sizeof(cached_text->settings)) != 0);
+
+    if(cached_text->dirty || dirty) {
+        size_t text_len = strlen(text);
+        if(text_len == 0) {
+            return;
+        }
+
+        if(cached_text->max_letters < (text_len * 4)) {
+            size_t size = (text_len * sizeof(letter) * 4);
```
Nopey, line 468: What is magic number 4?

src/game/gui/text_render.c
```diff
@@ -80,27 +89,97 @@ static inline void render_char_surface(const text_settings *settings, text_mode
...
+    return sur->w;
 }
 
-int text_render_char(const text_settings *settings, text_mode state, int x, int y, char ch) {
-    surface *sur = get_font_surface(settings, ch);
-    render_char_shadow_surface(settings, sur, x, y);
-    render_char_surface(settings, state, sur, x, y);
-    return sur->w;
+void letter_set_parameters(text_object *cached_text, surface *sur, int x, int y, int offset, int limit) {
+    cached_text->cur_letter->x = x;
```
Nopey, line 99: Please add an assert that checks that `cached_text->cur_letter < cached_text->letter + cached_text->max_letters`

src/game/gui/text_render.c
Nopey: It looks like 4 is the number of draw calls per letter.
At the very least, it should be a named constant.

Perhaps all four draw calls should be stored in one `struct letter`? This would increase the number of fields in `struct letter`, but should decrease overall memory usage.

nopjne: It's the shadows, we support up to 3 shadows per letter, including the letter itself. Actually per the code we support 4 shadows per letter (+letter). But I've never seen more than 3 being used.

src/game/utils/score.c
nopjne: other names I've considered: not_copy_safe, needs_intermediate_texture, not_static_composition,

Nopey: is `uncachable` or `changes_frequently` appropriate?

nopjne: It's more of a hint to the renderer telling it not to 2nd level optimize.
0th lvl - render each letter.
1st lvl - allocate a new texture, render each letter to texture on dirty bit, use the allocated texture on all invocations without the dirty bit.
2nd lvl - render each letter to FB on dirty bit, copy from FB without dirty bit.

Vagabond:
Ok, I think I've gained some familiarity with this PR now, and I'd like to suggest some changes:

* I'd like the invalidation of a text cache to be explicit, not automatic. We can know when we invalidate text strings and update them appropriately.
* I'd like to separate flowing text into a text object and rendering a text object in the API.
* I'd like not to hold a pointer to the 'text' inside the text cache, since it should no longer be needed as per point 1
* It might be useful, in some cases, to render multiple strings to the same text_cache. A good example is the fight stats page in tournament mode, or the HAR/Player stats in the mechlab. If we aren't tracking strings anymore, this also seems doable. This would massively reduce the number of text cache objects we have to maintain.

cc @nopjne @katajakasa

Vagabond:
> > It might be useful, in some cases, to render multiple strings to the same text_cache. A good example is the fight stats page in tournament mode, or the HAR/Player stats in the mechlab. If we aren't tracking strings anymore, this also seems doable. This would massively reduce the number of text cache objects we have to maintain.
> 
> As long as that doesn't unnecessarily extend the bounds of the text object.

My idea would be that if we were flowing several chunks of text into the same text_object, then we could 'finish' the flow that would allow any underlying texture allocation stuff to happen, with the final set of bounds?

katajakasa:
Probably deprecated now ?

Vagabond:
Yes.

PR #774, opened 2024-11-20
Fix incorrect config string handling in menu_video.c

katajakasa:


src/game/scenes/mainmenu/menu_video.c
```diff
@@ -82,7 +82,12 @@ void resolution_toggled(component *c, void *userdata, int pos) {
 
 void renderer_toggled(component *c, void *userdata, int pos) {
     settings_video *v = &settings_get()->video;
-    video_get_renderer_info(pos, &v->renderer, NULL);
+    const char *renderer;
+    video_get_renderer_info(pos, &renderer, NULL);
+    if(v->renderer) {
+        omf_free(v->renderer);
```
Nopey, line 88: omf_free does not need to be guarded from NULL pointers.

katajakasa: Hah, yes, realized that myself one second after pushing ;)

Fixed now.

Nopey (approved):

PR #776, opened 2024-11-20
Use boolean in more places

katajakasa:
I got annoyed.

### DIFF
--- a/src/audio/audio.c
+++ b/src/audio/audio.c
@@ -75,7 +75,7 @@ static Mix_Chunk *audio_get_chunk(int id, float volume, float pitch) {
     SDL_AudioCVT cvt;
 
     // Load sample (8000Hz, mono, 8bit)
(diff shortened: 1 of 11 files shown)


src/audio/audio.c
```diff
@@ -75,7 +75,7 @@ static Mix_Chunk *audio_get_chunk(int id, float volume, float pitch) {
     SDL_AudioCVT cvt;
 
     // Load sample (8000Hz, mono, 8bit)
-    if(sounds_loader_get(id, &src_buf, &src_len) != 0) {
+    if(!sounds_loader_get(id, &src_buf, &src_len) != 0) {
```
Nopey, line 78: Please remove the ` != 0`

aside: I'm surprised clang's not linting this, i wonder if -Wall is doing its job or not

Nopey (commented):
other than that `if` in audio.c, looks good

PR #782, opened 2024-11-22
CMakeLists.txt: Make ENABLED_RENDER_PLUGINS an option the user can set

mrannanj:


Nopey:
I think this can be closed in favor of #782?

PR #783, opened 2024-11-22
Null renderer

mrannanj:


src/engine.c
```diff
@@ -39,6 +39,8 @@ int engine_init(void) {
     float music_volume = setting->sound.music_vol / 10.0;
     float sound_volume = setting->sound.sound_vol / 10.0;
     const char *renderer = setting->video.renderer;
+    if(*init_flags->renderer != '\0')
```
Vagabond, line 42: strlen maybe?

src/main.c
```diff
@@ -132,6 +131,8 @@ int main(int argc, char *argv[]) {
     } else if(rec->count > 0) {
         init_flags.record = 1;
         strncpy(init_flags.rec_file, rec->filename[0], 254);
+    } else if(renderer->count > 0) {
```
Vagabond, line 134: Might be better to parse the input here, and throw an error if the renderer is unknown/unavailable?

src/main.c
```diff
@@ -76,12 +74,13 @@ int main(int argc, char *argv[]) {
     struct arg_lit *vers = arg_lit0("v", "version", "print version information and exit");
     struct arg_lit *listen = arg_lit0("l", "listen", "Start a network game server");
     struct arg_str *connect = arg_str0("c", "connect", "<host>", "Connect to a remote game");
+    struct arg_str *renderer = arg_str0("X", "renderer", "<renderer>", "Renderer to use (default: opengl3)");
```
Nopey, line 77: Could you name this "force_renderer" or similar, to communicate that this is not the primary way of setting the renderer?
Same for the engine_init struct.

Nopey: I don't think this needs a one-letter shorthand either, feel free to replace "X" (which allows -X) with NULL

src/main.c
mrannanj: Well, that would duplicate things. Currently, if I run `/openomf --renderer foobar`, I get
```
[      0][D] video_scan_renderers(): Renderer 'OpenGL3' is available
[      0][D] video_scan_renderers(): Renderer 'NULL' is available
[      0][E] video_find_renderer(): Unable to find specified renderer 'foobar', trying other alternatives ...
[      0][I] video_find_renderer(): Found available renderer 'OpenGL3'!
```
So in this sense, all is fine already.

src/main.c
Nopey: Please do not specify the default as opengl3 here, that may not be true in the future

src/main.c
```diff
@@ -132,6 +131,8 @@ int main(int argc, char *argv[]) {
     } else if(rec->count > 0) {
         init_flags.record = 1;
         strncpy(init_flags.rec_file, rec->filename[0], 254);
+    } else if(renderer->count > 0) {
+        strncpy(init_flags.renderer, renderer->sval[0], sizeof(init_flags.renderer));
```
Nopey, line 135: This strncpy does not null terminate the output string with overlong --renderer flags
However, please leave this as-is (broken) until we have some saner alternatives to strncpy

Nopey (approved):

PR #784, opened 2024-11-22
Support changing video plugin in menu

Nopey:
When the videomode is changed, video_close and then video_init to apply it.
notify VGA state so it will be re-uploaded to the render plugin.

Also: replace all usages of C strdup with a new function, omf_strdup-- which is effectively the same as C strdup except it uses omf_malloc instead of malloc. This too lives in c_string_util.h


src/utils/c_string_util.c
```diff
@@ -1,4 +1,6 @@
 #include "utils/c_string_util.h"
+
```
mrannanj, line 2: This extra newline seems off?

src/video/vga_state.c
```diff
@@ -75,6 +75,11 @@ void vga_state_mark_remaps_flushed(void) {
     state.dirty_remaps = false;
 }
 
+void vga_state_mark_invalidate_render_plugin(void) {
```
katajakasa, line 78: vga_state_mark_dirty() instead ? Seems unnecessary to make this specific for renderer plugin validation.

katajakasa (changes requested):

katajakasa (approved):

PR #786, opened 2024-11-25
Improve str strings

Nopey:
str.h strings now store more characters in a small string than before, store the heap capacity for large strings (to avoid useless reallocs), and all this without increasing the size of struct str or additional heap usage.

Heavily inspired by Facebook's fbstring as described in a 2016 CppCon talk: https://youtu.be/kPR8h4-qZdk (although that has additional bells and whistles, including CoW semantics for very large strings).

### DIFF
--- a/src/utils/str.c
+++ b/src/utils/str.c
@@ -12,18 +12,47 @@
 #include <string.h>
 
 #define STR_STACK_SIZE sizeof(str)
+// NOTE: fbstring has 2 flag bits so they can signal their "large" (>255 byte)
+// string copy-on-write semantics. Hopefully we don't need that optimization.
 #define STR_FLAGBIT_COUNT 1
+
+// ------------------------ Memory Layout ------------------------
+
+// here are both str string representations, "normal" and then "small":
+//+------------+------------+------------+------------+------------+-----------+
+//| char *data              | size_t size             | capacity               |
+//+------------+------------+------------+------------+------------+-----------+
+//| char small[]                                                   | spare     |
+//+------------+------------+------------+------------+------------+-----------+
+//
+// `spare` is shorthand for referring to the last index of the small array, and
+// will become the null terminating byte when the small string is at capacity.
+// In a small string, `spare` counts the unused bytes.. the spare capacity.
+//
+// There is a flag bit to signal small/normal at the very end of the structure,
+// overlapping both capacity and spare. Because small strings must have a null
+// byte as their final byte for termination, a flag bit of 0 signals "small,"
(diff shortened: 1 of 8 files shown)


src/utils/compat.c
```diff
@@ -12,3 +12,10 @@ char *strdup(const char *s) {
     return d;
 }
 #endif
+
+#ifndef _WIN32
```
katajakasa, line 16: Should we let cmake do this check, just like HAVE_STD_STRDUP ?

src/utils/str.c
```diff
@@ -10,93 +11,172 @@
...
+
+// the number of spare capacity in the small string representation.
+// also contains the flag bits for detecting small/large strings.
+static inline uint8_t str_smallspare(str const *s) {
+    uint8_t spare;
+    memcpy(&spare, &s->small[STR_STACK_SIZE - 1], 1);
+    return spare;
+}
+
+static inline size_t str_normalcapacity(str const *s) {
```
katajakasa, line 27: Some comments to these on what happens and WHY would be excellent.

testing/test_str.c
```diff
@@ -1,12 +1,24 @@
 #include <CUnit/Basic.h>
 #include <CUnit/CUnit.h>
+#include <assert.h>
 #include <utils/str.h>
 
+// NOTE: this is an implementation detail, so shouldn't really be covered by unit tests. oh well!
```
katajakasa, line 6: The unit-tests were made to test implementation details.

katajakasa (changes requested):

src/utils/str.c
Nopey: Documented the memory layout, and added a bunch of comments throughout that section.

src/utils/compat.c
katajakasa: No, I mean drop the define check and rename the function omf_strnlen_s or somesuch.

And yes, If possible, we should keep the platform-specific define-checks to minimum. I realize it's not possible to remove them all, but let's at least try. Otherwise we are going to end up macro-soup.

Nopey: Sounds good, I'll provide it as omf_strnlen_s in c_string_util.h

Nopey:
image
I checked that msvc can eliminate the strlen calls by disassembling ai_controller.c's is_special_move, but if we think one of the compilers might not manage the same I can switch from `inline static` methods to functionlike-macros in str.h.

(Debug builds of course, do not eliminate the call.)

katajakasa (approved):

PR #787, opened 2024-11-25
CMakeLists.txt: Do not recover from sanitizer errors

mrannanj:


CMakeLists.txt
```diff
@@ -56,6 +56,8 @@ if(USE_SANITIZERS)
     else()
         add_compile_options("-fsanitize=address,undefined")
         add_link_options("-fsanitize=address,undefined")
+        add_compile_options("-fno-sanitize-recover=all")
```
Vagabond, line 59: Will this then crash on any leaks as well?

mrannanj: It will, but my intent is to only have this on with the unit tests currently.

testing/cunit_helpers.h
```diff
@@ -0,0 +1,17 @@
+#ifndef CUNIT_HELPERS_H
+#define CUNIT_HELPERS_H
+
+#include <CUnit/Basic.h>
+#include <CUnit/CUnit.h>
+
+#define CU_BUFFER_SIZE 256
+
+#define CU_ASSERT_INT_EQUAL(actual, expected)                                                                          \
```
Nopey, line 9: I would avoid adding anything to CUnit's CU namespace. Perhaps this should be OMF_CU_ASSERT_INT_EQUAL?

mrannanj: cunit stuff is dropped from this.

CMakeLists.txt
mrannanj: So now, these options are under USE_FATAL_SANITIZERS

PR #791, opened 2024-11-26
Use str.h for error message formatting in languagetool

Nopey:
These strings are leaked, right as we `exit(EXIT_FAILURE)` the program.


Nopey:
Did not help martti's stack overflow, so the char[100]s are not the culprit

PR #793, opened 2024-11-29
Fix some bugs, mostly crashes

Nopey:
Mostly from running demo with warp speed until it crashes, then fixing the crash and repeating.

Changes that are not related to crashes:
- "Drop ticks that are extremely late" ensures that the game doesn't try to catch up for more than 100ms worth of ticks-- this solves the issue where the game goes warpspeed after hitting a debugger breakpoint or other extreme stall.
- Make stamina regenerate about three thousand times faster than before, scaling it by the endurance_max. Neither the old code nor the new code is truly faithful to the original game, but this makes it regenerate at a somewhat similar rate to the original (whereas before, the effect was not noticable)

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -1452,8 +1452,9 @@ void har_tick(object *obj) {
     // to show the sprite with animation string that interpolates opacity down
     // Mark new object as the owner of the animation, so that the animation gets
     // removed when the object is finished.
-    if(object_has_effect(obj, EFFECT_TRAIL) && obj->age % 2 == 0) {
-        sprite *cur_sprite = animation_get_sprite(obj->cur_animation, obj->cur_sprite_id);
+    sprite *cur_sprite;
+    if(object_has_effect(obj, EFFECT_TRAIL) && obj->age % 2 == 0 &&
+       (cur_sprite = animation_get_sprite(obj->cur_animation, obj->cur_sprite_id))) {
         sprite *nsp = sprite_copy(cur_sprite);
         surface_flatten_to_mask(nsp->data, 1);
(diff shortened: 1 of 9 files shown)


src/video/vga_state.c
```diff
@@ -60,7 +60,7 @@ void vga_state_render(void) {
         for(unsigned int i = 0; i < state.transformer_count; i++) {
             state.transformers[i].callback(&tmp, &state.current, state.transformers[i].userdata);
         }
-        damage_copy(&state.dmg_current, &tmp);
+        damage_combine(&state.dmg_current, &tmp);
```
katajakasa, line 63: Doesn't this cause the dmg_current to always include full palette, because the ranges can only grow ?

Nopey: no, because of `vga_state_mark_palette_flushed` (called from opengl3):
```c
void vga_state_mark_palette_flushed(void) {
    damage_reset(&state.dmg_current);
}
```

src/game/objects/har.c
```diff
@@ -2024,9 +2024,6 @@ void har_finished(object *obj) {
         h->state = STATE_DONE;
         har_set_ani(obj, ANIM_VICTORY, 0);
     } else if(h->state == STATE_VICTORY || h->state == STATE_DONE) {
-        // end the arena
-        DEBUG("ending arena!");
-        game_state_set_next(obj->gs, SCENE_MENU);
```
katajakasa, line 2028: Something somewhere else still handles this ?

Nopey: arena_end handles this.

katajakasa (approved):

PR #798, opened 2024-12-01
Separate menu input from per-player input

### LINKED ISSUE #265: Arrow keys should work in menu, regardless of player one key binds.
Arrow keys should work in all menus, regardless of how player one or player two keys have been bound.

### LINKED ISSUE #266: ESC key should immediately return in main menu
ESC key should immediately return to previous menu in main menu. Only in the base level of the mainmenu should ESC move the selection to the last item in menu.

### LINKED ISSUE #499: Demo mode can't be stopped with using only gamepads / joysticks
Hi,

It would seem demo mode where the cpu's fight each other, can't be stopped if the only input device is a game pad / joystick.

I don't know if this was also the case in the original game, but i though i'd mention it. I was testing using latest git on a RG35XX handheld which only had gamecontroller support (i edited the conf files to be able to control game) but it seems at least the demo mode i can't seem to exit using a gamecontroller only

Nopey:
The only things that are still controlled by per-player input (controller_poll) are Melee, Vs, and Arena.
Everything else now uses `game_state_menu_poll`, a drop-in replacement for `controller_poll` that lets all detected input devices control the menus.

Users are unable to rebind the keyboard menu controls, to avoid complications around textinput, multiple keys performing the same action, and ensuring no single key performs multiple actions.
 
All detected game controllers/joysticks generate menu input events, even when no player has a joystick bound.

Closes #499 
Closes #266 
Closes #265

### DIFF
--- a/src/controller/controller.c
+++ b/src/controller/controller.c
@@ -55,6 +55,23 @@ void controller_free(controller *ctrl) {
     ctrl->free_fun(ctrl);
 }
 
+static inline void ctrl_action_push(ctrl_event **ev, int action) {
+    ctrl_event *new = omf_calloc(1, sizeof(ctrl_event));
+
+    new->type = EVENT_TYPE_ACTION;
+    new->event_data.action = action;
+
+    if(*ev == NULL) {
+        *ev = new;
+    } else {
+        ctrl_event *i = *ev;
+        while(i->next) {
+            i = i->next;
+        }
+        i->next = new;
+    }
+}
+
 void controller_cmd(controller *ctrl, int action, ctrl_event **ev) {
     ctrl->current |= action;
     if((ctrl->last & action) && (!ctrl->repeat || action == ACT_KICK || action == ACT_PUNCH || action == ACT_ESC)) {
@@ -64,27 +81,12 @@ void controller_cmd(controller *ctrl, int action, ctrl_event **ev) {
     // fire any installed hooks
     iterator it;
     hook_function **p = 0;
-    ctrl_event *i;
-    ctrl_event *new;
-
     list_iter_begin(&ctrl->hooks, &it);
     while((p = iter_next(&it)) != NULL) {
         ((*p)->fp)((*p)->source, action);
     }
 
-    new = omf_calloc(1, sizeof(ctrl_event));
-    new->type = EVENT_TYPE_ACTION;
-    new->event_data.action = action;
-
-    if(*ev == NULL) {
-        *ev = new;
-    } else {
-        i = *ev;
-        while(i->next) {
-            i = i->next;
-        }
-        i->next = new;
-    }
+    ctrl_action_push(ev, action);
 }
 
 void controller_close(controller *ctrl, ctrl_event **ev) {
--- a/src/controller/controller.c
+++ b/src/controller/controller.c
@@ -56,6 +56,11 @@ void controller_free(controller *ctrl) {
 }
 
 void controller_cmd(controller *ctrl, int action, ctrl_event **ev) {
+    ctrl->current |= action;
+    if((ctrl->last & action) && (!ctrl->repeat || action == ACT_KICK || action == ACT_PUNCH || action == ACT_ESC)) {
+        return;
+    }
+
     // fire any installed hooks
     iterator it;
     hook_function **p = 0;
--- a/src/controller/controller.c
+++ b/src/controller/controller.c
@@ -95,8 +95,9 @@ void controller_cmd(controller *ctrl, int action, ctrl_event **ev) {
     iterator it;
     hook_function **p = 0;
     list_iter_begin(&ctrl->hooks, &it);
-    while((p = iter_next(&it)) != NULL) {
-        ((*p)->fp)((*p)->source, action);
+    while((p = iter_next(&it))) {
+        hook_function hook = **p;
+        (hook.fp)(hook.source, action);
     }
 
     ctrl_action_push(ev, action);
--- a/src/controller/controller.c
+++ b/src/controller/controller.c
@@ -74,9 +74,22 @@ static inline void ctrl_action_push(ctrl_event **ev, int action) {
 
 void controller_cmd(controller *ctrl, int action, ctrl_event **ev) {
     ctrl->current |= action;
-    if((ctrl->last & action) && (!ctrl->repeat || action == ACT_KICK || action == ACT_PUNCH || action == ACT_ESC)) {
-        return;
+    if((ctrl->last & action)) {
+        if(action & (ACT_KICK | ACT_PUNCH | ACT_ESC))
+            // never keyrepeat these actions
+            return;
+        else if(ctrl->repeat)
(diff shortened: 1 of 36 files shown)


src/controller/controller.c
```diff
@@ -55,33 +55,53 @@ void controller_free(controller *ctrl) {
...
+    }
+    if(action & ACT_Mask_Dirs)
+        ctrl->repeat_tick = 30;
+
+    // fire any installed hooks
+    iterator it;
+    hook_function **p = 0;
+    list_iter_begin(&ctrl->hooks, &it);
+    while((p = iter_next(&it)) != NULL) {
+        ((*p)->fp)((*p)->source, action);
```
katajakasa, line 99: This honestly looks like someone puked ASCII ;) Could you at least move the function-ptr to a variable or something ?

src/controller/joystick.c
```diff
@@ -205,3 +211,63 @@ int joystick_create(controller *ctrl, int joystick_id) {
...
+    }
+    vector_free(&every_gamepad);
+}
+
+void joystick_menu_poll_all(controller *menu_ctrl, ctrl_event **ev) {
+    if(vector_size(&every_gamepad) == 0)
+        return;
+
+    joystick k;
+    memset(&k, 0, sizeof k);
```
katajakasa, line 241: sizeof() for consistenly

src/game/gui/menu.c
```diff
@@ -184,9 +188,11 @@ static int menu_action(component *mc, int action) {
             c = sizer_get(mc, m->selected);
 
         } while(component_is_disabled(c));
-        // Play menu sound
-        audio_play_sound(19, 0.5f, 0.0f, 2.0f);
-        component_select(c, 1);
+        if(c != old_c) {
+            // Play menu sound
+            audio_play_sound(19, 0.5f, 0.0f, 2.0f);
```
katajakasa, line 193: If you feel like it, there could be either
a) Constants for the default values here, or
b) audio_play_sound_with_defaults(19) or somesuch

This is not necessarily part of this PR, just came to mind while reading.

src/utils/vector.c
```diff
@@ -109,14 +109,26 @@ int vector_delete_at(vector *vec, unsigned index) {
...
-    if(vec->blocks > 0) {
-        vec->blocks--;
-    }
+    vec->blocks--;
 
     // Return success
     return 0;
 }
 
+int vector_swapdelete_at(vector *vec, unsigned index) {
```
katajakasa, line 118: The name feels a bit, but I couldn't come up with anything better. I considered making this the default, but I think we rely on vector keeping the order in most places. Blah.

src/game/scenes/arena.c
```diff
@@ -1071,25 +1054,41 @@ void arena_static_tick(scene *scene, int paused) {
...
+        arena_handle_events(scene, player2, p2);
+        controller_free_chain(p1);
+        controller_free_chain(p2);
     }
-    return 0;
+
+    ctrl_event *menu_ev = NULL;
+    game_state_menu_poll(scene->gs, &menu_ev);
+
+    for(ctrl_event *i = menu_ev; i; i = i->next) {
```
katajakasa, line 1076: Something to consider later: https://github.com/omf2097/openomf/issues/800

src/utils/vector.h
```diff
@@ -23,6 +23,8 @@ int vector_prepend(vector *vector, const void *value);
 void vector_sort(vector *vector, vector_compare_func cf);
```
katajakasa, line 23: Something to do later: https://github.com/omf2097/openomf/issues/801

katajakasa (dismissed):
Probably better than it was. Might be a good idea to get an okay from @Vagabond too.

mrannanj (approved):

mrannanj (dismissed):

Vagabond:
The highlight on player 2 pilot select no longer renders properly.

Vagabond:
Also, if you return from HAR select to pilot select in 2 player, the inactive pilot portraits have a corrupted palette.

Nopey:
> The highlight on player 2 pilot select no longer renders properly.

was a regression on master for a minute, will go away with my rebase in a moment

> palette corruption

i'll look into this, thanks

Nopey:
rebased, fixed the two bugs in melee scene.

Nopey:
@Vagabond, I take it you're not opposed to the addition of `game_state_menu_poll` and the new behavior of allowing all known inputs (keyboard, joysticks at the moment) to control menus?

Vagabond:
> @Vagabond, I take it you're not opposed to the addition of `game_state_menu_poll` and the new behavior of allowing all known inputs (keyboard, joysticks at the moment) to control menus?

I don't think so, I was originally concerned with how this would affect netplay, so I'd like to try this branch now it's been rebased with netplay to make sure everything still works.

Nopey:
What is the current state of menus in netplay?
For the rest of this post, I'm assuming the only effected menu is the in-match arena menu-- correct me if that's not true.

From reading what the code was prior to this PR, whoever "player 1" was (probably the person who initiated the fight) is able to pull up & control the in-game menu for both players' games, and "player 2" is unable to do any menuing. Is this a correct assessment? I've been unable to enter a match with oomf netplay (to-do at a later date: figure out why).

My expectation of how the game will break with this PR: either player can pause their own game, and the remote client will then desync (as there is no networking of the pause, as it's no longer based on player 1's controller inputs).

<hr>
What do other multiplayer 2p fighting games do?

Some alternatives to the "only let player 1 use the pause menu" status quo ordered by ascending programmer-effort:
1. Don't pause netplay games during the in-game menu. Common in first person shooters & other games with more than 2 players. Dead simple to implement, probably the easiest way to unblock this PR.
2. Pause both players' games, and show a "other player has paused" message for the player who isn't menuing-- have a 3 second countdown to resume gameplay once both players are not menuing. The pause event would probably be a new controller event, in the vein of other non-button controller events like EVENT_TYPE_HB (net controller heartbeat).
3. Same as previous, but also network where the player is in the menu-- this is like the Street Fighter 6 "Play Again?" after match UI where players' cursor is networked. Has the benefit of letting the waiting player know if their opponent has paused & walked away or if their opponent is doing something in the menus.

Vagabond:
I believe #1 should be fine. I don't feel like we need anything fancier.

src/game/game_state.c
```diff
@@ -618,6 +624,7 @@ void game_state_call_move(game_state *gs) {
 }
 
 void game_state_tick_controllers(game_state *gs) {
+    controller_tick(gs->menu_ctrl, gs->int_tick, &gs->menu_ctrl->extra_events);
```
Vagabond, line 627: So the menu is only statically ticked?

Nopey: Yes, that was my intention.
Dynamic ticks speed up and slow down during hitstun and other gameplay things.

Vagabond (approved):
This looks a lot more consistent than before, and everything still seems to work.

PR #799, opened 2024-12-01
Null audio

mrannanj:


src/formats/internal/memreader.c
```diff
@@ -49,7 +49,7 @@ void memreader_close(memreader *reader) {
 }
 
 int memread_buf(memreader *reader, char *buf, int len) {
-    if(reader->pos + len > reader->len) {
+    if(reader->pos + len > reader->len || reader->buf == NULL || buf == NULL) {
```
katajakasa, line 52: hum. Should this assert instead ?

mrannanj: Well, sure, I'll change to that.

katajakasa (approved):

PR #808, opened 2024-12-09
game: Enable network menu in debug mode only

mrannanj:


src/game/scenes/mainmenu.c
```diff
@@ -138,6 +138,7 @@ int mainmenu_create(scene *scene) {
     scene_set_dynamic_tick_cb(scene, mainmenu_tick);
     scene_set_startup_cb(scene, mainmenu_startup);
 
+#ifdef DEBUGMODE
     if(scene->gs->net_mode == NET_MODE_CLIENT) {
         component_action(guiframe_find(local->frame, NETWORK_BUTTON_ID), ACT_PUNCH);
         component_action(guiframe_find(local->frame, NETWORK_CONNECT_BUTTON_ID), ACT_PUNCH);
```
Nopey, line 144: Please do not break the net_mode field, this is set via commandline options that this PR is not removing.
I think leaving the network game launch options in is fine, as most users will not discover them-- and they will probably be useful when we need to check if netplay works with CI pipeline-built binaries.

src/game/scenes/mainmenu/menu_main.c
```diff
@@ -105,9 +105,11 @@ component *menu_main_create(scene *s) {
     menu_attach(menu, textbutton_create(&tconf, "ONE PLAYER GAME", NULL, COM_ENABLED, mainmenu_1v1, s));
     menu_attach(menu, textbutton_create(&tconf, "TWO PLAYER GAME", NULL, COM_ENABLED, mainmenu_1v2, s));
     menu_attach(menu, textbutton_create(&tconf, "TOURNAMENT PLAY", NULL, COM_ENABLED, mainmenu_mechlab, s));
+#ifdef DEBUGMODE
     component *net = textbutton_create(&tconf, "NETWORK PLAY", NULL, COM_ENABLED, mainmenu_enter_network, s);
     widget_set_id(net, NETWORK_BUTTON_ID);
     menu_attach(menu, net);
+#endif
```
Nopey, line 112: Maybe worth putting an `#else` with a disabled "NETWORK PLAY" button, similar to the "HELP" menu we haven't implemented yet?

Nopey (changes requested):

src/game/scenes/mainmenu.c
mrannanj: Okay, I let the `scene->gs->net_mode = NET_MODE_NONE;` stay, but it doesn't affect anything, as it is 0 anyway. About the CI build, that uses debug mode for testing, so don't need it there..

mrannanj: Oh yeah, and disabled the command line stuff too here.

src/game/scenes/mainmenu/menu_main.c
mrannanj: Added the disabled network play item.

Nopey (approved):
Unsure what the motivation for removing the launch options is (`--connect` and `--port`), but the actual code changes to accomplish this are fine.

I think leaving the network game launch options in is a good idea, as they will probably be useful when we need to check if netplay works with CI pipeline-built binaries. (and users are unlikely to discover them in error)

Nopey (approved):

Vagabond (approved):
I haven't tested this but it looks correct.

PR #820, opened 2025-01-28
Don't fail to parse the config file due to an unknown option

Vagabond:
If we add or remove config options, confuse will throw an error loading the config file, and replace it with a default one. This patch enables the confuse option to not do that, and allow the config to load.

See https://github.com/libconfuse/libconfuse/pull/41


Vagabond:
I don't know if we also need this flag when writing back out the config.

katajakasa (approved):

PR #823, opened 2025-01-28
Fix pilot gender in tournament mode

Vagabond:
When loading the CHR file, load the gender from the PIC file using the 10th byte of the unknown enemy footer.

For the tournament player, consult the PLAYERS.PIC for the gender.

### DIFF
--- a/src/formats/chr.c
+++ b/src/formats/chr.c
@@ -195,7 +195,7 @@ int sd_chr_load(sd_chr_file *chr, const char *filename) {
         chr->enemies[i]->photo_id = memread_ubyte(mr);
         memread_buf(mr, chr->enemies[i]->unknown_b, 15);
         if(trn_loaded) {
-            chr->enemies[i]->pilot.sex = pic.photos[(uint8_t)chr->enemies[i]->photo_id]->sex;
+            chr->enemies[i]->pilot.sex = pic.photos[(uint16_t)chr->enemies[i]->photo_id]->sex;
         }
         for(int m = 0; m < 10; m++) {
             if(trn_loaded && trn.enemies[i]->quotes[m]) {
--- a/src/formats/chr.c
+++ b/src/formats/chr.c
@@ -190,9 +190,13 @@ int sd_chr_load(sd_chr_file *chr, const char *filename) {
             chr->enemies[i]->pilot.winnings = trn.enemies[i]->winnings;
             chr->enemies[i]->pilot.total_value = trn.enemies[i]->total_value;
             chr->enemies[i]->pilot.photo_id = trn.enemies[i]->photo_id;
-            chr->enemies[i]->pilot.sex = pic.photos[(uint8_t)chr->enemies[i]->unknown[9]]->sex;
         }
-        memread_buf(mr, chr->enemies[i]->unknown, 25);
+        memread_buf(mr, chr->enemies[i]->unknown_a, 9);
+        chr->enemies[i]->photo_id = memread_ubyte(mr);
+        memread_buf(mr, chr->enemies[i]->unknown_b, 15);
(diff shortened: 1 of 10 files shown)


src/resources/pilots.c
```diff
@@ -4,17 +4,20 @@
 void pilot_get_info(pilot *pilot, int id) {
     // TODO read this from MASTER.DAT
     // XXX the colors are eyeballed
+    pilot->sex = 0;
```
Nopey, line 7: Shouldn't this be using PILOT_SEX_MALE and PILOT_SEX_FEMALE instead of 0 and 1?

src/resources/pilots.c
```diff
@@ -4,17 +4,20 @@
...
     switch(id) {
         case 0:
+            // CRYSTAL
```
Nopey, line 10:
```suggestion
        case PILOT_CRYSTAL:
```

Use the enums instead of adding comments, please

src/formats/chr.c
```diff
@@ -175,6 +190,7 @@ int sd_chr_load(sd_chr_file *chr, const char *filename) {
             chr->enemies[i]->pilot.winnings = trn.enemies[i]->winnings;
             chr->enemies[i]->pilot.total_value = trn.enemies[i]->total_value;
             chr->enemies[i]->pilot.photo_id = trn.enemies[i]->photo_id;
+            chr->enemies[i]->pilot.sex = pic.photos[(uint8_t)chr->enemies[i]->unknown[9]]->sex;
```
Nopey, line 193: Please name unknown[9]

Nopey (changes requested):

katajakasa (approved):

src/formats/chr.c
```diff
@@ -176,7 +191,12 @@ int sd_chr_load(sd_chr_file *chr, const char *filename) {
             chr->enemies[i]->pilot.total_value = trn.enemies[i]->total_value;
             chr->enemies[i]->pilot.photo_id = trn.enemies[i]->photo_id;
         }
-        memread_buf(mr, chr->enemies[i]->unknown, 25);
+        memread_buf(mr, chr->enemies[i]->unknown_a, 9);
+        chr->enemies[i]->photo_id = memread_ubyte(mr);
+        memread_buf(mr, chr->enemies[i]->unknown_b, 15);
+        if(trn_loaded) {
+            chr->enemies[i]->pilot.sex = pic.photos[(uint8_t)chr->enemies[i]->photo_id]->sex;
```
Nopey, line 198: is there a reason you're casting the uint16_t photo_id to a uint8_t?
if overreading the photos buffer is the concern, please perform a check against MAX_PIC_PHOTOS.

Vagabond: This is a great question, I don't actually know the size of the field in the CHR file. That said, I don't think any PIC files with more than 255 entries exist..

Vagabond, line 198:
```suggestion
            chr->enemies[i]->pilot.sex = pic.photos[(uint16_t)chr->enemies[i]->photo_id]->sex;
```

Vagabond: I made it a uint16_t now, better?

Nopey: Why are you casting?

Vagabond: It was to fix a compiler warning.

Nopey: I ran a clang and gcc build (using CI, see PR #828) and do not see a compiler warning.

I suspect the warning was solved when you broke `char unknown[25]` into three: unknown_a, photo_id, and unknown_b-- and the cast is a vestige of the unknown char array's signedness.

Vagabond: You seem to be correct.

PR #824, opened 2025-01-28
Fix game version in CMakeLists

Vagabond:


CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 cmake_minimum_required(VERSION 3.16)
 
 set(VERSION_MAJOR "0")
-set(VERSION_MINOR "6")
-set(VERSION_PATCH "6")
+set(VERSION_MINOR "7")
```
katajakasa, line 4: master should probably be 0.8.0

Nopey:
Could we bring the two sources of truth (currently, the github CI workflow and the cmake script) together?
the way it is done now is error prone.

PR #827, opened 2025-01-30
Hack to make player 2 in netplay AI

Vagabond:


Vagabond:
This probably could be put behind some kind of command line flag and merged. It is very helpful for testing netplay without another person.

PR #829, opened 2025-01-31
formats: Truncate tournament description to registration fee line

mrannanj:
Otherwise, War Invitational includes part of the next string.


Vagabond:
That's part of the original string, we should simply be parsing out that markup.

PR #830, opened 2025-01-31
Text objects rebased

mrannanj:
This is a rebased version of PR #773
Replaces PRs #773 and #677


Nopey:
Might be better to push your continuation of this work to the existing PR 773 (branch [adt/text-objects](https://github.com/omf2097/openomf/tree/adt/text-objects))?

As there are unresolved conversations on the existing PR

PR #831, opened 2025-02-02
Simplify netplay and remove bugs

Vagabond:
Fixed an unsafe clone of the scene ticktimer, removed some more obsolete code, removed overcomplicated 'seen_peer' event tracking and made sure when we save the game state we've applied any inputs for that frame.


Vagabond:
Merging this because it works way better than the old code.

PR #837, opened 2025-02-05
github: compilation.yml - Set OPENOMF_VERSION based on git tag

mrannanj:


Vagabond:
Looks like `git describe --tags --match "[0-9]*.[0-9]*.[0-9]*"` works, which will find only exact x.y.z numeric version tags, if we want.

CMakeLists.txt
```diff
@@ -132,6 +126,27 @@ if(GIT_FOUND)
...
+        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+        OUTPUT_VARIABLE GIT_VERSION
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    string(REGEX MATCH "^([0-9]+)\\.([0-9]+)\\.([0-9]+)(-.*)?" GIT_VERSION ${GIT_VERSION})
+    set(VERSION_MAJOR ${CMAKE_MATCH_1})
+    set(VERSION_MINOR ${CMAKE_MATCH_2})
+    set(VERSION_PATCH ${CMAKE_MATCH_3})
+    set(VERSION_LABEL ${CMAKE_MATCH_4})
+    add_definitions(-DV_MAJOR=${VERSION_MAJOR} -DV_MINOR=${VERSION_MINOR} -DV_PATCH=${VERSION_PATCH})
```
Vagabond, line 141: We should add a definition for label as well.

.github/workflows/compilation.yml
```diff
@@ -263,6 +267,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: '0'
+
+    - name: Set OPENOMF_VERSION
+      run: echo "OPENOMF_VERSION=$(git describe --tags --match '[0-9]*.[0-9]*.[0-9]*')" >> "$GITHUB_ENV"
```
Nopey, line 274: (if we're keeping git describe)
Could you leave a comment stating that `--match` does not take a regex, but rather a glob(7)?
https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---matchltpatterngt
https://linux.die.net/man/7/glob

could use a glob that looks less like a regex, too, like `--match '[[:digit:]]*'; this glob matches any tag starting with a digit, and omits the check for periods for brevity.

PR #846, opened 2025-02-07
Show new version labels, where appropriate

Vagabond:


CMakeLists.txt
```diff
@@ -153,7 +153,7 @@ if(GIT_FOUND)
         math(EXPR VERSION_PATCH "${VERSION_PATCH} + 1")
     endif()
     add_definitions(-DV_MAJOR=${VERSION_MAJOR} -DV_MINOR=${VERSION_MINOR}
-                    -DV_PATCH=${VERSION_PATCH} -DV_LABEL=${VERSION_LABEL})
+                    -DV_PATCH=${VERSION_PATCH} -DV_LABEL="${VERSION_LABEL}")
```
mrannanj, line 156: Well, could put quotes for all of them for consistency

src/game/scenes/lobby.c
```diff
@@ -641,10 +641,10 @@ void lobby_entered_name(component *c, void *userdata) {
             event.peer->data = nat;
             strncpy_or_truncate(local->name, textinput_value(c), sizeof(local->name));
 
-            char version[15];
+            char version[30];
             // TODO support git version when not on a tag
-            snprintf(version, 14, "%d.%d.%d", V_MAJOR, V_MINOR, V_PATCH);
-            version[14] = 0;
+            snprintf(version, 29, "%d.%d.%d%s", V_MAJOR, V_MINOR, V_PATCH, V_LABEL);
+            version[29] = 0;
```
mrannanj, line 647: Using `snprintf(version, sizeof(version), ..` would make the `version[29] = 0` redundant here and generally sizeof is nicer to use than magic constants.

src/game/scenes/lobby.c
```diff
@@ -899,7 +899,7 @@ void lobby_tick(scene *scene, int paused) {
                         user.losses = serial_read_int8(&ser);
                         user.status = serial_read_int8(&ser);
                         uint8_t version_len = serial_read_int8(&ser);
-                        if(version_len < 15) {
+                        if(version_len < 30) {
```
mrannanj, line 902: ugh, maybe this 30 needs to be some `#define VERSION_BUF_LEN` or such?

mrannanj (changes requested):

CMakeLists.txt
Nopey, line 156: Please do not use `add_definitions` to introduce V_LABEL to every sourcefile.
V_LABEL is going to change every time the latest commit hash changes, and then every file will recompile because the macro changed.

Please do what the SHA1_HASH macro does instead, where it is added to a specific C file (or maybe multiple files, in this case?).

as V_LABEL is used in multiple sourcefiles, you might consider a new `.c` file that exports a `extern char const *const openomf_version_label;` in `.h`?

mrannanj (dismissed):

CMakeLists.txt
```diff
@@ -153,7 +153,9 @@ if(GIT_FOUND)
         math(EXPR VERSION_PATCH "${VERSION_PATCH} + 1")
     endif()
     add_definitions(-DV_MAJOR=${VERSION_MAJOR} -DV_MINOR=${VERSION_MINOR}
-                    -DV_PATCH=${VERSION_PATCH} -DV_LABEL=${VERSION_LABEL})
```
Vagabond, line 156: I wonder if we should move all of these to source file properties?

PR #853, opened 2025-02-08
Type improvements

katajakasa:


mrannanj (approved):

turol:
Greenspun's 10th rule:

"Any sufficiently complicated C or Fortran program contains an ad hoc, informally-specified, bug-ridden, slow implementation of half of Common Lisp."

Or in this case, the C++ standard library. You could just switch to C++ (or Rust :smile:) and delete almost 10% of the code.

katajakasa:
> Greenspun's 10th rule:
> 
> "Any sufficiently complicated C or Fortran program contains an ad hoc, informally-specified, bug-ridden, slow implementation of half of Common Lisp."
> 
> Or in this case, the C++ standard library. You could just switch to C++ (or Rust 😄) and delete almost 10% of the code.

Right, but then we would have to add that 10% back as C++ boilerplate ;)

PR #854, opened 2025-02-09
Allow disabling libpng

katajakasa:
This should help with N64 and other systems where supporting libpng is either impossible or not a priority.


src/formats/error.c
```diff
@@ -27,6 +27,8 @@ const char *sd_get_error(enum SD_ERRORCODE error_code) {
             return "File could not be read";
         case SD_FILE_UNLINK_ERROR:
             return "File could not be unlinked";
+        case SD_FAILURE:
```
Vagabond, line 30: Maybe unsupported might be clearer?

Vagabond (approved):

src/utils/png_reader.c
```diff
@@ -0,0 +1,94 @@
...
+    for(int y = 0; y < h; y++) {
+        omf_free(row_pointers[y]);
+    }
+    omf_free(row_pointers);
+error_2:
+    png_destroy_read_struct(&png_ptr, &info_ptr, NULL);
+error_1:
+    fclose(handle);
+error_0:
+    return false;
```
mrannanj, line 84: Always returning false seems pessimistic? Is there no way this succeeds?

katajakasa: Yes, there is. Looks like I somehow removed return true; from the function.

src/formats/error.c
katajakasa: I don't know if it matters. The default will be that libpng is enabled; if someone disables it, then it's a known case.

I sort of want to get rid of these codes completely and just abort() on problems :/

src/utils/png_reader.h
```diff
@@ -0,0 +1,9 @@
+#ifndef PNG_READER_H
+#define PNG_READER_H
+
+#include "video/vga_palette.h"
+#include <stdbool.h>
+
+bool png_read_paletted(const char *filename, unsigned char *dst);
```
Nopey, line 7: Perhaps your png_reader methods should return SD_ERRORCODE instead of boolean?
then we can have real errorcodes rather than just "General Error"

Nopey, line 7: Please do not create a function that starts with `png_`, as that's what the real libpng functions start with, such as png_read_png.

CMakeLists.txt
```diff
@@ -247,11 +248,18 @@ target_link_libraries(openomf_core PRIVATE
...
     openomf::SDL2
     openomf::SDL2_mixer
     openomf::xmp
 )
+
+if (USE_LIBPNG)
+    target_link_libraries(openomf_core PUBLIC
+        openomf::png
+    )
+    add_definitions(-DUSE_LIBPNG=1)
```
Nopey, line 260: Existing C macros for library detection are of the form: MINIUPNPC_FOUND, ARGTABLE3_FOUND, and NATPMP_FOUND.

Perhaps this macro should be PNG_FOUND instead of USE_LIBPNG?

Nopey (changes requested):

src/formats/vga_image.c
```diff
@@ -62,103 +63,13 @@ int sd_vga_image_decode(sd_rgba_image *dst, const sd_vga_image *src, const vga_p
...
-    row_pointers = omf_calloc(h, sizeof(png_bytep));
-    for(int y = 0; y < h; y++) {
-        row_pointers[y] = omf_calloc(1, png_get_rowbytes(png_ptr, info_ptr));
-    }
-
-    if(setjmp(png_jmpbuf(png_ptr))) {
-        ret = SD_FILE_INVALID_TYPE;
-        goto error_3;
+    if(sd_vga_image_create(img, 320, 200) != SD_SUCCESS) {
+        return SD_FAILURE;
```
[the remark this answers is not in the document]
katajakasa: I mentioned this before, but the error codes are essentially extra baggage at this point. Ideally we remove them and just abort() immediately.

src/utils/png_reader.h
katajakasa: We want to get rid of the sd_ prefix, since that is a remnant from old libsd days. Maybe read_png_ and write_png_ would be acceptable ?

CMakeLists.txt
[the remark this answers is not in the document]
katajakasa: After cmake is configured once, this is unlikely to change, so there is no reason cache-wise not to have it global. It's simpler this way. Are there some other pressing reasons to make it more complicated ?

katajakasa: I don't mind, that works too. I used USE_ prefix, since I think of it as a decision to use/not use libpng. FOUND has a slightly different connotation, but it works too.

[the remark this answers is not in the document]
katajakasa: find_package and add_library I will definitely fix.

CMakeLists.txt
```diff
@@ -247,11 +248,18 @@ target_link_libraries(openomf_core PRIVATE
...
-    openomf::png
     openomf::SDL2
     openomf::SDL2_mixer
     openomf::xmp
 )
+
+if (USE_LIBPNG)
+    target_link_libraries(openomf_core PUBLIC
+        openomf::png
+    )
```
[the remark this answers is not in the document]
katajakasa: Good point. Forgot to change this after copy-pasting.

src/utils/png_reader.h
katajakasa: We don't use these codes anywhere; hey were a leftover from when this stuff was in a separate library. TODO remove the old error code stuff and use abort() where necessary.

CMakeLists.txt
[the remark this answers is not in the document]
Nopey: the advice of "just don't use add_definitions" is easier to understand than the nuance justifying the "simpler" add_definitions code, and encourages future authors to avoid the function as well.

Nopey (approved):

PR #855, opened 2025-02-09
Logging improvements

katajakasa:
WIP; probably need to add logging threads so that output does not clog the application.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,11 +271,11 @@ if(WIN32 AND NOT MINGW)
     )
 endif()
 
-if (USE_COLORS)
+if(USE_COLORS)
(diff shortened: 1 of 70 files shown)


CMakeLists.txt
```diff
@@ -270,6 +271,12 @@ if(WIN32 AND NOT MINGW)
     )
 endif()
 
+if (USE_COLORS)
+    # If user wants terminal colors. This is mainly for log, but can be used elsewhere too.
+    add_definitions(-DUSE_COLORS)
+    message(STATUS "Enabled terminal colors")
+endif ()
```
mrannanj, line 278: The if and endif should be near attached to the parens for consistency with other stuff in this file

mrannanj (changes requested):

Vagabond (approved):

PR #866, opened 2025-02-11
CI: Cleanups and Release action

katajakasa:
Cleanups + add a manual release action

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -88,7 +88,6 @@ jobs:
         arch: ${{ matrix.config.arch }}
         build_languages: ${{ matrix.config.build_languages }}
 
-
   # Build ubuntu package, release artifact
   # -----------------------------------------------------------------------------------------------
   build_ubuntu:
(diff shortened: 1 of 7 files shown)


.github/workflows/compilation.yml
```diff
@@ -56,23 +56,8 @@ jobs:
...
-      run: wget -q "${{ env.OPENOMF_ASSETS_URL }}"
-
-    - name: Cache omf2097-assets.zip
-      if: steps.cache-assets.outputs.cache-hit != 'true'
-      uses: actions/cache/save@v4
-      with:
-        path: omf2097-assets.zip
-        key: omf2097-assets.zip
+    - name: Get OMF2097 assets
+      uses: ./.github/actions/assets
```
mrannanj, line 60: Is the `./` part in the beginning needed? Seems maybe it isn't? So I'd prefer just `.github/actions/assets` if it works.

Vagabond: I believe I had to do it when I did the version `uses`

PR #867, opened 2025-02-12
formats: Remove undefined behavior in sd_animation_load() coord unpacking

mrannanj:


Vagabond (dismissed):
These shifts seem pointless to me as well.

PR #872, opened 2025-02-14
Switch to a version API to access the version macro information

Vagabond:


katajakasa (approved):
+1

src/game/utils/version.h
```diff
@@ -0,0 +1,7 @@
+
```
mrannanj, line 1: include guards missing

src/game/utils/version.c
```diff
@@ -0,0 +1,31 @@
+#define STRINGIFY(x) #x
```
mrannanj, line 1: Missing include "version.h"

mrannanj (changes requested):

src/game/utils/version.c
Vagabond: I believe this is unnecessary?

mrannanj (approved):

PR #874, opened 2025-02-14
Completely disable pause menu in arena netplay

Vagabond:


src/game/scenes/arena.c
```diff
@@ -1118,6 +1118,10 @@ void arena_input_tick(scene *scene) {
         controller_free_chain(p2);
     }
 
+    if(is_netplay(scene->gs)) {
+        return;
```
katajakasa, line 1122: Maybe put a comment on what this does :/

Vagabond: Added.

PR #876, opened 2025-02-15
Make rollback perfectly consistent with local play & improve dedup

Vagabond:
When we were deduping inputs, we were not considering the direction the
HAR was facing. This led to a case where if you jumped over the opponent
and kept holding the forward and jump keys, once you landed the HAR, on
rollback, would start jumping the other way. This is because we were
encoding the direction into the input buffer, but not when deduping, so
because the direction of the HAR changed, we need to allow the new input
to come through.

Additionally, the dedup has been corrected so that if the last event the
player has not changed, and the HAR's direction has not changed, there
is no need to store or send any input events. This significantly reduces
the network events sent and thus the rollbacks that need to be done.

### DIFF
--- a/src/controller/net_controller.c
+++ b/src/controller/net_controller.c
@@ -375,21 +375,18 @@ int rewind_and_replay(wtf *data, game_state *gs_current) {
         assert(gs->int_tick - data->local_proposal == ev->tick);
 
         // feed in the inputs
-        int arena_state = arena_get_state(game_state_get_scene(gs));
-        if(arena_state == ARENA_STATE_FIGHTING) {
-            for(int j = 0; j < 2; j++) {
-                int player_id = j;
-                game_player *player = game_state_get_player(gs, player_id);
-                int k = 0;
-                do {
-                    object_act(game_state_find_object(gs, game_player_get_har_obj_id(player)), ev->events[j][k]);
-                    k++;
-                } while(ev->events[j][k]);
+        for(int j = 0; j < 2; j++) {
+            int player_id = j;
+            game_player *player = game_state_get_player(gs, player_id);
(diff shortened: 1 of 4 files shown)


katajakasa:
Formatting.

src/controller/net_controller.c
```diff
@@ -356,21 +375,18 @@ int rewind_and_replay(wtf *data, game_state *gs_current) {
...
+            game_player *player = game_state_get_player(gs, player_id);
+            int k = 0;
+            do {
+                object_act(game_state_find_object(gs, game_player_get_har_obj_id(player)), ev->events[j][k]);
+                k++;
+            } while(ev->events[j][k]);
+
+            // write_rec_move(gs->sc, player, ev->events[j]);
+            //} else {
+            //    object_act(game_state_find_object(gs, game_player_get_har_obj_id(player)), ACT_STOP);
```
katajakasa, line 389: maybe just remove the dead code

src/controller/net_controller.c
```diff
@@ -908,19 +926,21 @@ void controller_hook(controller *ctrl, int action) {
 
     game_player *player = game_state_get_player(ctrl->gs, data->id);
     object *har_obj = game_state_find_object(ctrl->gs, game_player_get_har_obj_id(player));
+    int direction = OBJECT_FACE_NONE;
     if(har_obj) {
         har *har = object_get_userdata(har_obj);
         // if(action == ACT_STOP && har->state == data->last_har_state) {
         //     return;
         // }
```
katajakasa, line 934: here too

katajakasa (approved):

katajakasa (approved):

PR #877, opened 2025-02-15
game: Change assert crash to warning when ignoring input

mrannanj:


src/game/objects/har.c
```diff
@@ -1581,7 +1581,7 @@ void add_input(char *buf, int act_type, int direction) {
             add_input_to_buffer(buf, '5');
             break;
         default:
-            assert(false);
+            log_warn("Ignored input: buf %s, act_type %d, direction %d", buf, act_type, direction);
```
Nopey, line 1584:
```suggestion
            log_warn("Ignored input: buf %s, act_type 0x%x, direction %d", buf, act_type, direction);
```
act_type is a bitmask of the ACT_*'s from controller.h, which are specified in hex.

mrannanj: okay, changed to that

PR #881, opened 2025-02-16
Add support for hyper mode and tournament enhancements

Vagabond:
Implement the animation string search strategy for hyper/normal/enhancement animation strings when loading the HAR.

Essentially, in normal 1/2 player mode, if extra string 0 exists and is not blank or '!', use it. In hyper mode if extra string 1 exists and is not blank, use it.

For tournament mode, if the user has an enhancement, check index 1+enhancement level, if that exists, use it, otherwise use the highest populated string in the extra strings table. If there's no enhancement, honor the hyper mode setting.

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -15,6 +15,7 @@
 #include "game/protos/intersect.h"
 #include "game/scenes/arena.h"
 #include "game/utils/serial.h"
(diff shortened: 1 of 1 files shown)


Vagabond:
Rebased and switched to using str_set @katajakasa

katajakasa (approved):

PR #882, opened 2025-02-16
game,formats: Move tournament description parsing to sd_tournament_lo…

mrannanj:
…ad() helper

### DIFF
--- a/src/formats/tournament.c
+++ b/src/formats/tournament.c
@@ -37,6 +37,9 @@ static void free_locales(sd_tournament_file *trn) {
             if(trn->locales[i]->description) {
                 omf_free(trn->locales[i]->description);
             }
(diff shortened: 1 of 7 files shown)


src/formats/tournament.c
```diff
@@ -70,6 +70,64 @@ int sd_tournament_set_pic_name(sd_tournament_file *trn, const char *pic_name) {
...
+        start = desc;
+    } else {
+        start++;
+    }
+    locale->desc_width = width;
+    locale->desc_center = center;
+    locale->desc_vmove = vmove;
+    locale->desc_size = size;
+    locale->desc_color = color;
+    locale->description = omf_strdup(start);
```
mrannanj, line 123: This seems to be a bad idea, better  keep the original string for trntool.

PR #884, opened 2025-02-16
Fix 2 movement bugs:

### LINKED ISSUE #814: Reduced walkspeed after bumping enemy
After walking into an opponent, ones HAR cannot walk backwards as fast as it normally can.

Attached video starts by showing that this does not happen in DOS, and then switches to OpenOMF to show the bug.

https://github.com/user-attachments/assets/2747c5ff-6923-4171-9034-42a4a2476fbb

Vagabond:
* Fixes #814 When walking away from a HAR, restore normal walking speed
* Pushback from a move decreases over time if you're touching the ground. This is commonly seen when you do like a roundhouse while the enemy is up against the wall, you'll be pushed back for the duration of the attack animation.

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -593,7 +593,21 @@ void har_move(object *obj) {
         } else if(h->state != STATE_WALKFROM && h->state != STATE_WALKTO) {
             // add some friction from the floor if we're not walking
             // This is important to dampen/eliminate the velocity added from pushing away from the other HAR
-            object_set_vel(obj, vec2f_create(vel.x / 2.0f, vel.y));
+            if(vel.x > 0.0f) {
+                if(vel.x - 0.2f > 0.0f) {
+                    vel.x = 0.0f;
+                } else {
+                    vel.x -= 0.2f;
+                }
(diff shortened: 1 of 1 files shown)


src/game/objects/har.c
```diff
@@ -590,6 +590,24 @@ void har_move(object *obj) {
                     har_finished(obj);
                 }
             }
+        } else if(h->state != STATE_WALKFROM && h->state != STATE_WALKTO) {
+            // add some friction from the floor if we're not walking
+            // This is important to dampen/eliminate the velocity added from pushing away from the other HAR
+            if(vel.x > 0.0f) {
+                if(vel.x - 0.2f > 0.0f) {
```
katajakasa, line 597: This could be an edge function btw ;)

katajakasa (approved):

src/game/objects/har.c
Vagabond: what is that?

katajakasa: Sorry, step function. Brain fart. E.g. https://thebookofshaders.com/glossary/?search=step

PR #890, opened 2025-02-17
Try to make fight rounds end smoother:

Vagabond:
* Count the round over as soon as a player has 0 health
* Defer the you lose/win animation until both hars are at rest

The goal here is to be more like the original, where you instantly lose input control (other than scrap/destruction) and there's not a weird gap before the round starts to end.

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -620,14 +620,14 @@ void har_move(object *obj) {
 
                 h->state = STATE_DEFEAT;
                 har_set_ani(obj, ANIM_DEFEAT, 0);
-                har_event_defeat(h, ctrl);
+                // har_event_defeat(h, ctrl);
             } else if(pos.y >= (ARENA_FLOOR - 5) && IS_ZERO(vel.x) && player_is_last_frame(obj)) {
                 if(h->state == STATE_FALLEN) {
                     if(h->health <= 0) {
                         // fallen, but done bouncing
                         h->state = STATE_DEFEAT;
                         har_set_ani(obj, ANIM_DEFEAT, 0);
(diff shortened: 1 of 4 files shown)


Nopey:
The winner no longer waits for their defeated opponent to settle on the ground.
In DOS, the winner is unable to do more moves after their enemy's health is removed, and does not win-pose until both HARs are settled on solid ground.

Vagabond:
I reworked this a bunch.

src/game/objects/har.c
```diff
@@ -1653,6 +1656,10 @@ af_move *match_move(object *obj, char *inputs) {
                     continue;
                 }
 
+                if(move->category != CAT_DESTRUCTION && h->state == STATE_SCRAP) {
```
Vagabond, line 1659: Unnecessary

Vagabond:
The way this works now is, once the defeated HAR has landed, and has no x velocity, the win/lose animation plays, once that animation is finished, and the victorious HAR has reached the last frame in their victory animation (or the animation is looping, for example with shredder or electra) then the tick countdown to the end of the round begins. It is still deferred if a HAR does a scrap/destruct. Some buggy scrap/destructs can still delay this (Shadow's destruct is one example), but those are arguably bugs we should fix.

Nopey:
Tested 248b5b50932860eaee2878b84b82fffffc2a1485 (with pr 902 locally reverted to avoid a round-end crash), behavior does not seem different from previous patchset (same PR).

https://github.com/user-attachments/assets/ee967365-7fdb-45b9-b692-ee1f170bd140

in his second victory, shirro starts win-posing a tick or two before angel has hit the wall.

PR #892, opened 2025-02-17
Fix corruption in palette_set_player_expanded_color

### LINKED ISSUE #891: North America Win Hangar has palette has a problem affecting the background
Buggy hangar view looks like this:
Image

Palette for this:
Image

However, it's not always red. The background has been purple as well, here's the palette for that:
Image

Nopey:
This bug was introduced by commit a1e6df845daaca348ae7e4e70986eeb52f847e29.
Fixes #891


mrannanj (approved):
Tested, works.

PR #896, opened 2025-02-19
Implement "walk to HAR" used in several scrap animations

Vagabond:
Many of the scrap animations have the winner walk over to where the loser is lying on the ground and pick them up, or ruin them where they lay, etc.

Shadow's scrap has the winner walk to the far arena wall and lean against it.

Both of these are now implemented.

Additionally several bugs with the ez-destruct cheat were fixed, and going to non-scrap/destruction moves during scrap/destruction are now properly disallowed.

### DIFF
--- a/src/game/protos/player.c
+++ b/src/game/protos/player.c
@@ -258,20 +258,14 @@ void player_run(object *obj) {
             int destination = 160;
             if(sd_script_isset(frame, "am") && sd_script_isset(frame, "e")) {
                 // destination is the enemy's position
-                destination = enemy->pos.x;
-                if(sd_script_isset(frame, "x")) {
-                    if(object_get_direction(obj) == OBJECT_FACE_RIGHT) {
-                        destination += sd_script_get(frame, "x");
-                    } else {
-                        destination -= sd_script_get(frame, "x");
-                    }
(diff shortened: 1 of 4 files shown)


src/game/protos/player.c
```diff
@@ -265,33 +254,40 @@ void player_run(object *obj) {
...
                 obj->animation_state.shadow_corner_hack = 1;
             } else {
-                log_error("unknown end position for BE tag");
+                destination = -1;
+            }
+            // clear this
+            trans_x = 0;
+            if(sd_script_get(frame, "bm") == 10 && destination > 0 && fabsf(obj->pos.x - destination) > 5.0) {
+                log_debug("HAR walk to %d from %d", destination, obj->pos.x);
+                har_walk_to(obj, destination);
```
katajakasa, line 287: object is not necessarily HAR. This seems like something that should be checked in har.c-

katajakasa (approved):

src/game/protos/player.c
Vagabond: As discussed, we will refactor this once we have all the required tags implemented so we have a full understanding of the problem space.

PR #898, opened 2025-02-19
New text renderer

katajakasa:
New text renderer. Note that not everything is switched over yet, and there is likely much more work to be done to support all use cases... But it does work now, and I think we should merge it to give us something to build on.

Feel free to review, if you have the time and patience :D


Nopey:
<details>
<summary>Two pieces of text have regressed (I compared against the 0.7.2 pre release commit gef585de5). click to expand.</summary>

image
image

</details>

testing/test_text_layout.c
```diff
@@ -0,0 +1,116 @@
...
+    // Characters up to and including ' ' and 'A' are 8x8
+    surface s;
+    for(int i = 0; i <= 33; i++) {
+        surface_create(&s, 8, 8);
+        vector_append(&font->surfaces, &s);
+    }
+
+    // 'B' is 4x8
+    surface_create(&s, 4, 4);
+    vector_append(&font->surfaces, &s);
```
Nopey, line 23: Please use an active voice in these comments, eg:
```
// Make 'B' 4x8
```

Also: is B 4x8 or 4x4? If the former, please comment as how a 4x4 surface in a 8x8 font makes 'B' an 4x8 character.

src/game/gui/text/enums.h
```diff
@@ -0,0 +1,43 @@
...
+    ALIGN_TEXT_TOP,
+    ALIGN_TEXT_MIDDLE,
+    ALIGN_TEXT_BOTTOM
+} text_vertical_align;
+
+typedef enum text_horizontal_align
+{
+    ALIGN_TEXT_LEFT,
+    ALIGN_TEXT_CENTER,
+    ALIGN_TEXT_RIGHT
```
Nopey, line 17:
```suggestion
    TEXT_ALIGN_LEFT,
    TEXT_ALIGN_CENTER,
    TEXT_ALIGN_RIGHT
```
& same for text_vertical_align.

testing/test_text_layout.c
```diff
@@ -0,0 +1,116 @@
...
+#include "resources/fonts.h"
+#include "utils/allocator.h"
+#include "utils/str.h"
+#include "video/surface.h"
+#include <CUnit/CUnit.h>
+
+void create_fake_font(font *font) {
+    font_create(font);
+    font->w = 8;
+    font->h = 8;
```
Nopey, line 11: Could you comment on what distinguishes the font width/height from the character surface width/height, or what they do?

testing/test_text_layout.c
```diff
@@ -0,0 +1,116 @@
...
+    font->size = FONT_BIG;
+
+    // Characters up to and including ' ' and 'A' are 8x8
+    surface s;
+    for(int i = 0; i <= 33; i++) {
```
Nopey, line 16: This magic number 33 could use some explanation.
```c
// Create 8x8 surfaces for characters from ' ' (32) to 'A' (65), inclusive.
// We don't need to set surfaces for characters '\0' (0) through '\US' (31) because
// they are control characters, which the text renderer does not display.
for(int i = 0; i <= 'A' - ' '; i++) {
```

Nopey (changes requested):
Changes look good, please address review comments and fix the two dialog regressions.

For the "Accept Resolution?" and "Do you wish to continue?" dialogs see [this comment](https://github.com/omf2097/openomf/pull/898#issuecomment-2745961506) in conversation for screenshots.

testing/test_text_layout.c
katajakasa: Old width and height are required only for the old renderer now, and are there for compatibility. I added comments to the font struct.

src/game/gui/text/enums.h
katajakasa: Probably sounds nicer. Done.

testing/test_text_layout.c
katajakasa: Fonts no longer have sizes, glyphs do. The font size is only there for compatibility for now... Maybe I will add a comment on that in the test.

katajakasa: Btw, what is this active voice thing ? Does this have something to do with english language or tests conventions or what ?

Nopey: the `'B' is 4x8` comment states it as a fact, but not *why* or *when* it is true.

Most of the time, 'B' is not 4x8. It is, to my knowledge, only 4x8 in this test case.

By using a passive voice, the "who" of an action is not present-- here, the "who" is create_fake_font.
It's unfortunate that my suggestion was not in active voice, but instead imperative voice.. yeah.

some examples:
> The linens are ironed. (passive)
> Mr. B ironed the linens. (active)
> Iron the linens. (imperative)

katajakasa: Aaha. I usually consider the "who" to be the test, as I write the comments in the test context (test file). But perhaps that's just in my head then. In any case, it's changed.

Nopey (approved):

PR #901, opened 2025-02-21
Netplay Connection fixes

Nopey:
First commit makes our upnp work on my netgear router (which gives an err 725), second commit enables the miniupnpc and natpmp libraries by default in all builds, third commit adds the ability to enter a port on the connect screen.

I don't want to add a port setting to the "Start Server" dialog, because it mostly ignores it and likes to try random ports until it finds an available one. Players are expected to communicate their IP and port from server player to client player.

### DIFF
--- a/src/game/scenes/mainmenu/menu_connect.c
+++ b/src/game/scenes/mainmenu/menu_connect.c
@@ -25,7 +25,7 @@ typedef struct {
 
 enum
 {
-    lowest_port = 1,
+    lowest_port = 1025,
     highest_port = 65535,
     highest_port_digits = 5, // 65535 is 5 digits long
 };
--- a/src/game/scenes/mainmenu/menu_connect.c
+++ b/src/game/scenes/mainmenu/menu_connect.c
@@ -17,10 +17,25 @@ typedef struct {
     int controllers_created;
     ENetHost *host;
     component *addr_input;
+    component *port_input;
     component *connect_button;
     component *cancel_button;
     scene *s;
 } connect_menu_data;
+
+enum
+{
+    lowest_port = 1,
+    highest_port = 65535,
+    highest_port_digits = 5, // 65535 is 5 digits long
+};
+
+static void menu_connect_reset_port_input(connect_menu_data *local) {
+    char port_s[highest_port_digits + 1];
+    snprintf(port_s, sizeof(port_s), "%d", settings_get()->net.net_connect_port);
+    port_s[highest_port_digits] = '\0';
+    textinput_set_text(local->port_input, port_s);
+}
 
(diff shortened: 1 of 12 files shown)


src/game/scenes/mainmenu/menu_connect.c
```diff
@@ -17,11 +17,26 @@ typedef struct {
...
     component *addr_input;
+    component *port_input;
     component *connect_button;
     component *cancel_button;
     scene *s;
 } connect_menu_data;
 
+enum
+{
+    lowest_port = 1,
```
katajakasa, line 28: we should probably start from 1025, since the first 1024 ports are still considered "special".

src/game/utils/nat.c
```diff
@@ -14,31 +14,43 @@ bool nat_create_upnp_mapping(nat_ctx *ctx, uint16_t int_port, uint16_t ext_port)
...
+
+        if(error == 0) {
+            log_debug("NAT-uPNP Port map successfully created from %d to %d!", int_port, ext_port);
+
+            ctx->int_port = int_port;
+            ctx->ext_port = ext_port;
+            return true;
+        } else {
+            log_debug("NAT-uPNP port %d -> %d mapping failed with %d", int_port, ext_port, error);
+            if(error == 725 /* OnlyPermanentLeasesSupported */ && !use_permanent_lease) {
```
Vagabond, line 45:
``` * 715 WildCardNotPermittedInSrcIP | The source IP address cannot be wild-carded
 * 716 WildCardNotPermittedInExtPort | The external port cannot be wild-carded
 * 718 ConflictInMappingEntry | The port mapping entry specified conflicts with a mapping assigned previously to another client
 * 724 SamePortValuesRequired | Internal and External port values must be the same
 * 725 OnlyPermanentLeasesSupported | The NAT implementation only supports permanent lease times on port mappings
 * 726 RemoteHostOnlySupportsWildcard | RemoteHost must be a wildcard and cannot be a specific IP address or DNS name
 * 727 ExternalPortOnlySupportsWildcard | ExternalPort must be a wildcard and cannot be a specific port value
 * 728 NoPortMapsAvailable | There are not enough free ports available to complete port mapping.
 * 729 ConflictWithOtherMechanisms | Attempted port mapping is not allowed due to conflict with other mechanisms.
 * 732 WildCardNotPermittedInIntPort | The internal port cannot be wild-carded``` Are there others we can support as well?

Vagabond: We could retry the loop on 718, and we could also switch to wildcarding the external port on 727.

724 would be handleable with some work to signal the current internal port needs to be changed.

Vagabond (approved):

PR #902, opened 2025-02-21
game: Calculate tournament winnings closer like the original does

mrannanj:


src/game/scenes/arena.c
```diff
@@ -571,7 +571,9 @@ void arena_har_defeat_hook(int player_id, scene *scene) {
             player_winner->pilot->wins++;
             player_loser->pilot->losses++;
             if(player_id == 1) {
-                fight_stats->winnings = player_loser->pilot->winnings;
+                float winnings_multiplier = game_state_get_player(scene->gs, 0)->chr->winnings_multiplier;
+                fight_stats->winnings =
+                    (player_loser->pilot->money + player_loser->pilot->winnings) * winnings_multiplier;
```
Nopey, line 576: This crashes when `chr` is NULL, such as in 1 player mode.

PR #906, opened 2025-02-22
game: Clean up arena_har_defeat_hook()

mrannanj:


src/game/scenes/arena.c
```diff
@@ -603,7 +603,7 @@ void arena_har_defeat_hook(int player_id, scene *scene) {
         har_set_ani(winner, ANIM_VICTORY, 0);
         winner_har->state = STATE_VICTORY;
         local->over = 1;
-        local->winner = player_id;
+        local->winner = winner_player_id;
```
mrannanj, line 606: @Vagabond This was previously setting `local->winner` as the loser player, I wonder what breaks when this is changed ..

PR #908, opened 2025-02-22
game: Only fill fight stats after arena for tournament mode

mrannanj:
This fixes arena end hook crashing in one player mode.


Vagabond:
I don't think this always ensure you win/lose will be displayed now

PR #912, opened 2025-02-22
Fix shadow grab by implementing finish callback for projectiles

Vagabond:


Vagabond:
I think this was broken by https://github.com/omf2097/openomf/commit/b077a18f198eacfe3b7febc02cfe42413ab39291

PR #913, opened 2025-02-22
0 damage moves should do 0 damage

Vagabond:


mrannanj (approved):
Tested, seems to work.

PR #921, opened 2025-02-24
Avoid AI stalls and bad round ends

Vagabond:
This sort of replaces #890 and #920.


Vagabond:
I've run demo mode over 33,000 times with this and no crashes/asserts, AI stalls or ending stalls now. I think we should merge this.

PR #922, opened 2025-02-25
Allow progressbars to animate when they're lowering

Vagabond:
This makes the health/endurance meters more like the original.

### DIFF
--- a/src/game/gui/progressbar.c
+++ b/src/game/gui/progressbar.c
@@ -44,6 +44,7 @@ typedef struct {
     surface *block;
     int orientation;
     int percentage;
+    int display_percentage;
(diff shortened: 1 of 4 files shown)


src/game/gui/progressbar.c
```diff
@@ -52,12 +53,16 @@ typedef struct {
     int refresh;
 } progressbar;
 
-void progressbar_set_progress(component *c, int percentage) {
+void progressbar_set_progress(component *c, int percentage, bool animate) {
```
katajakasa, line 56: Code docs on what the arguments do ? Took a while to read through it, so quick comments might help here.

PR #929, opened 2025-02-28
Store gameplay affecting settings in a game_state struct

Vagabond:
This struct avoids direct access to the global settings. This is because in some cases (tournament mode, netplay, REC playback) we don't want to use the global settings blindly. For example some options are ignored in netplay mode (eg player 1/2 power), agreeing on the options for netplay should be handled differently, and REC files contain their own set of all the gameplay settings.

### DIFF
--- a/src/game/game_state_type.h
+++ b/src/game/game_state_type.h
@@ -37,7 +37,7 @@ typedef struct ticktimer_t ticktimer;
 typedef struct controller_t controller;
 
 // roughly modeled after the configuration in REC files
-typedef struct match_settings_t {
+typedef struct {
     uint8_t throw_range;
     uint8_t hit_pause;
     uint8_t block_damage;
--- a/src/game/game_state_type.h
+++ b/src/game/game_state_type.h
@@ -36,14 +36,6 @@ typedef struct game_player_t game_player;
 typedef struct ticktimer_t ticktimer;
 typedef struct controller_t controller;
 
-typedef enum
-{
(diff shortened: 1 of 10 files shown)


src/game/game_state_type.h
```diff
@@ -35,6 +36,23 @@ typedef struct game_player_t game_player;
 typedef struct ticktimer_t ticktimer;
 typedef struct controller_t controller;
 
+// roughly modeled after the configuration in REC files
+typedef struct match_settings_t {
```
katajakasa, line 40: The _t prefix is technically reserved for stdlib; might be better to just have matching name for struct and typedef.

src/game/scenes/mechlab.c
```diff
@@ -485,6 +485,13 @@ int mechlab_create(scene *scene) {
     mechlab_local *local = omf_calloc(1, sizeof(mechlab_local));
     local->selling = false;
 
+    // reset the match settings and override the settings that don't apply in tournament mode
+    game_state_match_settings_reset(scene->gs);
+    scene->gs->match_settings.power1 = 5;
+    scene->gs->match_settings.power2 = 5;
+    scene->gs->match_settings.vitality = 100;
+    // TODO what others are there?
```
katajakasa, line 493: Todo before or after merge ? :P

Vagabond: I believe the only one missing is # of rounds. Will fix that now.

PR #937, opened 2025-03-02
Improve Wall Splat

LegendaryBlueShirt:
Just a quick change as my first PR, replaced some old handling with cleaner code and copied the animation string exactly from MASTER.DAT


Vagabond (approved):
Tested working.

PR #942, opened 2025-03-02
No longer treat walking as velocity, add walk speed to pos each frame

Vagabond:


katajakasa:
Why is this needed ?

Vagabond:
Treating walking as velocity vs a frame by frame position adjustment means that other sources of velocity don't interact weirdly with the HAR. cc @LegendaryBlueShirt

Vagabond:
> > Treating walking as velocity vs a frame by frame position adjustment means that other sources of velocity don't interact weirdly with the HAR. cc @LegendaryBlueShirt
> 
> Should jump also be separated in that case ?

I don't think so, because jump is subject to gravity. Walking is not subject to friction or anything (other than when you walk into the other HAR and slow down).

LegendaryBlueShirt:
As mentioned, walking is only a special case due to physics interactions, the biggest one that comes to mind is being able to negate pushback from hits if walking overwrote velocity, which is not what we observe in the source game.  Jumping is fine as it is.

katajakasa (approved):

PR #943, opened 2025-03-02
Make throws have priority

Vagabond:
If p1 threw p2, ignore what p2 is doing. If p2 threw p1, AND p1 didn't also do a throw, ignore p1's attack.

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -1109,6 +1109,8 @@ void har_debug(object *obj) {
 }
 #endif // DEBUGMODE
 
+// function to check if har A is hitting har B. Returns 1 if the har is executing a priority move which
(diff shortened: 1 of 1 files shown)


src/game/objects/har.c
```diff
@@ -1109,7 +1109,7 @@ void har_debug(object *obj) {
 }
 #endif // DEBUGMODE
 
-void har_collide_with_har(object *obj_a, object *obj_b, int loop) {
+int har_collide_with_har(object *obj_a, object *obj_b, int loop) {
```
katajakasa, line 1112: function docs on what is returned. Or enum ? Something to document the output.

katajakasa (approved):

PR #945, opened 2025-03-03
Attempt to make input buffering the same as the original

Vagabond:
There are a few tests here that confirmed our input buffering was wrong:

* You could chain from Jaguar 1k into k without releasing directional keys
* Electra allowed 65K6P in the original but not in openomf

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -1665,7 +1665,7 @@ af_move *match_move(object *obj, char prefix, char *inputs) {
     for(int i = 0; i < 70; i++) {
         if((move = af_get_move(h->af_data, i))) {
             len = str_size(&move->move_string);
-            if(str_c(&move->move_string)[0] == prefix &&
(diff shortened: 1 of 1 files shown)


src/game/objects/har.c
```diff
@@ -1653,27 +1653,26 @@ void add_input(char *buf, int act_type, int direction) {
...
-af_move *match_move(object *obj, char *inputs) {
+af_move *match_move(object *obj, char prefix, char *inputs) {
     har *h = object_get_userdata(obj);
     af_move *move = NULL;
     size_t len;
     for(int i = 0; i < 70; i++) {
         if((move = af_get_move(h->af_data, i))) {
             len = str_size(&move->move_string);
-            if(!strncmp(str_c(&move->move_string), inputs, len)) {
+            if(str_c(&move->move_string)[0] == prefix &&
```
katajakasa, line 1668: str_at(&move->move_string, 0) == prefix

katajakasa (approved):

PR #949, opened 2025-03-04
Add the ability to annotate REC files and to use them as test cases

Vagabond:
This branch allows assertions about the game state to be inserted into
REC files using the following syntax:

./rectool -f ../rectests/6P.REC --assert=29 --assertion_tick=312 --operator=eq --operand1=har2 --operand1_value=anim --operand2=literal --operand2_value=41 -o ../rectests/6P.REC

In this example we are asserting that at tick 312 har 2 is in animation
41. Rectest can also display assertions.

This branch also adds support to the REC controller to check the
assertions, and abort() if they are not met. Finally it adds a runner
for some included annotated RECs called run_rectest.sh, that can be run
from CI.


src/formats/rec.c
```diff
@@ -334,9 +334,13 @@ void sd_rec_finish(sd_rec_file *rec, unsigned int ticks) {
...
     memset(&move, 0, sizeof(move));
     move.tick = ticks;
-    move.lookup_id = 2;
+    move.lookup_id = 10;
     move.player_id = 0;
     move.action = 0;
+    move.raw_action = 0x23;
+
+    move.extra_data = omf_calloc(7, 1);
+    move.extra_data[0] = 0x78;
```
Vagabond, line 343: This is leftover test code, should be removed

PR #950, opened 2025-03-05
GUI improvements

katajakasa:
Minor things here and there, some struct hiding etc.


Nopey (approved):
Looks good. Please squash when merging.

PR #953, opened 2025-03-05
Improve REC printing, especially for type 18 actions

Vagabond:


src/formats/rec.c
```diff
@@ -41,6 +41,12 @@ void sd_rec_free(sd_rec_file *rec) {
         }
         omf_free(rec->moves);
     }
+
+    for(int i = 0; i < 2; i++) {
+        if(rec->pilots[i].has_photo) {
+            sd_sprite_free(&rec->pilots[i].photo);
```
katajakasa, line 47: sd_pilot_free() ?

Vagabond: but we don't allocate a pilot... we allocate a sprite on line 80.

Vagabond: oh, I guess we do, idk, this fixes the leak??

katajakasa: sd_pilot_free also does omf_free() for the photo, should that also be done here ?

Vagabond: the photo isn't even in the sd_pilot struct though

katajakasa (approved):

PR #954, opened 2025-03-05
Implement controllable projectiles via cx/cy tags

Vagabond:


LegendaryBlueShirt (approved):
Surface level, lgtm.  Will be testing it out later to see if anything needs to be tweaked.

src/game/objects/har.c
```diff
@@ -649,6 +649,35 @@ void har_move(object *obj) {
         } else if(h->state == STATE_WALKFROM) {
             obj->pos.x -= (h->back_speed * object_get_direction(obj)) * (h->hard_close ? 0.5 : 1.0);
         }
+
+        // TODO unify this code with the (very similar) code in har_move
+        // but we currently don't have a way in player.c to know if an object is a HAR or a projectile
+        if(player_frame_isset(obj, "cx")) {
+            float cx = player_frame_get(obj, "cx") / 10.0 * obj->horizontal_velocity_modifier;
+            log_debug("CX! %f, %c", cx, h->inputs[0]);
+            if(h->inputs[0] == '4') {
```
katajakasa, line 658: Switch-case in a shared function between har and projectile ?

katajakasa: Possibly a helper function in object.c ?

LegendaryBlueShirt:
Control feels very good on initial testing, the only problem I see is actually the initial velocity of concussion cannon

LegendaryBlueShirt (dismissed):

src/game/objects/har.c
Vagabond: Refactored into a common function as requested.

PR #955, opened 2025-03-05
Fix Q tag implementation

Vagabond:
The Q tag should allow subsequent hits if the # of hits this move has made already is less than the value of Q. For example a move like A1-q2B1-q2C1 should only hit twice, the second hit coming either on the B or C frame, but A1-q2B1-q3C1 can hit 3 times.

### DIFF
--- a/src/game/protos/object.c
+++ b/src/game/protos/object.c
@@ -74,8 +74,8 @@ void object_create(object *obj, game_state *gs, vec2i pos, vec2f vel) {
 
     random_seed(&obj->rand_state, rand_intmax());
 
-    // For enabling hit on the current and the next n-1 frames
-    obj->hit_frames = 0;
+    // For enabling multiple hits per move
+    obj->q_counter = 0;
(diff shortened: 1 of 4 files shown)


src/game/protos/object.c
```diff
@@ -74,8 +74,8 @@ void object_create(object *obj, game_state *gs, vec2i pos, vec2f vel) {
 
     random_seed(&obj->rand_state, rand_intmax());
 
-    // For enabling hit on the current and the next n-1 frames
-    obj->hit_frames = 0;
+    // For enabling multiple hits per move
+    obj->q_counter = 0;
```
katajakasa, line 78: q_hit_counter for better variable documentation ? :)

LegendaryBlueShirt (approved):
Tested working now!

src/game/protos/object.c
Vagabond: turns out it's not actually a hit counter, so leaving it as is.

PR #956, opened 2025-03-05
Fix a few issues with REC playback:

Vagabond:
* Use the pilot struct directly from the REC
* Always use the pilot name field; ensure it it set in 1/2 player

### DIFF
--- a/src/game/scenes/melee.c
+++ b/src/game/scenes/melee.c
@@ -16,6 +16,7 @@
 #include "resources/pilots.h"
 #include "resources/sprite.h"
 #include "utils/allocator.h"
+#include "utils/c_string_util.h"
 #include "utils/log.h"
 #include "utils/random.h"
 #include "video/vga_state.h"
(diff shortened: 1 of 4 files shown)


src/game/scenes/melee.c
```diff
@@ -399,6 +400,19 @@ void handle_action(scene *scene, int player, int action) {
...
+                    // 1player name strings end in a newline...
+                    if(player1->pilot->name[strlen(player1->pilot->name) - 1] == '\n') {
+                        player1->pilot->name[strlen(player1->pilot->name) - 1] = 0;
+                    }
+                    strncpy_or_truncate(player2->pilot->name, lang_get(player2->pilot->pilot_id + 20),
+                                        sizeof(player1->pilot->name));
+                    // 1player name strings end in a newline...
+                    if(player1->pilot->name[strlen(player1->pilot->name) - 1] == '\n') {
+                        player1->pilot->name[strlen(player1->pilot->name) - 1] = 0;
+                    }
```
mrannanj, line 415: Feels this newline stripping here should be done elsewhere, but also player1 is done twice here.

Nopey: the lang strings don't have to end in a newline, but whatever TXT file the originals were compiled from all do.
i'd propose we keep the newline stripping, and add a `// TODO: lang: remove (the need for) newline stripping` comment or similar.

mrannanj: Okay, @Nopey fixes the newline in another pull request, that's fine, but this code still is doing the operation twice to player1, and using `sizeof(player1->pilot->name)` to get the size of `player2->pilot->name`

Vagabond: will fix

src/game/scenes/melee.c
```diff
@@ -403,15 +403,17 @@ void handle_action(scene *scene, int player, int action) {
                     // TODO in netplay, use the lobby names
                     strncpy_or_truncate(player1->pilot->name, lang_get(player1->pilot->pilot_id + 20),
                                         sizeof(player1->pilot->name));
+                    // TODO: lang: remove (the need for) newline stripping
                     // 1player name strings end in a newline...
                     if(player1->pilot->name[strlen(player1->pilot->name) - 1] == '\n') {
                         player1->pilot->name[strlen(player1->pilot->name) - 1] = 0;
                     }
                     strncpy_or_truncate(player2->pilot->name, lang_get(player2->pilot->pilot_id + 20),
                                         sizeof(player1->pilot->name));
```
mrannanj, line 412: This line here still using player1 for the sizeof, although they would probably be the same size, it is confusing.

Vagabond: Fixed

katajakasa (approved):

PR #962, opened 2025-03-06
Fix kreissak custom defeat again, and add a rectest

Vagabond:


Vagabond:
I need to re-record the rectest for this one.

katajakasa (approved):

LegendaryBlueShirt:
Needs a check for whether or not it's the final round of the match I think

Vagabond:
Ok, this has been rebased to use #991 so the rectest can be much simpler and reliable. I additionally made it a 2 round rectest and assert on the first round that the custom defeat animation is not used.

katajakasa (approved):

PR #963, opened 2025-03-06
Restore lost hit-pause PR

Vagabond:
Restore code lost from #934 somehow


katajakasa (approved):
I think this already got approved earlier ?

Vagabond:
Yes, but got lost somehow

katajakasa:
> Yes, but got lost somehow

Just Ship It.

PR #968, opened 2025-03-07
Delay the end of round, and the assertion check, if scrap walking

Vagabond:


Vagabond:
I fixed the assert for cases where the winner doesn't end in the VICTORY animation, for example shadow scrap or destruct, and I fixed a problem where the horizontal velocity modifiers were causing the x offset from the wall to get scaled wrong.

katajakasa (approved):
Didn't immediately get it to misbehave, so LGTM.

PR #970, opened 2025-03-07
Restore BJ tag to fix infinite loop in gargoyle scrap

Vagabond:


Vagabond:
This was disabled before because of a crash, which is fixed by adding the return

katajakasa (approved):

PR #972, opened 2025-03-09
Implement reverse engineered air knockback formulae

Vagabond:


LegendaryBlueShirt (approved):
Checking the air hit formula it looks like there's a missing operand in the equation, 0.00390625 (1/256)
However without validating in cheatengine it feels fine as it is.

Vagabond:
I believe that's floating point conversion stuff

src/game/objects/har.c
```diff
@@ -758,11 +758,21 @@ void har_take_damage(object *obj, const str *string, float damage, float stun) {
 
             str n;
             str_from_slice(&n, string, 0, last_line);
-            str_append_c(&n, "-L2-M5-L2");
+            if(h->endurance <= 0) {
+                // this hit stunned them, so make them hit the floor stunned
+                str_append_c(&n, "L2-M5000");
```
LegendaryBlueShirt, line 763: Just noticed in ghidra the string is ever so slightly off
```suggestion
                str_append_c(&n, "L3-M5000");
```

Vagabond: Yes, I probably typoed

PR #976, opened 2025-03-09
Allow easily changing source samplerate for sounds

katajakasa:


LegendaryBlueShirt (dismissed):
Well the pitches are still wrong but this PR isn't meant to address that yet so seems fine in my book.

katajakasa:
This IS actually trying to deal with that. What I need here is someone who can tell me what direction they are off in, and how much I need to tune them to make them right.

Vagabond:
They're too low, by 4 semitones :P

katajakasa:
Feel free to tweak them in this PR. Simply change the "pitch = 1.0f + clamp(sf, -128, 128) / 128.0f;" line as required.

Vagabond:
I reverse engineered the pitches, using Crystal in a Jaguar. I didn't notice any pitch differences if I used Shirro, so it's possible it's agility independent.

I also added panning to sounds that didn't have a panning tag, and I fixed the landing sound panning.

PR #983, opened 2025-03-11
Fix key repeat behaviour in REC playback and add rectest

Vagabond:
Make the REC controller emulate the keyboard more accurately by repeating the last input each time it is polled. This fixes a case where you do jaguar's standing throw and keep holding the directional key, after the throw the player should continue in the same direction as before the throw, not change direction and follow the thrown HAR.


LegendaryBlueShirt (changes requested):
There's a line in keyboard.c that makes the dedupe work correctly
`ctrl->last = ctrl->current;`

Add that to the end of controller_cmd and you shouldn't have to have punch/kick dedupe logic in rec_controller anymore

Vagabond:
maybe @Nopey can chime in here

LegendaryBlueShirt (approved):

Nopey (approved):

PR #986, opened 2025-03-11
Fix projectiles hitting wall

Vagabond:
Only try to keep HARs onscreen, let projectiles fly free


LegendaryBlueShirt:
Testing with concussion cannon, doesn't seem to be changing anything and the blast hits the wall immediately

Vagabond:
try it with nova

LegendaryBlueShirt:
Nova's projectiles hit the wall, as expected.

Vagabond:
right, but they didn't before, right?

katajakasa (approved):

PR #989, opened 2025-03-12
Better Falling Behavior + Misc Nits

LegendaryBlueShirt:
Falling now behaves much more like the original game

Removed a bunch of places where horizontal velocity is cancelled that are no longer needed

Moving towards the wall no longer immediately cancels your velocity

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -589,11 +589,11 @@ void har_move(object *obj) {
                     obj->vel.x = obj->vel.x / 2;
                     if(h->id != 10) {
                         object_set_custom_string(obj, "l20s4sp13zzN3-zzM100");
-                        obj->gs->screen_shake_vertical = 5; //Multiplied by 5 to make it visible
+                        obj->gs->screen_shake_vertical = 5; // Multiplied by 5 to make it visible
                     } else {
-                        //Nova falls harder
+                        // Nova falls harder
                         object_set_custom_string(obj, "l40s4sp13zzN3-zzM100");
-                        obj->gs->screen_shake_vertical = 15; //Multiplied by 5 to make it visible
+                        obj->gs->screen_shake_vertical = 15; // Multiplied by 5 to make it visible
                     }
                 } else {
                     obj->vel.y = 0;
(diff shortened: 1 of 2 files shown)


src/game/objects/har.c
```diff
@@ -2022,12 +2036,8 @@ int har_act(object *obj, int act_type) {
...
 
     if(move) {
-        // Stop horizontal movement, when move is done
-        // TODO: Make this work better
         vec2f spd = object_get_vel(obj);
-        if(h->state != STATE_JUMPING) {
-            spd.x = 0.0f;
-        }
+
         object_set_vel(obj, spd);
```
Vagabond, line 2041:
```suggestion
```
Remove this too?

src/game/objects/har.c
```diff
@@ -765,8 +785,6 @@ void har_take_damage(object *obj, const str *string, float damage, float stun) {
             obj->vel.y = -7 * object_get_direction(obj);
             h->state = STATE_FALLEN;
             object_set_stride(obj, 1);
-        } else {
-            object_set_custom_string(obj, str_c(string));
```
Vagabond, line 769: This seems wrong, why would you not use the 'hit' animation string in this case?

Vagabond: oh you do it above, nm

src/game/objects/har.c
```diff
@@ -584,13 +582,30 @@ void har_move(object *obj) {
                 har_face_enemy(obj, obj_enemy);
             }
         } else if(h->state == STATE_FALLEN || h->state == STATE_RECOIL) {
-            if(obj->pos.y > ARENA_FLOOR) {
-                obj->pos.y = ARENA_FLOOR;
-                har_floor_landing_effects(obj);
+            if(obj->vel.y > 0) {
+                if(obj->vel.y > 6) {
```
Vagabond, line 586: So falling faster than 6 causes a bounce, otherwise it gets quiesced?

Vagabond, line 586:
```suggestion
                // bounce and screenshake if falling fast enough
                if(obj->vel.y > 6) {
```

Vagabond (dismissed):

src/game/objects/har.c
LegendaryBlueShirt: Indeed

PR #999, opened 2025-03-17
Mark projectile as having hit if it's blocked

Vagabond:


LegendaryBlueShirt (changes requested):
Projectiles should trigger their disappearing animation on contact, right now Nova's rocket among others pass through you

Vagabond:
> Projectiles should trigger their disappearing animation on contact, right now Nova's rocket among others pass through you

How about now?

LegendaryBlueShirt (approved):
The issue has been resolved, I'll create a follow-up issue for the lack of push on block

PR #1002, opened 2025-03-17
Fix some off by ones in SIM menu

Vagabond:


LegendaryBlueShirt (approved):
For anyone wondering, the last ranked challenger is missing from the sim mode selector without this PR.

PR #1009, opened 2025-03-18
AW and AB tag handling

LegendaryBlueShirt:
- ab tag now allows passing through walls
- aw tag now triggers animation changes for x position changes not just x velocity

Test cases

Chronos Matter Phasing (14K) should smoothly exit one wall and enter through the other.  For low agility pilots, they may not reach full screen.
Katana Wall Spin (14K) should trigger the followup animation correctly
Electra Rolling Thunder (656P) and Extended Rolling Thunder (3656P) should both ricochet off the wall


src/game/protos/object.h
```diff
@@ -81,6 +81,7 @@ struct object_t {
     float horizontal_velocity_modifier;
     int8_t direction;
     int8_t group;
+    bool wall_collision;
```
Nopey, line 84: Perhaps leave a comment to the effect that `wall_collision` tracks whether we *have* experienced a wall collision, as it does not control whether we *will* experience wall collision.

src/game/scenes/arena.c
```diff
@@ -1182,8 +1182,12 @@ void arena_dynamic_tick(scene *scene, int paused) {
...
 
         // check some invariants
-        assert(obj_har[0]->pos.x >= ARENA_LEFT_WALL && obj_har[0]->pos.x <= ARENA_RIGHT_WALL);
-        assert(obj_har[1]->pos.x >= ARENA_LEFT_WALL && obj_har[1]->pos.x <= ARENA_RIGHT_WALL);
+        if(!player_frame_isset(obj_har[0], "ab")) {
+            assert(obj_har[0]->pos.x >= ARENA_LEFT_WALL && obj_har[0]->pos.x <= ARENA_RIGHT_WALL);
+        }
+        if(!player_frame_isset(obj_har[1], "ab")) {
+            assert(obj_har[1]->pos.x >= ARENA_LEFT_WALL && obj_har[1]->pos.x <= ARENA_RIGHT_WALL);
+        }
```
Nopey, line 1190:
```suggestion
        assert(player_frame_isset(obj_har[0], "ab") ||
               (obj_har[0]->pos.x >= ARENA_LEFT_WALL && obj_har[0]->pos.x <= ARENA_RIGHT_WALL));
        assert(player_frame_isset(obj_har[1], "ab") ||
               (obj_har[1]->pos.x >= ARENA_LEFT_WALL && obj_har[1]->pos.x <= ARENA_RIGHT_WALL));
```
Keep the conditionals inside the `assert` so the `player_frame_isset` calls are not present in release builds.

Nopey (commented):
Changes look good in-game on Chronos, although I'm not nimble enough to see a difference with Katana or Electra's moves.

It would be nice to have rectests for these changes.

I've attached a recording where Jean-Paul spams matter phasing until Ib's Katana tears his Chronos in half (or very slightly desyncs, leaving Jean-Paul with not even a sliver of health). [20250324_115159.zip](https://github.com/user-attachments/files/19436366/20250324_115159.zip)

Vagabond:
Yes, some rectests (or just recs that can be annotated) would be good here.

Nopey:
Rebased, added three rectests, and squashed the assert change into the previous commit.

Nopey:
the no-idle-defeat rectest started desyncing, so I've made it more reliable by setting HAR1's health to 1 a few ticks before the final blow.

Nopey (dismissed):

src/game/protos/object.h
LegendaryBlueShirt: Err, it actually does mean the latter.  There are two seperate places where I must make a wall collision check and so I needed a way to propagate it.

Nopey: a good comment might help.
what i meant to say is: it tracks whether we are experiencing a wall collision, not whether we are capable of experiencing wall collisions.

Nopey (approved):

PR #1014, opened 2025-03-19
Object coordinates now have more precision

LegendaryBlueShirt:
fpos gives more precise coordinates while existing pos methods retain their original behavior

Methods and new equivalents
object_get_pos -> object_get_fpos
object_set_pos -> object_set_fpos
object_px -> object_fpx
object_py -> object_fpy
object_set_px -> object_set_fpx
object_set_py -> object_set_fpy

Planes in the desert are broken but everything else seems ok so far?
Most of the code actually references obj->pos directly without using one of these functions which was a bit annoying

Low agility pilots should appear to move faster now


src/game/scenes/arena.c
```diff
@@ -1520,8 +1520,8 @@ int arena_create(scene *scene) {
     // Initial har data
     vec2i pos[2];
     int dir[2] = {OBJECT_FACE_RIGHT, OBJECT_FACE_LEFT};
-    pos[0] = vec2i_create(HAR1_START_POS, ARENA_FLOOR);
-    pos[1] = vec2i_create(HAR2_START_POS, ARENA_FLOOR);
+    pos[0] = vec2i_create(HAR1_START_POS / 256, ARENA_FLOOR / 256);
```
Vagabond, line 1523: I feel like the 2i should hold the full pseudo float, and we have functions that  divide by 256 to get the pixel coordinate when we need it

Vagabond: I think we should do it all the right way and fix what breaks. I can assist if you want.

Vagabond:
How do we get this moving again?

src/game/protos/object.h
```diff
@@ -239,20 +239,26 @@ void object_set_pal_limit(object *obj, int limit);
...
 int object_py(const object *obj);
 float object_vx(const object *obj);
 float object_vy(const object *obj);
 
+void object_set_fpx(object *obj, int val);
 void object_set_px(object *obj, int val);
+void object_set_fpy(object *obj, int val);
 void object_set_py(object *obj, int val);
 void object_set_vx(object *obj, float val);
 void object_set_vy(object *obj, float val);
```
Nopey, line 264: please add comments to explain what distinguishes the many similarly named functions here.

Nopey:
Is it unreasonable to hope that all `float`s in desync-sensitive game logic are eventually replaced with fixed point math in one form or another?
Perhaps that is out of scope, but I believe it's relevant discussion here.

LegendaryBlueShirt:
I'm just busy with other things is all, haven't had the chance to make necessary changes to the spawn calls



> Is it unreasonable to hope that all `float`s in desync-sensitive game logic are eventually replaced with fixed point math in one form or another? Perhaps that is out of scope, but I believe it's relevant discussion here-- especially with this PR storing a fixed point decimal in a floating point field (object->pos is a vec2f).

Yes it absolutely should be a vec2i

Vagabond:
> I'm just busy with other things is all, haven't had the chance to make necessary changes to the spawn calls
> 
> > Is it unreasonable to hope that all `float`s in desync-sensitive game logic are eventually replaced with fixed point math in one form or another? Perhaps that is out of scope, but I believe it's relevant discussion here-- especially with this PR storing a fixed point decimal in a floating point field (object->pos is a vec2f).
> 
> Yes it absolutely should be a vec2i

I would definitely like to minimize our use of floating point math wherever possible. Luckily the way velocity is done should be relatively safe from the perspective of accumulating floating point error (although I do think we should *consider* switching that to a pseudo float as well).

katajakasa:
> > I'm just busy with other things is all, haven't had the chance to make necessary changes to the spawn calls
> > > Is it unreasonable to hope that all `float`s in desync-sensitive game logic are eventually replaced with fixed point math in one form or another? Perhaps that is out of scope, but I believe it's relevant discussion here-- especially with this PR storing a fixed point decimal in a floating point field (object->pos is a vec2f).
> > 
> > 
> > Yes it absolutely should be a vec2i
> 
> I would definitely like to minimize our use of floating point math wherever possible. Luckily the way velocity is done should be relatively safe from the perspective of accumulating floating point error (although I do think we should _consider_ switching that to a pseudo float as well).

Switching to integer math would certainly be better from portability side also.

src/game/objects/arena_constraints.h
```diff
@@ -1,8 +1,8 @@
 #ifndef ARENA_CONSTRAINTS_H
 #define ARENA_CONSTRAINTS_H
 
-#define ARENA_LEFT_WALL 20
-#define ARENA_RIGHT_WALL 300
-#define ARENA_FLOOR 190
+#define ARENA_LEFT_WALL 5120
+#define ARENA_RIGHT_WALL 76800
+#define ARENA_FLOOR 48640
```
katajakasa, line 6: Instead of rolling magic values, could these instead be the operation ? e.g. 

#define ARENA_LEFT_WALL (20 * 256)

Compiler will probably optimize that anyway, and it would be much clearer on what these values are.

Vagabond: That's why I suggested a macro so we could do something like `#define ARENA_FLOOR  PSEUDOFLOAT(190)`

src/game/objects/har.c
```diff
@@ -450,7 +451,7 @@ void har_floor_landing_effects(object *obj, bool play_sound) {
     int amount = rand_int(2) + 1;
     for(int i = 0; i < amount; i++) {
         int variance = rand_int(20) - 10;
-        vec2i coord = vec2i_create(obj->pos.x + variance + i * 10, obj->pos.y);
+        vec2i coord = vec2i_create(obj->pos.x / 256 + variance + i * 10, obj->pos.y / 256);
```
katajakasa, line 454: Would it be better to use bitshift for faster divide ? pos.x >> 8 ?

katajakasa: Though, compiler might already optimize this ? Maybe ?

Vagabond: being explicit would probably help portability on platforms with... more simplistic... C compilers.

LegendaryBlueShirt:
Closed in favor of https://github.com/omf2097/openomf/pull/1044

src/game/objects/har.c
turol: Any compiler released in the last 30 years should do this.

Nopey: I prefer division over bitshifts, because bitshifts are more difficult for me to use and understand:
- i have to remember my left from my right
- i have to remember not to shift negative numbers
- i have to remember another operator's precedence

PR #1017, opened 2025-03-19
Ensure we are loading pilot genders and 1p stats properly

Vagabond:


src/formats/chr.c
```diff
@@ -192,12 +192,13 @@ int sd_chr_load(sd_chr_file *chr, const char *filename) {
             chr->enemies[i]->pilot.total_value = trn.enemies[i]->total_value;
             chr->enemies[i]->pilot.photo_id = trn.enemies[i]->photo_id;
         }
+        if(trn_loaded) {
+            chr->enemies[i]->pilot.sex = pic.photos[trn.enemies[i]->photo_id]->sex;
+        }
```
mrannanj, line 197: If this is the correct change, then this if condition is shared with the block above, so could add this new line to that body. However, I'm uncertain this is the correct change.

mrannanj (approved):

mrannanj (changes requested):
Apparently there is confusion if this is the correct change still, so removing approval.

mrannanj (approved):

PR #1020, opened 2025-03-22
Use fixed size string in video settings

Nopey:
While testing https://github.com/omf2097/openomf/pull/898, the video settings' "renderer" string was some version of wild free (could have been a bug from master? wasn't able to repro with openomf_0.7.2-524-g071faf17_windows.zip release)

ASan didn't feel like giving me any more info (no alloc site, first free site, or anything), and it occured to me that the settings library would probably support fixed sized strings (which, it does!).

So, no more dynamically allocated `renderer` and associated bugs.


Vagabond (approved):

Vagabond:
We could probably use this strategy a few other places in the settings

PR #1023, opened 2025-03-23
Tournament REC fixes

Nopey:
- Tournament pilots names no longer get saved as "Crystal"
- Tournament pilots colors are now saved, and are loaded correctly (without being overwritten by colors from ALTPALs.)
- Watching player 1 win a tournament match in a REC no longer crashes the game trying to rank them up.
- Tournament pilots photos are now saved and loaded to REC, visible in replays.

Also: tested working with .RECs from DOS, ~~although the tournament rec caused some unrelated pointer problems regarding the round-end tournament score calculations.~~ fixed that crash.

### DIFF
--- a/src/formats/pilot.h
+++ b/src/formats/pilot.h
@@ -41,9 +41,9 @@ typedef struct {
     uint16_t offense;        ///< Offense preference value (100 high, should be under 200).
     uint16_t defense;        ///< Defense preference value (100 high, should be under 200).
     int32_t money;           ///< Amount of money the pilot currently has
-    uint8_t color_1;         ///< Color 1 field for the HAR (0-15). 255 means random.
-    uint8_t color_2;         ///< Color 2 field for the HAR (0-15). 255 means random.
-    uint8_t color_3;         ///< Color 3 field for the HAR (0-15). 255 means random.
+    uint8_t color_1;         ///< HAR Color 1. 0-15 are altpals, 16 means use 'palette' field, 255 means random.
+    uint8_t color_2;         ///< HAR Color 2. 0-15 are altpals, 16 means use 'palette' field, 255 means random.
+    uint8_t color_3;         ///< HAR Color 3. 0-15 are altpals, 16 means use 'palette' field, 255 means random.
     char trn_name[13];       ///< Tournament file
     char trn_desc[31];       ///< Tournament description
     char trn_image[13];      ///< Tournament image file
@@ -102,7 +102,7 @@ typedef struct {
     uint32_t total_value; ///< Total value for the pilot
     float unk_f_a;        ///< Unknown
     float unk_f_b;        ///< Unknown
-    vga_palette palette;  ///< Palette for photo ?
+    vga_palette palette;  ///< Photo palette, used as HAR colors when not using altpals.
     uint16_t unk_block_i; ///< Unknown
     uint16_t photo_id;    ///< Which face photo this pilot uses
 
--- a/src/formats/pilot.h
+++ b/src/formats/pilot.h
@@ -132,6 +132,9 @@ typedef enum
 int sd_pilot_create(sd_pilot *pilot);
 
 void sd_pilot_clone(sd_pilot *dest, const sd_pilot *src);
+/*! \brief Copies an sd_pilot, but not any child allocations (quotes, photo surface, etc..)
+ */
+void sd_pilot_copy_shallow(sd_pilot *dest, const sd_pilot *src);
 
 /*! \brief Free pilot structure
  *
--- a/src/formats/sprite.c
+++ b/src/formats/sprite.c
@@ -289,6 +289,9 @@ int sd_sprite_vga_decode(sd_vga_image *dst, const sd_sprite *src) {
         return SD_SUCCESS;
     }
 
+    // the number of bytes that can be written to `dst`.
+    unsigned int dst_size = dst->w * dst->h;
+
     // Walk through raw sprite data
     while(i < src->len) {
         // read a word
@@ -308,7 +311,13 @@ int sd_sprite_vga_decode(sd_vga_image *dst, const sd_sprite *src) {
             case 1:
                 while(data > 0) {
                     uint8_t b = src->data[i];
-                    int pos = ((y * src->width) + x);
+                    unsigned int pos = ((y * src->width) + x);
+                    // if we're about to overflow the `dst` buffer, don't.
+                    if(pos >= dst_size) {
+                        // it's not pretty, but this bounds check is necessary & used.
+                        // without it, tournament rec files can't load their pilot pics
+                        return SD_SUCCESS;
+                    }
                     dst->data[pos] = b;
                     i++; // we read 1 byte
                     x++;
--- a/src/formats/sprite.c
+++ b/src/formats/sprite.c
@@ -5,6 +5,7 @@
 #include "formats/palette.h"
 #include "formats/sprite.h"
 #include "utils/allocator.h"
+#include "utils/log.h"
 
 int sd_sprite_create(sd_sprite *sprite) {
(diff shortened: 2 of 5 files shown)


Vagabond (dismissed):

Nopey:
<details><summary>DOS OMF doesn't agree with how this PR records colors.</summary>

openomf can play back a rec file from dos and [gets the correct colors](https://github.com/user-attachments/assets/8f13cd61-e8c3-4e17-aeff-13149ca3b488).
here is openomf playing back a rec from dos:

image

(and, for comparsion, same rec in openomf)
image
hey! how is dos omf getting this bright orange color on Particular2's Jaguar?

</details>

unsure what we're doing wrong when recording the rec, although i don't think it matters terribly much? we can probably write it off as a visual bug that requires particular circumstances.

mrannanj (dismissed):

src/formats/sprite.c
```diff
@@ -308,7 +310,12 @@ int sd_sprite_vga_decode(sd_vga_image *dst, const sd_sprite *src) {
             case 1:
                 while(data > 0) {
                     uint8_t b = src->data[i];
-                    int pos = ((y * src->width) + x);
+                    unsigned int pos = ((y * src->width) + x);
+                    if(pos >= dst_size) {
```
Vagabond, line 314: Can you comment exactly what we're checking for here?

src/game/game_state.c
```diff
@@ -180,9 +180,11 @@ int game_state_create(game_state *gs, engine_init_flags *init_flags) {
             omf_free(gs->players[i]->pilot);
             gs->players[i]->pilot = &gs->rec->pilots[i].info;
             // this function alters the palette
-            sd_pilot_set_player_color(gs->players[i]->pilot, PRIMARY, gs->rec->pilots[i].info.color_3);
-            sd_pilot_set_player_color(gs->players[i]->pilot, SECONDARY, gs->rec->pilots[i].info.color_2);
-            sd_pilot_set_player_color(gs->players[i]->pilot, TERTIARY, gs->rec->pilots[i].info.color_1);
+            if(i == 0 || gs->rec->game_mode != REC_GAMEMODE_TOURNAMENT) {
```
Vagabond, line 183: Maybe comment that we always use altpals for the first player (although I believe it's possible to have a tournament player that has custom colors from the PIC file).

Nopey: Just realized: with the check I added in "Don't load altpals index 16 (default in omf tournament mode", this `if` is redundant.

Thanks for bringing my attention back to it.

src/formats/sprite.c
Nopey: Added a comment on dst_size and on the if, explaining that this is bounds checking for the `dst` buffer.

src/formats/pilot.h
```diff
@@ -41,9 +41,9 @@ typedef struct {
     uint16_t offense;        ///< Offense preference value (100 high, should be under 200).
     uint16_t defense;        ///< Defense preference value (100 high, should be under 200).
     int32_t money;           ///< Amount of money the pilot currently has
-    uint8_t color_1;         ///< Color 1 field for the HAR (0-15). 255 means random.
-    uint8_t color_2;         ///< Color 2 field for the HAR (0-15). 255 means random.
-    uint8_t color_3;         ///< Color 3 field for the HAR (0-15). 255 means random.
+    uint8_t color_1;         ///< Color 1 field for the HAR (0-15). 255 means random. 16 is default in tournament.
```
Vagabond, line 44: by "default" do you mean it came from the PIC?

src/formats/sprite.c
```diff
@@ -308,7 +311,13 @@ int sd_sprite_vga_decode(sd_vga_image *dst, const sd_sprite *src) {
             case 1:
                 while(data > 0) {
                     uint8_t b = src->data[i];
-                    int pos = ((y * src->width) + x);
+                    unsigned int pos = ((y * src->width) + x);
+                    // if we're about to overwrite the `dst` buffer, don't.
```
Vagabond, line 315:
```suggestion
                    // if we're about to overflow the `dst` buffer, don't.
```

src/formats/pilot.h
Nopey: when creating a tournament character, their color_1-3 is 16, and yes the color comes from the PIC

src/formats/sprite.c
Nopey: i will apply this wording change when i know what to do for the other conversation

src/formats/pilot.h
Nopey:
```c
src\formats\pilot.h
105:    vga_palette palette;  ///< Palette for photo ?
```
should probably update this comment too.
```
//< Palette for photo and HAR (if not using altpals)
```

Vagabond: so 0-16 are ALTPAL colors, 16 is color from the PIC and 255 is random?

Nopey: 0-15 are definitely the ALTPAL colors.
It might be the case that 16-255 is "use whatever is in the palette field", and the palette field as you mention is initialized from the PIC.

Just checked the git blame, and the "255 means random" comment comes from katajakasa in 2014:
commit 722ca232a162bf7aa20e2388b4154afd710c6f92
I don't think 255=random was ever implemented, unless we consider the colors from the PIC photo "random".

src/formats/sprite.c
```diff
@@ -308,7 +311,13 @@ int sd_sprite_vga_decode(sd_vga_image *dst, const sd_sprite *src) {
             case 1:
                 while(data > 0) {
                     uint8_t b = src->data[i];
-                    int pos = ((y * src->width) + x);
+                    unsigned int pos = ((y * src->width) + x);
+                    // if we're about to overflow the `dst` buffer, don't.
+                    if(pos >= dst_size) {
+                        // it's not pretty, but this bounds check is necessary & used.
+                        // without it, tournament rec files can't load their pilot pics
```
Vagabond, line 318: Do we know why? is this a bug in the original engine?

Vagabond (approved):

src/formats/sprite.c
Nopey: This overflow does get triggered when loading a REC from the original engine in OMF, such as attached [2025_TRN.zip](https://github.com/user-attachments/files/19436823/2025_TRN.zip).

I believe this is an openomf bug, where we are failing to recreate a quirk of the original.
See also:
```c
// (pic.c, lines 88-97)
        // Sprite
        pic->photos[i]->sprite = omf_calloc(1, sizeof(sd_sprite));
        sd_sprite_create(pic->photos[i]->sprite);
        if((ret = sd_sprite_load(r, pic->photos[i]->sprite)) != SD_SUCCESS) {
            goto error_1;
        }

        // Fix length and width
        pic->photos[i]->sprite->height++;
        pic->photos[i]->sprite->width++;

// (chr.c, lines 291-299)
    // Save photo. Hacky size fix.
    chr->photo->width--;
    chr->photo->height--;

    if(SD_SUCCESS != sd_sprite_save(w, chr->photo)) {
        return SD_FILE_WRITE_ERROR;
    }
    chr->photo->width++;
    chr->photo->height++;
```
whereas no similar width/height adjustment is being applied to the photos when loading from REC.

I like having a bounds check in sd_sprite_vga_decode, although i'd rather overflow be an error instead of silently dropping the excess data.

Vagabond: Yeah, we never figured out that off by one in the width/height. Maybe we need to reverse it here? In any case leaving a more explicit comment in the code would help us fix it later.

Nopey: Added a log_warn when hitting the bounds check, and a TODO to figure out why this is happening.

Vagabond (approved):

PR #1024, opened 2025-03-23
Make inserting assertions into RECs easier

Vagabond:
Make it easier to insert assertions from rectool, only require the tick at which to assert, not the slot in the REC file


Nopey (approved):
rectool change looks good.

I'm unaware what's going on with the HAR2_START_POS, but I imagine you've got that figured out.

PR #1026, opened 2025-03-24
Misc. Gameplay Fixes

Nopey:
Various fixes for things that happen in-game.

### DIFF
--- a/src/game/scenes/arena.c
+++ b/src/game/scenes/arena.c
@@ -1566,14 +1566,20 @@ int arena_create(scene *scene) {
     local->over = 0;
     local->winner = 0;
 
-    // When playing the desert arena in Tournament mode,
-    // pick a random palette to change the time of day.
-    if(scene->bk_data->file_id == 128 && is_tournament(scene->gs)) {
-        int pal_index = rand_int(vector_size(&scene->bk_data->palettes));
-        // 0 is selected by default, so nothing to do if we hit that.
-        if(pal_index > 0) {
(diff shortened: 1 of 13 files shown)


src/game/scenes/arena.c
```diff
@@ -1524,13 +1562,20 @@ int arena_create(scene *scene) {
...
-        }
+    // TODO: Fire & Ice will need to set the arena palette
+    unsigned pal_index = 0;
+    if(scene->gs->init_flags->playback == 1) {
+        // use palette index from rec
+        pal_index = scene->gs->rec->arena_palette;
+    } else if(scene->bk_data->file_id == 128 && is_tournament(scene->gs)) {
+        // When playing the desert arena in Tournament mode,
+        // pick a random palette to change the time of day.
+        pal_index = rand_int(vector_size(&scene->bk_data->palettes));
```
Vagabond, line 1573: According to @LegendaryBlueShirt  this should only happen on the final round (and maybe only if you're playing one of the higher round counts)

Nopey: Tournament mode does not have multiple rounds.

Vagabond: oh yes, my bad

Vagabond:
If you keep losing in tournament mode, it seems to use the same arena constantly?

Nopey:
rebased, and pushed a fix commit to randomize tournament mode arenas

Vagabond (approved):

src/game/scenes/arena.c
LegendaryBlueShirt: So I did some testing and locked the stage select to the desert in tournament mode with CheatEngine and... yeah it's random!  Cool.

Issue #935, opened 2025-03-01
Swap color_1 and color_3 entries in pilot structure


katajakasa:
They are currently the wrong  way around.

Nopey:
What order should they be on-disk?
Currently, sd_pilot_save_player_to_mem and sd_pilot_load_player_from_mem disagree:
```c
// save
    memwrite_dword(w, pilot->money);
    memwrite_ubyte(w, pilot->color_1); // TERTIARY
    memwrite_ubyte(w, pilot->color_2); // SECONDARY
    memwrite_ubyte(w, pilot->color_3); // PRIMARY

// load
    pilot->money = memread_dword(mr);
    sd_pilot_set_player_color(pilot, PRIMARY, memread_ubyte(mr));
    sd_pilot_set_player_color(pilot, SECONDARY, memread_ubyte(mr));
    sd_pilot_set_player_color(pilot, TERTIARY, memread_ubyte(mr));
```

I also recently removed this code without noticing it was switching the order (in commit 80de4b8970619c2c7b34af80a50289d4876ec62b):
```c
            scene->gs->rec->pilots[i].info.color_3 = player->pilot->color_1;
            scene->gs->rec->pilots[i].info.color_2 = player->pilot->color_2;
            scene->gs->rec->pilots[i].info.color_1 = player->pilot->color_3;
```
This code switched the order in-memory before serializing the rec to disk. (The colors are now wrong in the .REC, oops).

LegendaryBlueShirt:
The correct on-disk order is reversed.

Nopey:
![Image](https://github.com/user-attachments/assets/90ef27cb-42b6-4153-ab79-35f7b7de7af4)
So we're on the same page:
This jaguar's PRIMARY color is black, their SECONDARY is purple, and TERTIARY is red.

And the correct on-disk order is TERTIARY, then SECONDARY, then PRIMARY.

PR #1027, opened 2025-03-25
Correct color order, and fix tournament colors save-compatibility

Nopey:
I tested these changes by copy-pasting CHR's and RECs between OMF and openomf.

- Colors are now ordered correctly: TERTIARY, followed by SECONDARY, and finally PRIMARY.
- Tournament save-games from DOS now load fully: Previously, the HAR colors would be all-black.
- The mechlab now lets you select the PIC colors (color number 16) in the customization screen.
- A crash was fixed, where by mashing F3 and ALT-F4 the game would try to save a recording after the recording was already freed and nulled.

### DIFF
--- a/src/engine.c
+++ b/src/engine.c
@@ -184,7 +184,7 @@ void engine_run(engine_init_flags *init_flags) {
                         save_palette_shot();
                     }
                     if(e.key.keysym.sym == SDLK_F3) {
-                        if(init_flags->playback != 1) {
+                        if(init_flags->playback != 1 && gs->rec) {
                             save_rec(gs);
                         }
                     }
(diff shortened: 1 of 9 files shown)


Nopey:
gonna see if i can't straighten these colors out instead of reinstating the hack

Nopey:
I tested these changes by copy-pasting CHR's and RECs between OMF and openomf.

In addition to fixing the color ordering problem, our tournament mode now loads the HAR colors from altpals.dat/players.pic on save game load-- this solves the problem where OMF savefiles loaded as all black.

LegendaryBlueShirt (approved):

Vagabond:
Looks like this should fix #935?

Vagabond (approved):

PR #1028, opened 2025-03-25
Don't idle during defeat

Nopey:
When the defeated HAR idles during defeat, they softlock the game.


Nopey:
the rectest has the following three assertions:
```
--assert --assertion_tick 515 --operand1=har1 --operand1_value=health --operator=eq --operand2=literal --operand2_value=0
--assert --assertion_tick 515 --operand1=har1 --operand1_value=state --operator=eq --operand2=literal --operand2_value=8
--assert --assertion_tick 516 --operand1=har1 --operand1_value=state --operator=eq --operand2=literal --operand2_value=13
```
which ensure that if the rectest desyncs in any significant way, it'll fail.

Vagabond (approved):

PR #1033, opened 2025-03-25
Include math.h in miscmath.h for fabsf()

katajakasa:


katajakasa:
> In a vacuum I am unsure what issue this PR solves

It resolves the issue of miscmath.h not having math.h included for fabsf().

Vagabond (approved):

PR #1036, opened 2025-03-25
Implement time travel for the visual debugger

Vagabond:
The REC controller will take a game state snapshot every 10 ticks, and using backspace when replaying a REC will step the game state back to the last snapshot.

### DIFF
--- a/src/controller/rec_controller.c
+++ b/src/controller/rec_controller.c
@@ -3,6 +3,7 @@
 #include "formats/rec_assertion.h"
 #include "game/game_player.h"
 #include "game/game_state_type.h"
+#include "game/protos/scene.h"
 #include "utils/allocator.h"
 #include "utils/log.h"
 #include <inttypes.h>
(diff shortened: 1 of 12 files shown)


src/controller/rec_controller.c
```diff
@@ -190,11 +201,56 @@ int rec_controller_poll(controller *ctrl, ctrl_event **ev) {
...
+            c->gs = gs_new;
+        }
+    }
+}
+
 void rec_controller_create(controller *ctrl, int player, sd_rec_file *rec) {
     wtf *data = omf_calloc(1, sizeof(wtf));
     data->last_tick = 0;
     data->last_action = ACT_STOP;
+    data->id = player;
```
Nopey, line 251: please rename wtf's id field to player_id

Nopey (commented):
Is there a reason this feature is part of the rec controller?

When debugging other sorts of matches, being able to time travel may help-- and i've been thinking about adding a way to change the kind of a controller mid-round (to, for example, take keyboard control of a HAR after a REC sets up a scenario)

Vagabond:
> Is there a reason this feature is part of the rec controller?
> 
> When debugging other sorts of matches, being able to time travel may help-- and i've been thinking about adding a way to change the kind of a controller mid-round (to, for example, take keyboard control of a HAR after a REC sets up a scenario)

This was the most immediately useful thing to do, rewinding "live" controllers seems like it'd be harder. I don't object to doing it, but I think we'd need to do more planning/testing.

src/controller/rec_controller.c
Vagabond: Done, and I renamed the struct as well.

Nopey (dismissed):
Useful. Rewinding non-rec gameplay can be considered at a later date.

PR #1038, opened 2025-03-26
Set executing_move flag when transitioning into next_move

Nopey:
Electra's P656 plays move 35 (the forward spinny attack), followed by move 36 upon hitting an enemy (the spinny recoil through the air).

During the next_move, move 36, the executing_move flag should be set to prevent the player from interrupting the animation with something like move 22 (P41).


LegendaryBlueShirt:
The executing_move flag has been responsible for two bugs now that I've seen in short time I've been actively contributing to the codebase, sooner or later I'm going to push to remove it in favor of doing it the way OMF does... checking if you're in an idle animation or not.

LegendaryBlueShirt (approved):

PR #1039, opened 2025-03-26
Use 'sed' instead of 'xargs' to trim whitespace in run_rectests.sh

Nopey:
This permits quote marks in rectest descriptions.

### DIFF
--- a/run_rectests.sh
+++ b/run_rectests.sh
@@ -51,8 +51,8 @@ i=0
 for test in "${tests[@]}"; do
     IFS=':' read -r desc filename <<< "$test"
     # Trim whitespace from description and filename
-    desc=$(echo "$desc" | xargs)
-    filename=$(echo "$filename" | xargs)
+    desc=$(echo "$desc" | sed -re 's/^[[:blank:]]+|[[:blank:]]+$//g')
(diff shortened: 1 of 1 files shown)


Vagabond:
What exactly is the regular expression doing?

Nopey:
Removing all leading and trailing whitespace.

step-by-step unwrap:
- `s/^[[:blank:]]+|[[:blank:]]+$//g`
`s/...//g` is to replace all `...` with nothing.
- `^[[:blank:]]+|[[:blank:]]+$`
the alternation operator `|` in the middle grabs everything around it--
an equivalent expression would be `(^[[:blank:]]+)|([[:blank:]]+$)`
It means we'll replace any `^[[:blank:]]+`'s we find, and also any `[[:blank:]]+$`'s we find.
- `^[[:blank:]]+`
^ matches the start of the line, `[[:space:]]` matches whitespace, and the `[[:space:]]` is being repeated by the `+` one or more times.
So `^[[:blank:]]+` matches one or more whitespace characters at the start of the line.
- `[[:blank:]]+$`
Same as above, but with `$` matching the end of the line rather than the start.

Vagabond:
Ah, I was not familiar with the alternation operator in sed.

Vagabond (approved):

PR #1040, opened 2025-03-26
Face HAR to the direction they're attacked from

Nopey:
If the opponent does not walk during Chronos matter phasing, they will not flip; this much is a feature. However, if Chronos's matter phasing kick hits their back, they need to face Chronos and receive knockback in the correct direction.

The rectest waits until tick 112 before checking what direction Pyro is being knocked back in, because that's the tick where Chronos's kick lands after PR #1009 fixes wall interactions; without PR 1009, the kick lands slightly earlier on tick 103.

I do not know of a case where a projectile can hit a player in the back, so have not added the same logic to har_collide_with_projectile. Does anyone know of such a case, or know what DOS does in these cases?


LegendaryBlueShirt:
[TURNAROUND.zip](https://github.com/user-attachments/files/19462366/TURNAROUND.zip)
It is with great effort that I reproduced the backwards hitting projectile case

Vagabond:
You can make this rectest run much faster by deleting the final STOP action from it:

```
./rectool -f ../rectests/KNOCKBACK-DIR.REC --delete=14 -o ../rectests/KNOCKBACK-DIR.REC
```

Nopey:
Deleted excess moves from ELECTRA-MOVE36-EMFLAG.REC, KNOCKBACK-DIR.REC, KREISSACK.REC, and NO-IDLE-DEFEAT.REC

Vagabond (approved):

Issue #1012, opened 2025-03-18
[GAMEPLAY] Movement Rounding Errors


LegendaryBlueShirt:
**Steps to reproduce**
Choose a low agility pilot, walk.

What happened?
Speed is more sluggish than it should be

What did you expect to happen?
Raven should move forward around 20% faster

**Documentation**
Cheatengine shows Raven moves forward at a speed of 880 units, but OpenOMF only stores whole pixel coordinates so Raven only moves 768 units instead.

Nopey:
Quite the opposite-- Raven is moving too fast, actually.
Here is a comparison between Raven's walkspeed in DOS and OpenOMF, with the same REC.

[RAVENWALK.zip](https://github.com/user-attachments/files/19465283/RAVENWALK.zip)

DOS:

https://github.com/user-attachments/assets/8aca5333-7675-424a-b3e5-9f42c1127f9c

OpenOMF:

https://github.com/user-attachments/assets/b1615e32-0b39-42bf-8349-451576361f06

Notice how much longer Raven spends in OpenOMF walking left against the wall, and how his punch is performed after walking against the wall, rather than before.

Perhaps you're feeling a difference in game speed?

Nopey:
Just noticed that Raven didn't turn around in OpenOMF, which is a bug and results in him getting his fwd_speed the whole time.
Gonna do a quick n' dirty patch for that locally and see how it looks

Nopey:
punch comparison:
DOS
![Image](https://github.com/user-attachments/assets/e5f7151a-89cb-46bc-a4a1-9736cbf80547)
OpenOMF
![Image](https://github.com/user-attachments/assets/5125ba61-5e7f-438c-bc1f-c0ec218474f6)

Nopey:
Effect varies by HAR as well as Pilot, as walkspeed depends on both.
Raven piloting a Chronos is poorly effected by this.

PR #1041, opened 2025-03-26
Face jumping enemy

Nopey:
As Raven walks to the right, he should turn and face Shirro-- despite Shirro jumping.
This PR currently just adds a rectest, I'm working on something else right now so feel free to pick this up.

Bug discovered in issue https://github.com/omf2097/openomf/issues/1012 conversation.


Vagabond:
I pushed a fix that does fix the rectest, and doesn't break any other rectests, and reflects what I believe @LegendaryBlueShirt said about how the original uses directional keys as triggers to change direction even if the opponent is midair.

tools/rectool/main.c
```diff
@@ -468,6 +473,29 @@ int main(int argc, char *argv[]) {
             offset++;
         }
         printf("Deleted %d moves, final move count %d\n", offset, rec.move_count);
+    } else if(truncate->count > 0) {
+        unsigned truncate_tick = truncate->ival[0] < 0 ? 0 : truncate->ival[0];
```
Vagabond, line 477: You could also just fail if a negative number was supplied instead of silently turning it into 0, which will truncate all the entries?

tools/rectool/main.c
```diff
@@ -468,6 +473,29 @@ int main(int argc, char *argv[]) {
...
+            if(sd_rec_delete_action(&rec, rec.move_count - 1) != SD_SUCCESS) {
+                printf("deleting move %d failed\n", rec.move_count - 1);
+                abort();
+            }
+            deleted++;
+        }
+        if((rec.move_count == 0 || rec.moves[rec.move_count - 1].tick < truncate_tick) && truncate->ival[0] >= 0) {
+            sd_rec_move mv;
+            memset(&mv, 0, sizeof(sd_rec_move));
+            mv.action;
```
Vagabond, line 490: unused.

tools/rectool/main.c
Nopey: yeah. that was part of the "truncate inserts a stop action" behavior i just scrapped

Vagabond (approved):

PR #1044, opened 2025-03-26
Use fixed point math for position, endurance, health

Nopey:
This PR hopes to improve netplay and rec compatibility & determinism across platforms.

Positions and velocities (vec2f) are now stored as 24.8 fixed point numbers, which means there are 24 bits for the whole part and 8 for the fractional part.

This PR has the annoying quirk of taking an existing name and appending 'f' to mark it as fixedpoint, some of these should be un-renamed before merging-- this was done to ensure non-updated code would compiler error while writing the PR.


src/game/protos/object.h
```diff
@@ -240,6 +240,7 @@ int object_get_pal_limit(const object *obj);
 
 vec2i object_get_size(const object *obj);
 vec2i object_get_pos(const object *obj);
```
Nopey, line 242: object_get_pos still returns a vec2i.
it should probably be updated to return a vec2f.

src/game/protos/object.c
```diff
@@ -568,7 +570,10 @@ void object_set_animation(object *obj, animation *ani) {
 
     // Debug texts
     if(obj->cur_animation->id == -1) {
-        log_debug("Custom object set to (x,y) = (%f,%f).", obj->pos.x, obj->pos.y);
+        char bufx[FIXEDPT_STR_BUFSIZE], bufy[FIXEDPT_STR_BUFSIZE];
+        fixedpt_str(obj->pos.fx, bufx, -1);
+        fixedpt_str(obj->pos.fy, bufy, -1);
+        log_debug("Custom object set to (x,y) = (%s,%s).", bufx, bufy);
```
Nopey, line 576: Fixed point numbers can be formatted without conversion to floating point, but it's a bit verbose-- and fixedpt_str is a scary function (it writes to a string with no bounds checking-- like its 1982, despite the library being written in the modern year of 2012)

src/game/protos/player.c
```diff
@@ -146,16 +146,16 @@ void player_describe_frame(const sd_script_frame *frame) {
 
 void player_describe_object(object *obj) {
     log_debug("Object:");
-    log_debug("  - Start: %d, %d", obj->start.x, obj->start.y);
-    log_debug("  - Position: %d, %d", obj->pos.x, obj->pos.y);
-    log_debug("  - Velocity: %d, %d", obj->vel.x, obj->vel.y);
+    log_debug("  - Start:    %f, %f", fixedpt_tofloat(obj->start.fx), fixedpt_tofloat(obj->start.fy));
+    log_debug("  - Position: %f, %f", fixedpt_tofloat(obj->pos.fx), fixedpt_tofloat(obj->pos.fy));
+    log_debug("  - Velocity: %f, %f", fixedpt_tofloat(obj->vel.fx), fixedpt_tofloat(obj->vel.fy));
```
Nopey, line 151: This is the less verbose way of printing a fixed point number-- by converting it to a float, we can use float format strings.

src/utils/vec.h
```diff
@@ -85,10 +93,17 @@ static inline vec2i vec2i_create(int x, int y) {
...
+static inline vec2f vec2f_createf(fixedpt x, fixedpt y) {
     vec2f v;
-    v.x = x;
-    v.y = y;
+    v.fx = x;
+    v.fy = y;
+    return v;
+}
+
+static inline vec2f vec2f_from_float(float x, float y) {
```
Nopey, line 103: unused

katajakasa:
This is more precise, but it's also probably less accurate. The original game retains values as integers, where positions are pixel coords multiplied by 256. That is probably what openomf also needs to do to remain as close to omf2097 as possible.

LegendaryBlueShirt:
> This is more precise, but it's also probably less accurate. The original game retains values as integers, where positions are pixel coords multiplied by 256. That is probably what openomf also needs to do to remain as close to omf2097 as possible.

I believe you have the exact same misconception I had at first glance, that is precisely what is being described with the fixed point type, it's just not the name we had for it

katajakasa:
> > This is more precise, but it's also probably less accurate. The original game retains values as integers, where positions are pixel coords multiplied by 256. That is probably what openomf also needs to do to remain as close to omf2097 as possible.
> 
> I believe you have the exact same misconception I had at first glance, that is precisely what is being described with the fixed point type, it's just not the name we had for it

Ah. I see. I assumed a standard fixed point library.

src/formats/af.h
```diff
@@ -20,17 +20,17 @@
...
-    uint16_t health;      ///< HAR Health
-    float forward_speed;  ///< HAR fwd speed
-    float reverse_speed;  ///< HAR bwd speed
-    float jump_speed;     ///< HAR jump speed
-    float fall_speed;     ///< HAR fall speed
-    uint8_t unknown_c;    ///< Unknown value
-    uint8_t unknown_d;    ///< Unknown value
+    uint16_t file_id;      ///< File ID
+    uint16_t exec_window;  ///< Move execution window (?)
+    uint32_t endurance;    ///< HAR Endurance
```
Vagabond, line 25: I think the original game has endurance/stun go negative, so uint may not be a good idea here.

Nopey: uint32_t was chosen here to match the `sd_read_udword` that af.c uses to read this value.

I'll make har_endurance_t an int32_t, and change this to be an int32_t and use `sd_read_dword` (bonus: making this signed lets me nuke the saturating sub helper).

src/game/objects/har.h
```diff
@@ -21,6 +21,15 @@
 #define LAYER_PROJECTILE 0x20
 #define LAYER_HAZARD 0x40
 
+// Define what 1.0 is stored as in HAR endurance's fixed-point format
+#define HAR_ENDURANCE_FIXP_ONE 512
+typedef uint32_t har_endurance_t; // 32 bits should be enough for anybody
```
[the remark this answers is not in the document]
Nopey: I've removed the endurance fixedpoint scaling, and switched it to int32_t.

Nopey:
Will rebase onto latest master once #1054 is merged.

src/game/objects/arena_constraints.h
```diff
@@ -5,4 +5,9 @@
 #define ARENA_RIGHT_WALL 300
 #define ARENA_FLOOR 190
 
+// fixed point versions
```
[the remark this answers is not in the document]
Nopey: ARENA_FLOOR is important for object rendering, and we render in pixels-- not fixedpt fractions of pixels.
The arena walls can be fixedpt only, sure.

src/game/gui/trn_menu.c
```diff
@@ -165,15 +165,15 @@ static void trnmenu_layout(component *c, int x, int y, int w, int h) {
 }
 
 static vec2f center(component *c) {
-    return vec2f_create(c->x + c->w / 2, c->y + c->h / 2);
+    return vec2f_create(fixedpt_fromint(c->x + c->w) / 2, fixedpt_fromint(c->y + c->h) / 2);
```
LegendaryBlueShirt, line 168: Looks like the math is changing here due to order of operations

LegendaryBlueShirt:
Would it make sense//Is it possible to split this up into multiple PRs?  We can start with one PR for the fixed point library at least

Nopey:
Fixed-point health, damage, endurance is a good candidate for being split off to a separate PR; these do not use the 24.8 fixed point format that the on-screen positions use (24.8 implemented here in fixedptc.h).

When reviewing a PR like this, I suggest using the `Commits` tab of GitHub, as opposed to the `Changed Files` tab; I split the work across a dozen commits for readability and reviewability.

Issue #649, opened 2024-10-10
Renderer wishlist/buglist


katajakasa:
Just a big lump of stuff for now

TODO:
- [x] Make the shaders use remappings to get correct shadows and stuff https://github.com/omf2097/openomf/pull/633
- [x] Figure out the video renderer API -- we need an easy way to add new renderers.
- [x] Text shadows are broken, fix them
- [x] Intro rendering is broken, fix it
- [x] HAR transparency (bf tags) is broken, fix it
- [x] HAR shadows are broken, fix them (under har shadow)
- [x] HAR shadow trails are broken, fix them (jaguar leap)
- [x] Credits transparency is broken, fix it
- [x] Menu text (blue grid) background transparency is broken, fix it
- [x] Palettes should be built in tick functions, and then be complete when fed to the renderer. Also, add dirtiness flag and damage information.

katajakasa:
I moved the non-completed tasks to separate tickets and dropped from here. Let's close this one as completed.

PR #1047, opened 2025-03-27
gl3: Implement Electra & Pyros palette quirks

Nopey:
Electra electricity and Pyros Fire use palette indexes from outside the HAR range (first 48 colors), and should skip remapping to ensure they remain bright and colorful.

Part of issue #649

<details><summary>Visual Comparison, on Stadium (positional lighting remap is in effect)</summary>

Before:
image
After:
image

</details>

### DIFF
--- a/shaders/palette.frag
+++ b/shaders/palette.frag
@@ -19,6 +19,7 @@ in vec4 gl_FragCoord;
 bool REMAP_SPRITE = (options & 1u) != 0u;
 bool SPRITE_MASK = (options & 2u) != 0u;
 bool SPRITE_INDEX_ADD = (options & 4u) != 0u;
+bool USE_HAR_QUIRKS = (options & 8u) != 0u;
 
 
 float PHI = 1.61803398874989484820459;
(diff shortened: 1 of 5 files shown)


Nopey:
Note that later objects' draw calls can still apply remaps to the electricity and fire, such as the pause menu's transparent background here:
image

LegendaryBlueShirt (approved):
womm

Vagabond (approved):

shaders/palette.frag
```diff
@@ -65,13 +66,15 @@ void main() {
 
     vec4 remap = texture(remaps, vec2(texel.r, remap_offset / 18.0));
 
+    bool NO_REMAP = USE_HAR_QUIRKS && index > 0x30;
```
katajakasa, line 69: Wont this only work for HAR 1 ?

Nopey: No, the `index` is taken directly from the input texture-- which is the same for both players.

Nopey: a previous version of the patch used the texel.r (after pal_offset had been applied), but I thought using the unmodified texture data would lend itself better to out of order execution.

katajakasa: Har 1 palette is 1-48, and HAR2 palette is the next 48 colors. This will disable remaps for entire har2 palette when USE_HAR_QUIRKS is enabled for it, yes ?

katajakasa: Oh wait, never mind, I misread the file. This checks against the sprite index.

katajakasa (approved):

PR #1051, opened 2025-03-27
Make shards blockable

Vagabond:


LegendaryBlueShirt:
Technically achieved the intended result but ideally you get multiple blocks and pushes as a result of blocking the whole move

Vagabond:
Ah, I see, I also see why the HAR randomly crouches during the blockstun (because har_block is dumb). This needs more work.

Vagabond:
@LegendaryBlueShirt does this meet requirements now?

LegendaryBlueShirt (dismissed):

LegendaryBlueShirt (dismissed):

PR #1052, opened 2025-03-27
Fix issue when annotating RECs from console

Vagabond:
When playing back a REC, do not feed REC inputs into a REC again

### DIFF
--- a/src/game/game_state.c
+++ b/src/game/game_state.c
@@ -1430,3 +1430,7 @@ bool is_demoplay(game_state *gs) {
 bool is_twoplayer(game_state *gs) {
     return !is_demoplay(gs) && !is_netplay(gs) && !is_singleplayer(gs);
 }
+
+bool is_rec_playback(game_state *gs) {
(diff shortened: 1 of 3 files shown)


src/game/game_state.c
```diff
@@ -1430,3 +1430,7 @@ bool is_demoplay(game_state *gs) {
 bool is_twoplayer(game_state *gs) {
     return !is_demoplay(gs) && !is_netplay(gs) && !is_singleplayer(gs);
 }
+
+bool is_rec_playback(game_state *gs) {
+    return strlen(gs->init_flags->rec_file) > 0 && gs->init_flags->playback == 1;
```
Nopey, line 1435:
```suggestion
    return gs->init_flags->rec_file[0] != '\0' && gs->init_flags->playback == 1;
```
If you don't need the string length, don't call strlen.

Nopey (changes requested):

Nopey (approved):

PR #1053, opened 2025-03-27
gl3: Framerate limiter.

katajakasa:
Framerate limiter + minor fix

### DIFF
--- a/src/video/renderers/opengl3/gl3_renderer.c
+++ b/src/video/renderers/opengl3/gl3_renderer.c
@@ -40,6 +40,9 @@ typedef struct gl3_context {
     bool draw_atlas;
     SDL_Rect culling_area;
 
+    uint64_t framerate_limit;
+    uint64_t last_tick;
+
     object_array_blend_mode current_blend_mode;
     GLuint palette_prog_id;
     GLuint rgba_prog_id;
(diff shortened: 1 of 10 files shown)


src/video/renderers/opengl3/gl3_renderer.c
```diff
@@ -76,8 +78,8 @@ static bool setup_context(void *userdata, int window_w, int window_h, bool fulls
     if(!create_gl_context(&ctx->gl_context, ctx->window)) {
         goto error_1;
     }
-    if(!set_vsync(ctx->vsync)) {
-        goto error_2;
+    if((ctx->fake_vsync = !set_vsync(ctx->vsync)) == true) {
```
Nopey, line 81:
```suggestion
    ctx->fake_vsync = !set_vsync(ctx->vsync);
    if(ctx->fake_vsync) {
```
Let's avoid assignment in conditionals unless there's a strong motivation.
Also, please remove useless boolean == true comparisons.

Nopey (changes requested):

src/video/renderers/opengl3/gl3_renderer.c
katajakasa: The == true check makes it clear that it is an assignment and a comparison. I like this better stylistically. Faster to read.

Nopey (approved):
changes look good, will be glad to hear my fans spin less.

SDL3 port tomorrow™️

katajakasa:
> changes look good, will be glad to hear my fans spin less.
> 
> SDL3 port tomorrow™️

Yep, my radeon sounded like it was self-destructing with vsync off :P

PR #1055, opened 2025-03-27
Do not interrupt Jaguar Throw with idle animations when landing

Nopey:
- Once an enemy has been grabbed, Jaguar's throw can no longer be interrupted by landing on the ground.
- Jaguar's throw now grabs the enemy on the first tick of ani 22, rather than the second.
- Jaguar's throw's grab damage no longer spawns scrap/oil
- Jaguar's throw no longer counts as a two hit combo

mildly unrelated bugfixes that are also included:
- Faulty logic that was dropping inputs during the first 25 static_ticks of an arena match has been fixed-- this logic now only applies to non-arena scenes. (this caused some RECs to desync, depending on the gamespeed)
- "HARS facing player" is no longer log_debug'd every tick a player walks-- it is now only printed when they turn around.


Nopey:
just gonna merge OHT.REC and JAG-THROW-LAND.REC, i think.

Nopey:
<details><summary>DOSBOX video showing that the overhead throw cannot be interrupted once the opponent has been grabbed.</summary>

https://github.com/user-attachments/assets/1f753286-531e-474f-82f9-fe0cb3483b04

</details>

src/game/objects/har.c
```diff
@@ -1381,11 +1381,18 @@ int har_collide_with_har(object *obj_a, object *obj_b, int loop) {
             // prevent next move from being interrupted
             a->executing_move = 1;
 
+            int retval = move->category == CAT_CLOSE ? 1 : 0;
+
+            if(loop == 0) {
+                // recurse so jaguar's overhead throw grabs enemy on first tick of anim 22
```
Vagabond, line 1387: So if the move chains to another move, check collision on that immediately?

src/game/objects/har.c
```diff
@@ -2332,13 +2339,11 @@ void har_face_enemy(object *obj, object *obj_enemy) {
...
-            log_debug("HARS facing player %d LEFT", h->player_id);
-            object_set_direction(obj, OBJECT_FACE_LEFT);
-        } else {
-            log_debug("HARS facing player %d RIGHT", h->player_id);
-            object_set_direction(obj, OBJECT_FACE_RIGHT);
+        int new_facing = pos.x > pos_enemy.x ? OBJECT_FACE_LEFT : OBJECT_FACE_RIGHT;
+        if(obj->direction != new_facing) {
+            log_debug("HARS facing player %d %s", h->player_id, new_facing == OBJECT_FACE_LEFT ? "LEFT" : "RIGHT");
         }
+        object_set_direction(obj, new_facing);
```
Vagabond, line 2346: You can likely put this inside the conditional as well?

src/game/objects/har.c
Nopey: Yeah, that's the thought

src/game/objects/har.c
Nopey: will do

src/game/protos/scene.c
```diff
@@ -205,9 +205,9 @@ void scene_dynamic_tick(scene *scene, int paused) {
 }
 
 void scene_input_poll(scene *scene) {
-    if(scene->static_ticks_since_start < 25) {
+    if(scene->static_ticks_since_start < 25 && !(SCENE_ARENA0 <= scene->id && scene->id <= SCENE_ARENA4)) {
```
Vagabond, line 208: There's a `scene_is_arena` function that would be more future-proof here, if we ever add more arenas.

Vagabond (approved):

PR #1056, opened 2025-03-28
Check if inputs changed recently before using for superjump or attack

### LINKED ISSUE #712: Not all jumps from crouching should be superjumps
In both OpenOMF and DOS, pressing crouch and then jump quickly will give you a superjump, that brings the top of your head just offscreen.

In DOS, if you hold the crouch button (down) for long enough (half a second? key repeat delay? who knows).. then jumping from a crouch gives you the normal jump.
In OpenOMF, no matter how long you hold the crouch button for, you will get the superjump.

I have not tested what the DOS behavior is with a joystick.
EDIT: changed wording to refer to it as a "superjump" to match what it's called in our codebase.

Nopey:
## Superjump Freshness
If a player presses the DOWN button on tick 90, the latest they can press UP and still get a super jump is tick 96.

This is true even if they are busy executing a move when they press the down button, such as the roundhouse kick in SUPERJUMP.REC.

Closes #712

## Attacks Freshness
When entering the extended jaguar leap input chain, P6321, the only time-sensitive part is after the player releases 3 and is holding 6-- they have 9 ticks to do so.


Nopey:
Marking as draft while I investigate moves.

When performing an extended jaguar leap (P6321), the only time-sensitive part of the input string is after the player lets go of 3 and is holding 6-- then, i suspect they only have 6 ticks to input P to perform the move.

Vagabond:
So was this determined by editing REC files and replaying them in the original engine?

Nopey:
Yes. INPUT-FRESHNESS.REC and SUPERJUMP.REC play back in the original engine, and if DOS support rec_assertions would be passing these.

Nopey:
Postponing kick-noturn rectest and associated changes as they're unrelated, and the rectest was desyncing in CI.

src/engine.c
```diff
@@ -197,8 +197,12 @@ void engine_run(engine_init_flags *init_flags) {
...
                         visual_debugger = !visual_debugger;
                     }
-                    if(!console_window_is_open() && e.key.keysym.sym == SDLK_SPACE) {
-                        debugger_proceed = 1;
+                    if(visual_debugger && !console_window_is_open() && e.key.keysym.sym == SDLK_SPACE) {
+                        dynamic_wait += 20;
+                        static_wait += 20;
+                    } else if(visual_debugger && !console_window_is_open() &&
+                              (e.key.keysym.sym >= SDLK_1 && e.key.keysym.sym <= SDLK_9)) {
+                        debugger_proceed = 1 + e.key.keysym.sym - SDLK_1;
```
Vagabond, line 205: So this allows using the number keys to step more?

src/game/objects/har.c
```diff
@@ -2119,9 +2103,8 @@ int har_act(object *obj, int act_type) {
         return 0;
     }
 
-    int oldstate = h->state;
-
-    af_move *move = match_move(obj, prefix, h->inputs);
+    char truncated_inputs[2] = {h->inputs[0], '\0'};
```
Vagabond, line 2106: Maybe a clarifying comment here? I think what this is doing is keeping the input buffer un-truncated as long as you have input a new key in the last 9 ticks? So as long as you keep the input buffer "hot" with fresh input, it won't be considered truncated?

src/engine.c
Nopey: Yeah. `1` is also guaranteed to step exactly one dynamic tick, whereas spacebar steps some number of milliseconds.

src/game/objects/har.c
Nopey: Yeah, that's what it's doing. A comment is in order.

If you don't have any suggestions, I'll write one in the morning.

Vagabond: I'll add one if you're otherwise occupied.

src/engine.c
```diff
@@ -337,10 +337,15 @@ void engine_run(engine_init_flags *init_flags) {
             // Tick dynamic features. This is a dynamically changing tick, and it depends on things such as
             // hit-pause, hit slowdown and game-speed slider. It is meant for ticking everything that has to do
             // with the actual gameplay stuff.
-            has_dynamic = dynamic_wait > game_state_ms_per_dyntick(gs);
+            int dyntick_ms = game_state_ms_per_dyntick(gs);
+            if(debugger_proceed > 0) {
+                dynamic_wait += dyntick_ms;
```
Vagabond, line 342: I'm not sure it's wise to not increment static_wait here as well? This seems to avoid some things like controller_tick from happening, which breaks REC rewind if you only advance the REC using the number keys now.

src/game/objects/har.c
Nopey: Thanks for adding the comment, it reads well.

Vagabond (approved):

PR #1057, opened 2025-03-28
Fix various things about Kreissack

Nopey:
- He no longer crashes the game when defeated in a non-Nova HAR (possible via custom tournament or other hackery?)
- He now has his overpowered stats from the original game, as reported by rectool when inspecting a match from the original game.
- His nova detonation screenwide pallette tricks work again.
- The winner of a match can no longer attack during STATE_DONE (this was regressed recently by yours truly).

To reduce confusion, damage_add_range has been renamed and no longer marks the `end` color as dirty-- it's an exclusive range, of the same kind that its sister functions vga_palette_tint_range and vga_palette_mix_range take.


src/game/scenes/arena.c
```diff
@@ -700,7 +700,7 @@ void arena_har_defeat_hook(int loser_player_id, scene *scene) {
         local->winner = winner_player_id;
         if(is_singleplayer(gs)) {
             player_winner->sp_wins |= 2 << player_loser->pilot->pilot_id;
-            if(player_loser->pilot->pilot_id == PILOT_KREISSACK) {
+            if(player_loser->pilot->pilot_id == PILOT_KREISSACK && player_loser->pilot->har_id == HAR_NOVA) {
```
Vagabond, line 703: Technically this is behind a singleplayer check, so a hacked kreissack in tournament mode wouldn't trigger this code? I don't object to the additional check, but I don't know if it will actually fix anything a user is likely to experience.

Nopey: i have hacked around enough to be annoyed by it-- maybe i just like kreissack's "i'm not supposed to be on this side of the screen!" messages too much

src/game/objects/har.c
```diff
@@ -1630,14 +1630,14 @@ void har_collide(object *obj_a, object *obj_b) {
 static void process_range(const har *h, damage_tracker *damage, vga_palette *pal, uint8_t step) {
     // For player 0, we should use palette indexes 1 through 47. For player 1, 49 through 95 (skip black).
     // If pe flag is on, we need to switch to handling the other HAR.
-    const int start = 48 * (h->player_id ^ h->p_har_switch) + 1;
-    const int end = start + 47;
+    const int first = 48 * (h->player_id ^ h->p_har_switch) + 1;
+    const int last = first + 46;
```
katajakasa, line 1634: 47

src/video/damage_tracker.h
```diff
@@ -5,14 +5,14 @@
...
+    vga_index dirty_range_first;
+    vga_index dirty_range_last;
     bool dirty;
 } damage_tracker;
 
 void damage_reset(damage_tracker *tracker);
 void damage_copy(damage_tracker *dst, const damage_tracker *src);
 void damage_combine(damage_tracker *dst, const damage_tracker *src);
-void damage_set_range(damage_tracker *tracker, vga_index start, vga_index end);
+void damage_dirty_range(damage_tracker *tracker, vga_index first, vga_index last); // dirties a range
```
katajakasa, line 16: The struct is called damage; there is no special dirty range in damage because all damaged ranges are dirty ranges.

src/game/objects/har.c
Nopey: it's 46, because we're setting one of two ranges:
1 thru 47, or 49 thru 95.

47 - 1 and 95 - 49 are both 46.
I can write it as 47 - 1, if that helps

katajakasa: Both HARs have 48 colors. First color is black, 47 are palette.

src/video/damage_tracker.h
katajakasa: In this, damage is a noun, which we are operating on. I could agree to something like damage_add_range() or somesuch.

Nopey: i'm amenable to damage_add_range, will rename it to that

src/game/objects/har.c
katajakasa: I find these range changes confusing. Previously it was much easier to reason around the ranges -- now we seem to need -1 offset everywhere.

src/video/vga_palette.c
```diff
@@ -8,24 +8,26 @@ void vga_palette_init(vga_palette *palette) {
...
-    for(vga_index i = start; i < end; i++) {
+    vga_index i = first;
+    do {
         m = max3(pal->colors[i].r, pal->colors[i].g, pal->colors[i].b);
         u = (step * m) >> 8;
         pal->colors[i].r += u * (ref.r - pal->colors[i].r) >> 8;
         pal->colors[i].g += u * (ref.g - pal->colors[i].g) >> 8;
         pal->colors[i].b += u * (ref.b - pal->colors[i].b) >> 8;
-    }
+    } while(i++ != last);
```
katajakasa, line 21: Why the do-while loop ? For-loop seems like a clearer option (?)

src/game/objects/har.c
Nopey: Rewrote the fix, so now damage_add_range takes an exclusive range instead of an inclusive range.

src/video/vga_palette.c
Nopey: switched back to exclusive range, with the for loop.

Nopey:
re-recorded the kreissack rectest to fix his broken colors in there and also to highlight that the player defeating him can no longer attack ("Don't har_act during STATE_DONE").

Vagabond:
This needs a simple update to handle that pilot colors are no longer an array.

src/video/damage_tracker.h
```diff
@@ -5,14 +5,21 @@
...
+    vga_index dirty_range_last;
     bool dirty;
 } damage_tracker;
 
 void damage_reset(damage_tracker *tracker);
 void damage_copy(damage_tracker *dst, const damage_tracker *src);
 void damage_combine(damage_tracker *dst, const damage_tracker *src);
-void damage_set_range(damage_tracker *tracker, vga_index start, vga_index end);
+
+// Marks a contigous range as damaged.
```
katajakasa, line 17: For docs, the doxygen syntax we use elsewhere would be nice here, too.

```
/**
 * \brief blah blah
 *
 * \param tracker blah
 * \return blah
 */
 ```

File: src/video/damage_tracker.h [hunk 6]
```diff
@@ -5,14 +5,21 @@
 #include <stdbool.h>
 
 typedef struct damage_tracker {
-    vga_index dirty_range_start;
-    vga_index dirty_range_end;
+    vga_index dirty_range_first;
+    vga_index dirty_range_last;
```
katajakasa, line 9: Are the first and last names still relevant ?

src/video/damage_tracker.h
katajakasa: Btw -- if it seems that hte common pattern is damage_add_range(tracker, start, start + count); -- maybe something like this could/should be added:

```
damage_add_amount(tracker, start, count);
```

Or somesuch

src/video/vga_palette.c
```diff
@@ -21,9 +22,10 @@ void vga_palette_tint_range(vga_palette *pal, vga_index ref_index, vga_index sta
 }
 
 void vga_palette_mix_range(vga_palette *pal, vga_index ref_index, vga_index start, vga_index end, uint8_t step) {
+    assert(end + (vga_index)255 >= start);
     vga_color ref = pal->colors[ref_index];
     uint32_t inv = 255 - step;
-    for(vga_index i = start; i < end; i++) {
+    for(vga_index i = start; i != end; i++) {
```
katajakasa, line 28: I wonder, why the '!=' and not '<' ?

src/video/damage_tracker.h
Nopey: damgae_tracker uses an inclusive range, so I am intentionally renaming its internals to be first/last.

src/video/vga_palette.c
Nopey: Kreissack calls this function with (1, 1 + 255).
1 + 255 is 0, in vga_indexes.

Although we could use a wider integer type for these indexes, i prefer this solution because we don't need to worry about overflowing the pal->colors array.

katajakasa: I see, it's probably my mistake from when I moved these here. Okay, I'm fine with either, though a code comment might clear things for future generations.

src/video/damage_tracker.h
katajakasa: Hmmh, that's an annoying remnant. I've tried to use for-loop friendly ranges everywhere, but this one probably slipped through the net. Okay, let's keep it inclusive, though it might make sense to change to exclusive range (like we do elsewhere in the code) in the future.

src/video/damage_tracker.h
Nopey: Added doxygen comments to everything in damage_tracker.h. I don't think damage_add_amount is going to be an improvement to clarity, and am not currently feeling creative with names.

katajakasa: It's fine, it's not related to this PR anyways.

katajakasa (approved):
LGTM

Vagabond (dismissed):

src/video/vga_palette.c
Nopey: Added a comment about the inequality

katajakasa (dismissed):

PR #1058, opened 2025-03-28
Use the successor_id as the throw range for CLOSE moves

Vagabond:
Also implement throw_range scaling


LegendaryBlueShirt (approved):

src/game/protos/object.c
```diff
@@ -505,6 +505,12 @@ void object_move(object *obj) {
     }
 }
 
+float object_distance(object *a, object *b) {
+    float x_dist = b->pos.x - a->pos.x;
+    float y_dist = b->pos.y - a->pos.y;
+    return sqrtf((x_dist * x_dist) + (y_dist * y_dist));
```
Nopey, line 511:
```suggestion
    return vec2f_dist(a->pos, b->pos);
```
we have a helper for that

Nopey (approved):

PR #1059, opened 2025-03-29
Vendor and static link argtable3

Nopey:
This should fix the Windows CI.


Vagabond:
Looks like we're going to be patching argtable some, but I know @mrannanj was already struggling with some memory issues it had, so vendoring it will give us a chance to fix some of them.

katajakasa:
Off, that pile of errors though. Should we just ignore argtable in linters for now ? Though I agree, the end goal is probably to make it safe.

In any case, I like this much better than trying to find the library and link it.

Vagabond:
> Oooff, that pile of errors though. Should we just ignore argtable in linters for now ? Though I agree, the end goal is probably to make it safe.
> 
> In any case, I like this much better than trying to find the library and link it.

Let's just invest the time in it, we can even make some PRs upstream, because clearly they aren't using all the compiler flags/analysis we are.

katajakasa (approved):
lgtm

Vagabond (approved):

PR #1065, opened 2025-03-29
Implement soft shadows in GLSL

Nopey:
There's a [soft-shadows](https://github.com/omf2097/openomf/tree/soft-shadows) branch which implements this by creating a copy of the surface on the CPU, but I think we'd prefer to do it in shader.

To make the OMF soft shadow effect, we make not one but four samples of the texture as a mask and add them up-- this coverage determines which of the four first remaps we use to make the shadow.

image

### DIFF
--- a/shaders/palette.frag
+++ b/shaders/palette.frag
@@ -17,18 +17,19 @@ uniform sampler2D remaps;
 in vec4 gl_FragCoord;
 
 bool REMAP_SPRITE = (options & 1u) != 0u;
-bool SPRITE_MASK = (options & 2u) != 0u;
+bool SPRITE_SHADOWMASK = (options & 2u) != 0u;
 bool SPRITE_INDEX_ADD = (options & 4u) != 0u;
 bool USE_HAR_QUIRKS = (options & 8u) != 0u;
 
 
 float PHI = 1.61803398874989484820459;
+float ATLAS_H = 2048.0;
 
 float noise(in vec2 v) {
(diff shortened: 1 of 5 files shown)


Nopey:
Somewhat related to #649, but not listed there.

katajakasa:
Nice, this is one thing off my bucket list. I will do a review later when I have a better screen than my mobile phone ;)

And yes, shaders are exactly the place where we want to do this.

Nopey:
image
i've also taken a crack at the weird stasis blending over in PR #1066

shaders/palette.frag
```diff
@@ -42,6 +43,33 @@ vec4 handle(float index, float remap) {
...
+        }
+
+        // don't draw transparent pixels
+        if(coverage == 0) {
+            discard;
+        }
+
+        color = handle(float(coverage) / 256.0);
+        return;
+    }
```
katajakasa, line 70: Hum. On one hand, I like it that this is the graphics backends problem. On the other hand, I wonder what would be the best way if we implement the software renderer at some point.

Nopey: In the other branch, I lazily compute a squashed "shadow" surface which is no more expensive to draw than any other surface.

katajakasa: Let's just merge this one. We can always tinker with it if we need to.

katajakasa (approved):

PR #1066, opened 2025-03-29
video: Add SPRITE_DARK_TINT for Chronos and Shadow effects

Nopey:
SPRITE_DARK_TINT takes the color from the framebuffer behind the sprite we're drawing, remaps it through remap no. 5 (the dark blue one), and then combines that color index with the one from the sprite we're drawing.

This gives the sprite a see-through effect, and is the only thing making STASIS/DARK_TINT visible in the Desert arena (whose sixth remap is really mild, unlike the other arenas).

Comparison (DOS's pixels got squished and stretched a bit, soz)
image

### DIFF
--- a/shaders/palette.frag
+++ b/shaders/palette.frag
@@ -16,6 +16,7 @@ uniform sampler2D remaps;
 
 in vec4 gl_FragCoord;
 
+// for info about these options, see renderer_options in video/enums.h
 bool SPRITE_REMAP = (options & 1u) != 0u;
 bool SPRITE_SHADOWMASK = (options & 2u) != 0u;
 bool SPRITE_INDEX_ADD = (options & 4u) != 0u;
--- a/shaders/palette.frag
+++ b/shaders/palette.frag
@@ -34,9 +34,9 @@ float noise(in vec2 v) {
 
 vec4 handle(float index) {
     if (remap_rounds > 0) {
-        float r_index = remap_offset / 255.0 + index;
-        float r_rounds = remap_rounds / 255.0;
-        return vec4(0.0, r_index, r_rounds, 0.0);
+        float remap = float(remap_offset) / 255.0 + index
+            + float(remap_rounds) * 19.0 / 255.0;
+        return vec4(0.0, remap, 0.0, 0.0);
     }
     if (SPRITE_INDEX_ADD) {
         float add = (index * 255.0 * 60) / 255.0;
--- a/shaders/palette.frag
+++ b/shaders/palette.frag
@@ -13,7 +13,6 @@ flat in uint options;
 
 uniform sampler2D atlas;
 uniform sampler2D remaps;
-uniform sampler2D loopback;
 
 in vec4 gl_FragCoord;
 
@@ -21,7 +20,7 @@ bool SPRITE_REMAP = (options & 1u) != 0u;
 bool SPRITE_SHADOWMASK = (options & 2u) != 0u;
 bool SPRITE_INDEX_ADD = (options & 4u) != 0u;
 bool SPRITE_HAR_QUIRKS = (options & 8u) != 0u;
-bool SPRITE_LOOPBACK = (options & 0x10u) != 0u;
+bool SPRITE_DARK_TINT = (options & 0x10u) != 0u;
 
 
 float PHI = 1.61803398874989484820459;
@@ -33,9 +32,17 @@ float noise(in vec2 v) {
 }
 
 vec4 handle(float index) {
+    if (SPRITE_DARK_TINT) {
+        float remap_index = float(remap_offset) / 255.0;
+        // use magic rounds to detect if dark_tint's remap has been
+        // overwritten by the pause menu's background
+        int magic_remap_rounds = 12;
+        float remap = remap_index + float(magic_remap_rounds) * 19.0 / 255.0;
+        return vec4(0.0, remap, index, 0.0);
+    }
     if (remap_rounds > 0) {
-        float remap = float(remap_offset) / 255.0 + index
-            + float(remap_rounds) * 19.0 / 255.0;
+        float remap_index = float(remap_offset) / 255.0 + index;
+        float remap = remap_index + float(remap_rounds) * 19.0 / 255.0;
         return vec4(0.0, remap, 0.0, 0.0);
     }
     if (SPRITE_INDEX_ADD) {
@@ -89,24 +96,6 @@ void main() {
     }
 
     bool NO_REMAP = SPRITE_HAR_QUIRKS && index > 0x30;
-
-    // chronos's stasis & shadow's shadow transparency
-    if(SPRITE_LOOPBACK) {
-        index = int(texel.r * 255.0);
-        if(index < 0x60) {
-            // grab color index we're drawing ontop of
-            int behind = int(texture(loopback, gl_FragCoord.xy / NATIVE_SIZE).r * 255.0);
-            // lookup behind in remap 5 to get brightness
-            vec4 remap = texture(remaps, vec2(float(behind) / 255.0, 4.0 / 18.0));
-            behind = 1 + clamp(int(remap.r * 255.0) - 0xA8, 0, 7) * 2;
-
-            // this indexed blending is not exactly correct, but i am
-            // done fiddling with it for now-- if someone finds the
-            // code for this in MASTER.DAT, then we can correct it.
-            index = (index & 0xF0) + ((index & 0x0F) * 3 + behind * 2) / 5;
-        }
-        texel.r = float(index) / 255.0;
-    }
 
     // If remapping is on, do it now.
     if (SPRITE_REMAP && !NO_REMAP) {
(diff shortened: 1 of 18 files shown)


Nopey:
Related to issue #649, the renderer wishlist

shaders/palette.frag
```diff
@@ -85,10 +88,28 @@ void main() {
...
+            vec4 remap = texture(remaps, vec2(float(behind) / 255.0, 4.0 / 18.0));
+            behind = 1 + clamp(int(remap.r * 255.0) - 0xA8, 0, 7) * 2;
+
+            // this indexed blending is not exactly correct, but i am
+            // done fiddling with it for now-- if someone finds the
+            // code for this in MASTER.DAT, then we can correct it.
+            index = (index & 0xF0) + ((index & 0x0F) * 3 + behind * 2) / 5;
+        }
+        texel.r = float(index) / 255.0;
+    }
```
katajakasa, line 109: This seems like something that should be done in rgba shader. Then you would have the actual finished background image instead of the possibly half-baked one here ? Also, glCopyTexSubImage2D has a bad tendency to stall.

src/video/renderers/opengl3/gl3_renderer.c
```diff
@@ -223,8 +228,11 @@ static inline void video_set_blend_mode(gl3_context *ctx, object_array_blend_mod
         glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
     } else if(request_mode == MODE_ADD) {
         glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_TRUE);
-    } else {
+    } else if(request_mode == MODE_REMAP) {
         glColorMask(GL_FALSE, GL_TRUE, GL_TRUE, GL_TRUE);
+    } else if(request_mode == MODE_LOOPBACK) {
+        glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
+        texture_update_from_framebuffer(TEX_UNIT_LOOPBACK, ctx->tex_loopback, 0, 0, 0, 0, NATIVE_W, NATIVE_H);
```
katajakasa, line 235: video_set_blend_mode doesn't seem like a correct plate to fiddle with texture copies

shaders/palette.frag
Nopey: rgba shader would need access to both the HAR's `index` and the `index` beneath it.
could be done, but we'd need to rethink our texture channels.

Nopey: To clear up some texture channels:
- Pack remap into one byte `color.g = float(remap_index + remap_count * 19) / 255.0;`, which works as long as remap_count isn't greater than 13.
- Hot-bunk a channel between SPRITE_INDEX_ADD  ('bg' tag is only used in credits) and DARK_TINT (only used in arena).

Brings us down to three channels, in case we need the fourth for something in the future.

katajakasa: I wonder if just throwing in another texture as render target would work. Theoretically possible, but never tested it.

Nopey: Shuffled the data around.
I didn't implement the SPRITE_INDEX_ADD/SPRITE_DARK_TINT hot-bunking idea, but if we need another channel that will still save us one.

Nopey: Stopped deferring SPRITE_INDEX_ADD via the alpha channel-- it's now written directly to the red channel.
Alpha channel's free in case we ever need it

katajakasa: Yep. The additive blending thing was the reason why the whole two-stage pipeline was made. We can't add in the palette stage, because we need to do the other operations first, and because it's not actually additive blending; it's index addition with conditions :P

Nopey: Freed up the SPRITE_INDEX_ADD color channel when not in SCENE_CREDITS, so we're currently up one channel.
woop!

Nopey:
> gl3: Blend REMAPs with MAX to make overlapping shadows look better
If there's sprites where this blendfunc is a problem, we can separate gl3's MODE_SPRITE_SHADOW from MODE_REMAP (and only do the GL_MAX blendfunc in the former).

src/video/enums.h
```diff
@@ -5,15 +5,26 @@ typedef enum
...
+    // use weird stasis/shadow color blending
+    SPRITE_DARK_TINT = 0x10,
 } renderer_options;
 
+typedef enum
+{
+    // Set when displaying the credits.
+    // SPRITE_INDEX_ADD requires this, and is allowed to break
+    // whatever effects aren't used on the credits.
+    FBUFOPT_CREDITS = 0x01,
```
katajakasa, line 25: Where does the "UF" come from ?

src/video/renderers/opengl3/helpers/texture.c
```diff
@@ -20,6 +20,11 @@ void texture_update(GLuint tex_unit, GLuint id, int x, int y, int w, int h, GLen
     glTexSubImage2D(GL_TEXTURE_2D, 0, x, y, w, h, format, GL_UNSIGNED_BYTE, bytes);
 }
 
+void texture_update_from_framebuffer(GLuint tex_unit, GLuint id, int xoffset, int yoffset, int x, int y, int w, int h) {
```
katajakasa, line 23: Consider dropping this if it's not used.

src/video/enums.h
Nopey: Frame BUFfer OPTion

shaders/palette.frag
```diff
@@ -16,24 +16,34 @@ uniform sampler2D remaps;
 
 in vec4 gl_FragCoord;
 
-bool REMAP_SPRITE = (options & 1u) != 0u;
+bool SPRITE_REMAP = (options & 1u) != 0u;
 bool SPRITE_SHADOWMASK = (options & 2u) != 0u;
 bool SPRITE_INDEX_ADD = (options & 4u) != 0u;
-bool USE_HAR_QUIRKS = (options & 8u) != 0u;
+bool SPRITE_HAR_QUIRKS = (options & 8u) != 0u;
+bool SPRITE_DARK_TINT = (options & 0x10u) != 0u;
```
katajakasa, line 23: Could you add some comments on which things these are used for ? Would probably make it easier in the future to reason around the shader code.

Nopey: I'll add a comment directing the reader to video\enums.h (where those comments already live)

src/video/renderers/opengl3/gl3_renderer.c
```diff
@@ -223,8 +226,21 @@ static inline void video_set_blend_mode(gl3_context *ctx, object_array_blend_mod
         glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
     } else if(request_mode == MODE_ADD) {
         glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_TRUE);
-    } else {
+    } else if(request_mode == MODE_REMAP) {
+        glColorMask(GL_FALSE, GL_TRUE, GL_FALSE, GL_FALSE);
+    } else if(request_mode == MODE_SPRITE_SHADOW) {
+        glColorMask(GL_FALSE, GL_TRUE, GL_FALSE, GL_FALSE);
+        glEnable(GL_BLEND);
+        glBlendEquation(GL_MAX);
```
katajakasa, line 234: Hmmh, I guess if there is some other remapping effect underneath, this can cause some interesting corruption ? This sort of feels like class-A footgun along with the round+index combining.

I can't really think of anything that would be there in arenas, that would also be picked by GL_MAX blend function though :/

Nopey: MODE_SPRITE_SHADOW is only used on the SPRITE_SHADOWs, so we'd need something to be drawn before and underneath the character's shadows.
GL_MAX would then pick whichever remap has a greater remap index-- since shadows use the first four remaps, the shadows would not overwrite whatever remap was already drawn into the buffer.

katajakasa: Hum. Did the bullets hitting the ground in desert or lightning flashes in the power plant do any remapping ? I can't remember what other effects we had on the arenas. Also, the arena menu help box probably overlaps.

Nopey: Those are remapping effects, and will overwrite the shadows because they are drawn late (in MODE_REMAP, which does not do blending.)

Nopey: i thought the lightning flashes were a palette transformer, not a remap

Nopey: arena menu help box will be drawn after the shadows, and so what blend func the shadows use will not effect it.

katajakasa: I think we are safe, then. Let's just go with this -- there seems to be enough comments in the code to be able to figure out any bugs if any show up later.

katajakasa (approved):

PR #1067, opened 2025-03-29
Implement pilot stats cheat from the original

Nopey:
Players can enable the pilot stats cheat by moving their cursor across all pilots and wrapping on both rows, then pressing KICK.
While holding KICK, the stats can be reallocated with the directional input.

### DIFF
--- a/src/game/scenes/melee.c
+++ b/src/game/scenes/melee.c
@@ -96,7 +96,7 @@ typedef struct {
     text *player_bio[2];
 
     // nova selection cheat
-    unsigned char har_selected[2][10];
+    unsigned char cheat_selected[2][10];
     unsigned char katana_down_count[2];
 } melee_local;
 
@@ -348,9 +348,10 @@ void handle_action(scene *scene, int player, int action) {
                     local->pilot_id_a = CURSOR_INDEX(local, 0);
                     local->pilot_id_b = CURSOR_INDEX(local, 1);
 
-                    // nova selection cheat
-                    local->har_selected[0][local->pilot_id_a] = 1;
-                    local->har_selected[1][local->pilot_id_b] = 1;
+                    // prepare for nova selection cheat
+                    memset(local->cheat_selected, 0, sizeof(local->cheat_selected));
+                    local->cheat_selected[0][local->pilot_id_a] = 1;
+                    local->cheat_selected[1][local->pilot_id_b] = 1;
 
                     load_pilot(scene, 0);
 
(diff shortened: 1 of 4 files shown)


src/game/scenes/melee.c
```diff
@@ -308,7 +400,58 @@ void handle_action(scene *scene, int player, int action) {
...
+                }
+                refresh_pilot_stats(scene, player);
+            }
+            break;
         case ACT_KICK:
+            if(local->page == PILOT_SELECT && check_pilot_stat_cheat(scene, player)) {
+                progressbar_set_highlight(local->bar_stat[player][local->cheat_pilot_stats_stat[player]], 1);
+                break;
+            }
+            // [[fallthrough]]
```
Vagabond, line 454: GCC doesn't seem to like this

Vagabond, line 454:
```suggestion
            // [[fallthrough]]
            __attribute__ ((fallthrough));
```

katajakasa: huh, I didn't know that one existed

Nopey: that is a nonstandard extension, not supported by MSVC.

Vagabond: Gotta find something that keeps all 3 happy, somehow

Nopey: CI passed. Do you mean to say, GCC doesn't like if you uncomment the `[[fallthrough]]`?
We're not setting the C standard to C23, so GCC should reject it.

katajakasa: C23 would be nice, it arguably makes C clearer. Probably not should wait a couple of years to wait for compiler support though.

Vagabond: I was unable to compile this branch locally with GCC 13.2.0

Nopey: OMF_FALLTHROUGH macro that expands to one of them?
I didn't think it was necessary to explicitly mark it here, but left the comment for us to find when we later support C23.

Nopey: GCC 13 defaults to `-std=gnu23`, but we do explicitly set CMAKE_C_STANDARD to 11 in CMakeLists.
Could you post GCC's error message, and check that -std=c11 or -std=gnu11 is present?

Nopey: After consulting the manual and godbolt compiler explorer, I think this should now compile with your GCC 13.2.0

Vagabond: Confirmed.

Vagabond (approved):

PR #1068, opened 2025-03-29
Implement 'md' tag for HARs to delete an existing projectile

Vagabond:
Projectile must be owned by the HAR and have the corresponding animation ID. The md tag is only used in the shadow grab in a HAR context.


katajakasa (approved):
Seems to work.

PR #1069, opened 2025-03-29
Rewrote match_move function

LegendaryBlueShirt:
Removed a lot of duplicate/special cases with regards to Scrap/Destruction handling

Implemented JJ and JG tags, and you should be able to throw and perform special moves during the getup animation.  Shredder flip kick and Katana wall spin now also cancel into air moves.

Broke air landing...

It was unclear to me what `enqueued` was meant to do so I would appreciate some explanation there.


src/game/objects/har.c
```diff
@@ -1909,123 +1909,104 @@ static bool add_input(char *buf, int act_type, int direction) {
...
+        switch(move->category) {
+            case CAT_JUMPING:
+                // JZ should be checked in a bunch of other places but since it's only used for one move we'll cheat.
+                if(player_frame_isset(obj, "jz") || player_frame_isset(obj, "jj") ||
+                 (is_har_idle_air(obj) && allowed_in_idle && !h->air_attacked)) {
+                    allowed = true;
+                }
+                break;
+            case CAT_CLOSE:
+                if(player_frame_isset(obj, "jg") || (is_har_idle_grounded(obj) && allowed_in_idle)) {
```
Vagabond, line 1953: Won't this break the overhead throw?

LegendaryBlueShirt: match_move is only interested in checking whether you are allowed to use move commands, not automatic transitions.
The overhead throw startup (ID 39) is of type CAT_JUMPING and thus unaffected

rectests/SHARDS-ARE-BLOCKABLE.REC
Nopey, line 1: re-recorded SHARDS-ARE-BLOCKABLE.REC has backwards pilot colors and reduces test coverage.

image
(left is new, right is old)

Please edit the existing rectest using rectool.
I expect the assertions on moves 14 through 16 will need to at least have their ticks adjusted, and the assertion on tick 17 is buggy and so should probably be deleted outright until a followup PR makes blocked shards do their proper damage.

Some notes from playing back the old version of the rec in this PR's openomf:
- Shirro enters STATE_BLOCKSTUN for the first time on tick 220.
- Shirro returns to STATE_WALKFROM inbetween blocks, which is correct behavior.
- Shirro returns to anim 10 one tick before returning to STATE_WALKFROM, which may be considered a bug.
- Shirro does not restart anim 5 (block) from the beginning when blocking an additional shard, which I believe is a bug.
- Shirro does not take damage from blocking shards, which is a bug.
- Crystal is knocked back by the first block, but none of the subsequent. This is a bug.

I imagine most of these bugs are preexisting, but haven't which of them `master` suffers from.

Nopey (changes requested):

rectests/SHARDS-ARE-BLOCKABLE.REC
LegendaryBlueShirt: I did use rectool, though it might be out of date

Nopey: I'll double check that latest rectool isn't switching color order, and I trust this PR is based on latest master and so has all the color fixes from earlier today.

Vagabond (approved):
Overall this looks reasonable and seems to play well

rectests/SHARDS-ARE-BLOCKABLE.REC
Nopey: Backwards color palette is annoying, but is likely just an outdated rectool.
We'll deal with the above bugs and loose threads from PR #1051 in another PR, I suppose

Nopey (approved):

PR #1070, opened 2025-03-30
Properly load Shirro for second player in 2-player scene

Nopey:
Regressed in my recent pr, #1067.
Was accidentally loading a bunch of Crystal's data instead of Shirro when creating the melee scene.

### DIFF
--- a/src/game/scenes/melee.c
+++ b/src/game/scenes/melee.c
@@ -874,6 +874,7 @@ int melee_create(scene *scene) {
 
     local->cursor[1].row = 0;
     local->cursor[1].column = 4;
+    local->pilot_id_b = CURSOR_INDEX(local, 1);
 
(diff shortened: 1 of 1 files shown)


src/game/scenes/melee.c
```diff
@@ -874,6 +874,7 @@ int melee_create(scene *scene) {
 
     local->cursor[1].row = 0;
     local->cursor[1].column = 4;
+    local->pilot_id_b = CURSOR_INDEX(local, 1);
```
Vagabond, line 877: So we just rely on pilot_id_a being unitialized?

Nopey: omf_calloc zeroes it.

Vagabond: Right, that's what I meant.

Vagabond (approved):

PR #1071, opened 2025-03-30
Kill shadow projections when the originating HAR takes a hit

Vagabond:


LegendaryBlueShirt:
The Shadow clones should trigger their disappearing animation, not vanish instantly.

LegendaryBlueShirt:
It looks like this behavior is actually the result of shadow clones behaving as if they were hit themselves, so other than the fact that they are still projectiles they need to enter some kind of getting hit animation

LegendaryBlueShirt (approved):
Working much better!

PR #1073, opened 2025-03-30
Only use int_tick inside of net_controller

Nopey:
gs->int_tick is a nice monotonically increasing tick counter for netplay.

It is not a counter of dynamic ticks.

Use gs->tick instead

### DIFF
--- a/src/utils/log.c
+++ b/src/utils/log.c
@@ -140,7 +140,11 @@ void log_msg(log_level level, const char *fmt, ...) {
         va_start(args, fmt);
         vfprintf(target->fp, fmt, args);
         va_end(args);
-        fprintf(target->fp, "\n");
+        if(state->colors && target->colors) {
+            fprintf(target->fp, "\x1b[0m\n");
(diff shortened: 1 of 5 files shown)


src/utils/log.c
```diff
@@ -140,7 +140,11 @@ void log_msg(log_level level, const char *fmt, ...) {
         va_start(args, fmt);
         vfprintf(target->fp, fmt, args);
         va_end(args);
-        fprintf(target->fp, "\n");
+        if(state->colors && target->colors) {
```
Vagabond, line 143: Unrelated change?

Vagabond (approved):
I'm going to trust you've tested this and are confident in the change, but the code changes look reasonable.

src/utils/log.c
Nopey: Yeah, this unrelated change just clears the colors after when we call log_debug with a color.
It's in a seperate commit so things don't get confused

PR #1075, opened 2025-03-30
Return to HAR select from VS and support the arcade altpals cheat

Nopey:
To access the altpals cheat, press a number key (not numpad) 1 through 3 to change HAR 1 colors on the HAR select.
For the second player in 2-player mode, use keys 4 through 6.

Returning to the melee scene from VS we now place the player at the HAR select just like the original.

### DIFF
--- a/src/game/scenes/melee.c
+++ b/src/game/scenes/melee.c
@@ -867,10 +867,31 @@ static void load_hars(scene *scene, melee_local *local, bool player2_is_selectab
     }
 }
 
+int melee_event_cb(scene *scene, SDL_Event *e) {
+    melee_local *local = scene_get_userdata(scene);
+    if(local->page == HAR_SELECT && e->type == SDL_KEYDOWN && SDLK_1 <= e->key.keysym.sym &&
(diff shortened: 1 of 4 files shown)


src/game/scenes/melee.c
```diff
@@ -848,44 +862,67 @@ static void load_har_portraits(scene *scene, melee_local *local) {
...
-
-    local->cursor[1].row = 0;
-    local->cursor[1].column = 4;
-    local->pilot_id_b = CURSOR_INDEX(local, 1);
+    scene->event = melee_event_cb;
 
     game_player *player1 = game_state_get_player(scene->gs, 0);
     game_player *player2 = game_state_get_player(scene->gs, 1);
 
+    // if we already have a pilot name, we're coming back from VS.
```
Vagabond, line 905: in 2 player mode, we also go back to VS, but this currently doesn't reset the pilots like the original does.

Vagabond: I pushed a fix to make this happen.

Vagabond (approved):

PR #1076, opened 2025-03-30
Make sure we're using the latest config when starting a fight

Vagabond:
This fixes the issue where changing the round count, etc might not happen immediately.


Nopey (dismissed):

LegendaryBlueShirt (approved):
Works on my machine

LegendaryBlueShirt (approved):

PR #1078, opened 2025-03-30
Scoreboard fixes

katajakasa:
Use the new text renderer in scoreboard + more correctness fixes.

### DIFF
--- a/src/console/console_cmd.c
+++ b/src/console/console_cmd.c
@@ -350,6 +350,17 @@ int console_cmd_rank(game_state *gs, int argc, char **argv) {
     return 1;
 }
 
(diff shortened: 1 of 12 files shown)


.clang-format
```diff
@@ -19,3 +19,24 @@ AllowShortBlocksOnASingleLine: 'Never'
 AlignArrayOfStructures: Left
 IndentCaseLabels: true
 AllowShortEnumsOnASingleLine: false
+---
+Language:  C
```
Vagabond, line 23: This seems unrelated?

katajakasa: Right, I need to use that to be able to run clang-format. Accidentally committed it.

Vagabond (approved):

PR #1081, opened 2025-03-30
Console text fixes

katajakasa:
Use new text renderer in console. Also other fixes.

### DIFF
--- a/src/console/console.c
+++ b/src/console/console.c
@@ -26,7 +26,7 @@ console *con = NULL;
 // defined in console_cmd.c
 void console_init_cmd(void);
 
-int make_argv(char *p, char **argv) {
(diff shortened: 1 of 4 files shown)


LegendaryBlueShirt (changes requested):
I get an assert when I use the help command and then press up in the console.
image

katajakasa:
> I get an assert when I use the help command and then press up in the console. image

Cannot repeat on linux/gdb. Can you get a stacktrace ?

LegendaryBlueShirt:
From the debug window -
`Assertion failed: false && "Encountered unprintable character!", file C:\openomfgithub\openomf\src\game\gui\text\text_layout.c, line 53`

LegendaryBlueShirt (approved):

PR #1082, opened 2025-03-31
Throw Conditions Fix

LegendaryBlueShirt:
Can no longer throw enemies who are in the air, being hit, or blocking

Can no longer throw enemies out of range.  CAT_CLOSE moves won't come out, and other throws will come out but won't hit.


src/game/objects/har.c
```diff
@@ -350,13 +350,33 @@ int har_is_blocking(har *h, af_move *move) {
...
 int har_is_invincible(object *obj, af_move *move) {
     if(player_frame_isset(obj, "zz")) {
         // blocks everything
         return 1;
     }
+
     switch(move->category) {
-        // XX 'zg' is not handled here, but the game doesn't use it...
+        case CAT_CLOSE:
+            if(player_frame_isset(obj, "zg") || obj->cur_animation->id == ANIM_DAMAGE ||
```
Vagabond, line 375: I thought throws were unblockable?

Vagabond:
This looks ok other than the is_invincible changes that check for obj->cur_animation->id == ANIM_STANDING_BLOCK || obj->cur_animation->id == ANIM_CROUCHING_BLOCK

Vagabond:
Oh, I see now.

Vagabond (approved):

src/game/objects/har.c
LegendaryBlueShirt: You can't combo into a regular throw.

PR #1083, opened 2025-03-31
Improve Gargoyle Flight

LegendaryBlueShirt:
The H tag is now enforced more strictly which gives Gargoyle much needed height on 236P


Vagabond:
Generally it's better to reorder the rectest commit to be before the fix, so it's easier to test the rectest actually shows a bug before the fix. (you can easily reorder the commits with `git rebase -i origin/master` and just reorder the 2 commits in the editor).

Vagabond (approved):

PR #1084, opened 2025-03-31
Fix Gargoyle Dive

LegendaryBlueShirt:
The opponent's position after Gargoyle dive was wrong because it's flipped due to improper ordering of tags.


Vagabond:
A rectest would be good here, but it does appear to fix the bug.

Vagabond (approved):

Vagabond (approved):

PR #1087, opened 2025-03-31
console: clean up more remnants

katajakasa:
More fixes.

### DIFF
--- a/src/console/console.c
+++ b/src/console/console.c
@@ -51,16 +51,14 @@ static int make_argv(char *p, char **argv) {
 }
 
 static void console_refresh(void) {
-    char row[128] = {0};
-    int x = 0;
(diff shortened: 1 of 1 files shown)


LegendaryBlueShirt (approved):
Didn't break any of my limited console usage

src/console/console.c
```diff
@@ -51,16 +51,14 @@ static int make_argv(char *p, char **argv) {
 }
 
 static void console_refresh(void) {
-    char row[128] = {0};
-    int x = 0;
+    char row[2] = {0, 0};
```
mrannanj, line 54: Now, x is gone, good, but is `row[1]` used somewhere now?

mrannanj: Okay, heard it was a null terminator.

PR #1096, opened 2025-04-01
Implement stadium lighting effect

Nopey:
TODO: Troubleshoot why the HARs get brighter in the corners. Perhaps the original game does two passes of palette transforming in that case, one to darken it?


LegendaryBlueShirt:
The X and Y axes appear to work inverse, xdist goes from 160 to 0 based on your distance from the center whereas ydist goes from 0 to whatever your height is

src/game/objects/har.c
```diff
@@ -1721,8 +1721,27 @@ static void process_range(const har *h, damage_tracker *damage, vga_palette *pal
...
+
+    if(object_has_effect(obj, EFFECT_POSITIONAL_LIGHTING)) {
+        int const mid = NATIVE_W / 2;
+        int xDist = mid - abs(mid - object_px(obj));
+        int yDist = ARENA_FLOOR - object_py(obj);
+        int iBlendFact = xDist / 2 - yDist / 2 - 30;
+
+        iBlendFact = max2(iBlendFact, -41);
+        uint8_t iGray = iBlendFact < 1 ? 1 : 53;
+        iBlendFact = abs(iBlendFact);
```
katajakasa, line 1736: camelCase in otherwise underscore application :)

src/video/vga_palette.c
```diff
@@ -34,3 +34,15 @@ void vga_palette_mix_range(vga_palette *pal, vga_index ref_index, vga_index star
         pal->colors[i].b = (pal->colors[i].b * inv + ref.b * step) >> 8;
     }
 }
+
+void vga_palette_light_range(vga_palette *pal, uint8_t iGray, int iBlendFact, vga_index start, vga_index end) {
+    assert(end + (vga_index)255 >= start);
+    int targetGray = iGray * 4; // original game uses range of 0 through 63 for colors for VGA reasons
```
katajakasa, line 40: Eh, just mention the 6bit palette directly.

Nopey:
addressed review feedback

src/game/protos/object.c
```diff
@@ -411,11 +411,6 @@ void object_render(object *obj) {
         remap_rounds = 0;
         remap_offset = 5;
         options |= SPRITE_REMAP | SPRITE_DARK_TINT;
-    } else if(object_has_effect(obj, EFFECT_POSITIONAL_LIGHTING)) {
-        int rx = x + (w >> 1);
-        remap_rounds = 0;
```
katajakasa, line 416: There were some signs in the original game engine that the stadium lighting does in fact use the remaps, which is why we used it here. We didn't have the time to figure out why the effect was too dark. One guess was that we would need to run the remapping multiple times.

katajakasa: ^ Agreed with the above.

LegendaryBlueShirt (approved):
Works!

PR #1097, opened 2025-04-01
Crossfade inbetween arena rounds, improve round start timing accuracy

Nopey:


LegendaryBlueShirt (approved):
Testing this, seems fine, a macro to move all the rectest ticks up would be helpful here though 😂

Vagabond:
We can probably write a quick rectest patch to move all assert ticks forward by X, if needed.

Nopey:
> a macro to move all the rectest ticks up would be helpful here though

Added `--fixup pr1097` option to rectool, in case any RECs need such fixing up.

src/game/scenes/arena.c
```diff
@@ -1210,6 +1198,45 @@ void push_players(scene *scene, game_player *p1, game_player *p2) {
...
+    int const target_tick = local->state == ARENA_STATE_STARTING ? 0 : (80 + ARENA_CROSSFADE_TICKS);
+    int progress = abs(local->state_ticks - target_tick);
+    float fraction = progress / (float)ARENA_CROSSFADE_TICKS;
+
+    // Set palette darkness value.
+    for(int i = 0; i < 256; i++) {
+        pal->colors[i].r *= fraction;
+        pal->colors[i].g *= fraction;
+        pal->colors[i].b *= fraction;
+    }
```
katajakasa: maybe vga_palette_darken() for consistency with other palette tricks

tools/rectool/main.c
```diff
@@ -527,6 +529,21 @@ int main(int argc, char *argv[]) {
         print_rec_root_info(&rec);
     }
 
+    if(fixup->count == 1) {
```
katajakasa: Could you explain here shortly why this is needed, so that one doesn't need to fire up the browser :)

Nopey: why the fixup is needed? or, the check for fixup->count?
unsure what you're asking for

katajakasa: Why the fixup is needed.

src/game/scenes/arena.c
```diff
@@ -432,9 +414,47 @@ static void arena_end(scene *sc) {
     }
 }
 
+static void arena_create_roundstart_anim(scene *scene) {
```
katajakasa, line 417: I like this added DRYness.

PR #1098, opened 2025-04-01
Generate ESC from keyboard again, and handle it in har.c

Vagabond:


katajakasa (approved):

Vagabond:
We decided on #1099 instead.

PR #1101, opened 2025-04-01
Fix the spawn/destroy/disable callbacks to be netplay safe

Vagabond:


Vagabond:
Pack the object ID into the userdata pointer, instead of using a rollback unsafe pointer.

Nopey (approved):
looks fine, but do keep this hack off of the master branch-- we shouldn't be casting object id's to pointers and back 😅

Vagabond:
> looks fine, but do keep this hack off of the master branch-- we shouldn't be casting object id's to pointers and back 😅

Yeah, the tooling seems to hate it. I guess I can fix it another way.

katajakasa (approved):

PR #1108, opened 2025-04-02
Calculate version in CMake, support a shallow checkout

Nopey:
Advantages include:
- our Actions' clones are shallow again, and version.cmake will run a `fetch --shallow-since` as needed.
- ~~releases which fork off of the main branch (e.g. 0.8.0) are now handled elegantly-- in such cases, `LATEST_RELEASE_FORKED_FROM` should be set to the commit where the release was forked.~~
- VERSION_LABEL's commit hash is no longer prefixed with a `g`.
- Errors in the version calculation do not block the build-- instead, the version calculation error is shoved in the version's label (e.g.  `0.0.0-GITNOTFOUND`)
- builds that are after a prerelease, but before the stable release no longer get their PATCH version bumped (e.g. 3 commits after 0.8.1-rc1 no longer gets labelled as 0.8.2-rc1-2-COMMITHASH)

### DIFF
--- a/cmake-scripts/version.cmake
+++ b/cmake-scripts/version.cmake
@@ -1,0 +1,256 @@
+#
+# version.cmake has the job of setting five variables:
+# - SHA1_HASH, VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH, AND VERSION_LABEL
+#
+# It calculates these by running git commands.
+#
+# Although it is used in CMakeLists.txt, version.cmake is also run from github actions
+# as a standalone cmake script, so that our actions have access to the version information as well.
+#
+
+set(GIT_REMOTE origin)
+
+# set version in case we error
+set(VERSION_MAJOR 0)
+set(VERSION_MINOR 0)
+set(VERSION_PATCH 0)
+set(VERSION_LABEL)
+
+# file-internal macro for things we want to do before finishing this script,
+# whether we're on the success path or error path.
+macro(version_finally)
+    if(DEFINED SHA1_HASH_SHORT)
+        string(APPEND VERSION_LABEL "-${SHA1_HASH_SHORT}")
+    endif()
+
+    if(DEFINED CMAKE_SCRIPT_MODE_FILE)
+        # output exactly what the github actions wants
+        message("${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}${VERSION_LABEL}")
+    else()
+        message(STATUS "Git SHA1 Hash: ${SHA1_HASH}")
+        message(STATUS "OpenOMF Version: ${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}${VERSION_LABEL}")
+    endif()
+endmacro()
+
+# file-internal macro for reporting a fatal error.
+#
+# Although the error is fatal to version.cmake, it is only
+# raised as a cmake fatal error if VERSION_REQUIRED is set.
+# This lets the project be built even if version detection fails.
+macro(version_error error)
+    # put error into version label
+    string(APPEND VERSION_LABEL "-${error}")
+
+    if(NOT DEFINED SHA1_HASH)
+        set(SHA1_HASH "${error}")
+    endif()
+
+    version_finally()
+
+    # Enable VERSION_REQUIRED to treat any errors as fatal.
+    if(VERSION_REQUIRED)
+        message(FATAL_ERROR "Version did not succeed, ${error}.")
+    endif()
+
+    # exit the version.cmake script, we're done.
+    return()
+endmacro()
+
+
+if(DEFINED CMAKE_SCRIPT_MODE_FILE)
+    # suppress find_package's status message in script mode
+    set(GIT_FIND_QUIETLY QUIET)
+endif()
+
+find_package(Git ${GIT_FIND_QUIETLY})
+if(NOT Git_FOUND)
+    message(WARNING "Git not found")
+    version_error("GITNOTFOUND")
+endif()
+
+# Set SHA1_HASH
+execute_process(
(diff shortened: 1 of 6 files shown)


cmake-scripts/version.cmake
```diff
@@ -0,0 +1,225 @@
+# Calculate OMF_ROOT, the path to the root of the repo.
+if(DEFINED CMAKE_SCRIPT_MODE_FILE)
+    set(OMF_ROOT "${CMAKE_SCRIPT_MODE_FILE}")
+    get_filename_component(OMF_ROOT "${OMF_ROOT}" DIRECTORY)
+    get_filename_component(OMF_ROOT "${OMF_ROOT}" DIRECTORY)
```
mrannanj, line 5: Duplicate line here, probably not deliberate?

mrannanj (changes requested):

cmake-scripts/version.cmake
Nopey: That is deliberate. One is to remove the `version.cmake` from the path, the second removes `cmake-scripts/` from the path.

cmake-scripts/version.cmake
```diff
@@ -0,0 +1,225 @@
+# Calculate OMF_ROOT, the path to the root of the repo.
```
katajakasa, line 1: An explanation on what this file does and why in the header would be helpful.

katajakasa: OMF_ROOT => OMF_REPOSITORY_ROOT ?

cmake-scripts/version.cmake
```diff
@@ -0,0 +1,225 @@
...
+    if(DEFINED CMAKE_SCRIPT_MODE_FILE)
+        # output exactly what the github actions wants
+        message("${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}${VERSION_LABEL}")
+    else()
+        message(STATUS "Git SHA1 Hash: ${SHA1_HASH}")
+        message(STATUS "OpenOMF Version: ${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}${VERSION_LABEL}")
+    endif()
+endmacro()
+
+macro(version_error error)
```
katajakasa, line 31: Code comments on what this error is and why we add it

cmake-scripts/LatestRelease.cmake
```diff
@@ -0,0 +1,7 @@
+# This file tracks the latest release info
+
+SET(LATEST_RELEASE_TAG "0.8.0")
+
+# If set, LATEST_RELEASE_FORKED_FROM tells of the last main branch's commit
+# that is in the latest release.
```
katajakasa, line 6: What do we do with this ? How does it affect things ? Code comment.

cmake-scripts/version.cmake
```diff
@@ -0,0 +1,225 @@
...
+            return()
+        endif()
+    endif()
+
+    set(VERSION_COUNTFROM "${LATEST_RELEASE_FORKED_FROM}")
+else()
+    set(VERSION_COUNTFROM "${LATEST_RELEASE_TAG}")
+endif()
+
+# count commits
```
katajakasa, line 164: Why ? Code comment.

katajakasa:
I don't know Cmake well enough to be able to read the new code properly. So what I would appreciate here most of all is code comments on /why/ something happens. The "what" is nice too, but it often doesn't help all that much.

Vagabond:
I'd really prefer if we didn't have versions and commits tracked in a file somewhere we have to edit. It was *very* nice to just do a tag for 0.8.0 and not have to remember to edit several other files.

If we only have to edit the tag/commit when we make a feature branch, that is probably acceptable, but I want to try to minimize the scope for missing things we need to do when updating. Also please update RELEASE.md with instructions, if applicable.

Nopey:
Added instructions to RELEASING.MD (sic), and a big comment at the top of version.cmake that explains what it's doing and what the parts of our versioon (major/minor/patch/label) even mean.

cmake-scripts/version.cmake
```diff
@@ -0,0 +1,298 @@
...
+#
+# A NOTE ON VERSION LABELS.
+# as a technical note, -C-HHHHHHHH together form the "version label".
+# any version label (e.g. "-rc1") from the latest release's tag is also
+# carried forward, and so a version can look like 0.8.1-rc1-29-01dcec0a.
+#
+#
+# COUNTING COMMITS.
+# We use git to count the commits. There is one twist, because of versions like
+# 0.8.0 (where commits were applied to the release, but not master--)
```
Vagabond, line 39: So, to be clear, we can and should merge the 0.8.0 branch back into master, which should solve this problem? Release branches should always be a subset of master, I think?

katajakasa: Yes, release branch is supposed to be merged to master.

Nopey: After 0.8.0 is merged into master, I'll simplify this PR by removing LATEST_RELEASE_FORKED_FROM and (if my next idea on how to abuse git fetches works out) removing LatestRelease.cmake

Vagabond:
> Drafting until 0.8.0 is merged into master.

I believe @katajakasa did this yesterday

Vagabond:
Is there a reason we abandoned the old git-describe approach? This seems to replicate a bunch of that functionality but by doing it manually? Is it to support the `builds that are after a prerelease, but before the stable release no longer get their PATCH version bumped ` feature?

Nopey:
The previous approach does not support shallow checkouts.

`git describe` only works when you have the entire history between the HEAD commit and the previous release already fetched.

Vagabond:
Propose we merge this after we tag 0.8.1 so we can let it soak for a bit.

katajakasa (approved):

PR #1109, opened 2025-04-02
Various Mechlab/Tournament UI Fixes

### LINKED ISSUE #1106: [GAMEPLAY] Running out of money on tournament leads to a forever disabled pilot
**Steps to reproduce**
What happened?
I lost too much money in tournament, Plug warned me that I had one final chance to win, but failed that, and then I got kicked out of the tournament and my money was set to $0 K. With this pilot, the only choices I have is to load another pilot or create a new one.

What did you expect to happen?
I expected there to be some way to recover from this financial hardship I entered, and continue with this pilot into another tournament.

Version was v0.8.0.

Image

Nopey:
Fixes #1106

### DIFF
--- a/src/formats/chr.h
+++ b/src/formats/chr.h
@@ -40,8 +40,8 @@ typedef struct {
     sd_sprite *photo;                       ///< Pilot photo
     float winnings_multiplier;              ///< Tournament winnings multiplier
     sd_chr_enemy *enemies[MAX_CHR_ENEMIES]; ///< List of enemy states in current tournament
-    char bk_name[14];                       //< cutscene bk for end of tournament
-    int tournament_id;                      //< misnomer? set to 4 for WORLD.BK cutscene quirks
+    char bk_name[14];                       ///< cutscene bk for end of tournament
+    int tournament_id;                      ///< set to 4 for WORLD.BK cutscene quirks, also known as TOURNAMENT NUMBER
     char *cutscene_text[10];                ///< cutscene dialog text
 } sd_chr_file;
 
--- a/src/formats/chr.h
+++ b/src/formats/chr.h
@@ -41,6 +41,7 @@ typedef struct {
     float winnings_multiplier;              ///< Tournament winnings multiplier
     sd_chr_enemy *enemies[MAX_CHR_ENEMIES]; ///< List of enemy states in current tournament
     unsigned int cutscene;                  ///< cutscene id for end of tournament
+    int tournament_id;                      //< misnomer? set to 4 for WORLD.BK cutscene quirks
     char *cutscene_text[10];                ///< cutscene dialog text
 } sd_chr_file;
 
--- a/src/formats/chr.h
+++ b/src/formats/chr.h
@@ -40,7 +40,7 @@ typedef struct {
     sd_sprite *photo;                       ///< Pilot photo
     float winnings_multiplier;              ///< Tournament winnings multiplier
     sd_chr_enemy *enemies[MAX_CHR_ENEMIES]; ///< List of enemy states in current tournament
-    unsigned int cutscene;                  ///< cutscene id for end of tournament
+    char bk_name[14];                       //< cutscene bk for end of tournament
     int tournament_id;                      //< misnomer? set to 4 for WORLD.BK cutscene quirks
     char *cutscene_text[10];                ///< cutscene dialog text
 } sd_chr_file;
--- a/src/game/scenes/mechlab/lab_menu_main.c
+++ b/src/game/scenes/mechlab/lab_menu_main.c
@@ -141,17 +141,17 @@ enum lab_buttons
 
 // clang-format off
 static const button_details details_list[] = {
-    // CB, Text, Text align, Halign, Valigh, Pad top, Pad bottom, Pad left, Pad right
-    {lab_menu_main_arena,          "ARENA",            TEXT_ROW_HORIZONTAL, TEXT_ALIGN_CENTER, TEXT_ALIGN_TOP,    {0, 0, 2,  0}},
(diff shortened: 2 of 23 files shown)


katajakasa (approved):
Tested, works, code seems fine. LGTM.

src/game/scenes/mechlab/lab_menu_main.c
```diff
@@ -252,7 +266,7 @@ static const spritebutton_tick_cb tick_cbs[] = {
     lab_menu_tick_in_tournament, // lab_menu_tick_delete,
     lab_menu_tick_in_tournament, // lab_menu_tick_sim,
     NULL,                        // lab_menu_tick_quit,
-    lab_menu_tick_in_tournament, // lab_menu_tick_tournament,
```
Vagabond, line 255: This should be disabled when there's not a player loaded, at least.

Nopey: The loop in lab_menu_main_create handles it.

that loop now also now sets supports_select, so the buttons the loop disables can't be selected.

Vagabond:
Erasing the "last" pilot name from my openomf config, booting the game and then loading a character gives me a greyed out new tournament button.

Vagabond:
Leaving and re-entering tournament mode seems to fix it.

Vagabond:
I set my money to 0 on a fresh character, and lost to cosette twice. I got kicked out of the tournment and was able to rejoin with 0 credits (my credits were negative before). I don't remember what's supposed to happen, though.

src/resources/trnmanager.c
```diff
@@ -9,6 +9,12 @@
 #include <stdio.h>
 #include <string.h>
 
+static int trn_sort_compare_fn(const void *a, const void *b) {
+    sd_tournament_file const *trn_a = a;
+    sd_tournament_file const *trn_b = b;
+    return (trn_a->registration_fee > trn_b->registration_fee) - (trn_a->registration_fee < trn_b->registration_fee);
```
mrannanj, line 15: What's the problem with just `return trn_a->registration_fee - trn_b->registration_fee;` here?

Nopey: i don't want to overflow or underflow an integer.

Nopey:
> I set my money to 0 on a fresh character, and lost to cosette twice. I got kicked out of the tournment and was able to rejoin with 0 credits (my credits were negative before). I don't remember what's supposed to happen, though.

We're missing the "you have to work as a mechanic" and "while working as a mechanic your HAR upgrades / training courses rotted" behaviors.

The credits going negative (after the warning) and then zero (after being kicked out) is correct.

Nopey:
pushed more fixes:
- rewrote hardcoded tournament filename behaviors (the original four tournaments are not special-cased, and are not even referenced anywhere in MASTER.DAT).
- `scene` cheat now works with scene names (previously, the strings would never have matching case)
- implemented tournament cutscene replay cheat 'e'

src/formats/chr.h
```diff
@@ -40,7 +40,7 @@ typedef struct {
     sd_sprite *photo;                       ///< Pilot photo
     float winnings_multiplier;              ///< Tournament winnings multiplier
     sd_chr_enemy *enemies[MAX_CHR_ENEMIES]; ///< List of enemy states in current tournament
-    unsigned int cutscene;                  ///< cutscene id for end of tournament
+    char bk_name[14];                       //< cutscene bk for end of tournament
```
mrannanj, line 43: Well, guess use 3 slashes on all since the others have it too

mrannanj (changes requested):

src/formats/chr.h
katajakasa: It's doxygen syntax. Reminds me, it has probably been a while since anyone has printed the game docs :/

src/formats/chr.h
```diff
@@ -40,7 +40,8 @@ typedef struct {
     sd_sprite *photo;                       ///< Pilot photo
     float winnings_multiplier;              ///< Tournament winnings multiplier
     sd_chr_enemy *enemies[MAX_CHR_ENEMIES]; ///< List of enemy states in current tournament
-    unsigned int cutscene;                  ///< cutscene id for end of tournament
+    char bk_name[14];                       ///< cutscene bk for end of tournament
+    int tournament_id;                      ///< misnomer? set to 4 for WORLD.BK cutscene quirks
```
Vagabond, line 44: It does correspond to TOURNAMENT_NUMBER in the tournament compiler LST files.

Nopey: Ah, thanks. I did a search for "TOURNAMENT_ID" and came up blank :P
will remove the misnomer comment.

src/resources/trnmanager.c
```diff
@@ -3,26 +3,29 @@
...
 #include <stdio.h>
 #include <string.h>
 
-static void trnlist_node_free_callback(void *data) {
-    sd_tournament_file *trn = data;
-    sd_tournament_free(trn);
+static int trn_sort_compare_fn(const void *a, const void *b) {
+    sd_tournament_file const *trn_a = a;
+    sd_tournament_file const *trn_b = b;
+    return (trn_a->registration_fee > trn_b->registration_fee) - (trn_a->registration_fee < trn_b->registration_fee);
```
Vagabond, line 15: I *believe* the original game sorts by the ID first and then by the fee, if there's more than 1 with the same ID? See https://omf2097.com/wiki/doku.php?id=omf2097:robyrt_tourn

Nopey: thanks, i'll do a quick n' dirty TRN file edit to verify and then use the correct logic.

Vagabond (approved):

src/resources/trnmanager.c
Nopey: the linked wiki is wrong. tournaments are not sorted by `TOURNAMENT NUMBER`.

Vagabond: Good to know.

mrannanj (approved):

PR #1110, opened 2025-04-02
Use random deterministically

Nopey:
- game_state's rand is now seeded with pilot name, har choice, and player inputs.
- to prevent short (smaller than 32 tick) netplay rollbacks from changing rng outcomes, the player inputs' entropy is delayed before being fed to the rng.
- game_state rand is now used for HAR's visual effects including scrap, landing particles, etc.. ensuring they are consistent across clients and across rollbacks in netplay.

Both REC playback and netplay should benefit from this.


Nopey:
Converting to draft until I figure out why netplay desyncs

Vagabond:
I definitely think we do not care about visual effects using deterministic rand, at least not yet.

Nopey:
> I definitely think we do not care about visual effects using deterministic rand, at least not yet.

i don't see a current harm in visual effects using deterministic rand, and if we do not use deterministic rand for visual effects a single client can see multiple outcomes of a random outcome (first and second time its simulated after a rollback).

> clients might want to turn off those completely on per-client basis. Or add more.

sure? alternatively, such things could be a match setting (and thus, the same for both clients). this is my plan for the REIN cheat-- to make it apply to both net clients.

katajakasa:
> > clients might want to turn off those completely on per-client basis. Or add more.
> 
> sure? alternatively, such things could be a match setting (and thus, the same for both clients). this is my plan for the REIN cheat-- to make it apply to both net clients.

What is the use case for applying the cheat on both clients ? And even if we do, why does the scrap have to be synchronized ? This seems like an overly complicated implementation for no good reason. Even if users want to synchronize the amount of scrap in match settings, it still doesn't need to be synchronized precisely.

And when we add local settings for scrap amount, people will likely expect that to matter in all games, network games included (like it goes with all visual changes in all network games that I know). Overriding personal visual preference seems odd.

Nopey:
We can hide scrap in netplay matches (in accordance with player-local visual setting) while still simulating scrap.

Simulating scrap in a deterministic manner is required for it to maintain visually consistent across netplay rollbacks.

Nopey:
If we want scrap to be deterministic *and* skip simulating it when a player changes a setting, perhaps we add a separate rand_visual rng state to game_state for use in optional visual effects? then it would still be consistent across rollbacks

Vagabond:
> If we want scrap to be deterministic _and_ skip simulating it when a player changes a setting, perhaps we add a separate rand_visual rng state to game_state for use in optional visual effects? then it would still be consistent across rollbacks

Yes, this is a reasonable idea, if we want to avoid scrap pop on rollback.

PR #1111, opened 2025-04-02
Language loader fixes

katajakasa:
Done:
- Remove trailing newline in language strings
- Make arena use the new text renderer (bacause arena texts broke due to the newline removal)
- Some color define fixes

### DIFF
--- a/src/console/console.c
+++ b/src/console/console.c
@@ -17,6 +17,7 @@
 #define HISTORY_MAX 100
 #define BUFFER_INC(b) (((b) + 1) % sizeof(con->output))
 #define BUFFER_DEC(b) (((b) + sizeof(con->output) - 1) % sizeof(con->output))
+#define CONSOLE_TEXT_COLOR 0xFE
 
 // Console State
 console *con = NULL;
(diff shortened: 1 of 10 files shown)


src/formats/language.c
```diff
@@ -94,6 +94,9 @@ int sd_language_load(sd_language *language, const char *filename) {
         memreader *mr = memreader_open_from_reader(r, len);
         memreader_xor(mr, len & 0xFF);
         memread_buf(mr, language->strings[i].data, len);
+        if(language->strings[i].data[len - 1] == '\n') {
+            language->strings[i].data[len - 1] = '\0';
+        }
```
Nopey, line 99: Please apply this fixup in `resources/languages.c` instead of `formats/language.c` so as to not change languagetool behavior.

Nopey (changes requested):

src/formats/language.c
katajakasa: I considered handling linebreak in reader and writer, but decided it wasn't worth it and ended up doing this.

Nopey (approved):

PR #1112, opened 2025-04-03
Implement 'mi' (make iterations) tag

Nopey:
The original game uses this any time it needs to spawn more than one scrap in an animation, such as Jaguar's Scrap animation and Kreissack's Ker-Boom.

Because Kreissack's Ker-Boom now spawns 25 pieces of scrap with just `m12mi25`, the 2014 hack where his `m25`, `m27`, and `m29` tags spawned scrap (in addition to what they're supposed to spawn) has been removed.


LegendaryBlueShirt:
The spawning function in har.c has a hack that calls har_spawn_scrap whenever it's told to spawn something with the ID scrap, bolt, screw, or spark.  This needs to be reworked.

Nopey:
I know where that one is, its in har spawn scrap. Will fix in the morning.

Vagabond:
The scrap destruction seems pretty limp now

src/formats/taglist.c
```diff
@@ -81,7 +81,7 @@ const sd_tag sd_taglist[] = {
     {"mc",  0, NULL                                                                                                   },
     {"md",  1, "Destroy animation N"                                                                                  },
     {"mg",  1, "Gravity for spawned animation, default 0"                                                             },
-    {"mi",  1, NULL                                                                                                   },
+    {"mi",  1, "Instance count for spawned animation, default 1"                                                      },
```
katajakasa, line 84: The taglist is autogenerated from pyomftools, would be nice to get this to the csv file in there too.

katajakasa: (or if it's in the wiki, I will add it from there eventually)

katajakasa (approved):
Tested, nothing immediately horrible caught my eye.

PR #1113, opened 2025-04-03
Implemented new stun system

LegendaryBlueShirt:
Implemented the new stun system which has two parts

1.  Rather than going from max to 0, stun goes from 0 until it hits the endurance max, at which point a formula to set the stun value negative kicks in
2. Stun recovery formulas count up from a negative value back to 0

Minor: Also adjusted idle stance speed

Our stun damage appears to be accurate, save for a case where the base value should be halved.  I am still investigating this, but it's not a blocker for this.


LegendaryBlueShirt:
Drafting until I've fixed the rectests.

src/game/objects/har.c
```diff
@@ -779,20 +779,20 @@ void har_take_damage(object *obj, const str *string, float damage, float stun) {
...
     if(h->endurance < 1.0f) {
         if(h->state == STATE_STUNNED) {
             // refill endurance
-            h->endurance = h->endurance_max;
-        } else {
-            h->endurance = 0.0f;
+            h->endurance = 0;
         }
+    } else if(h->endurance >= h->endurance_max) {
+        h->endurance = ((((h->endurance - h->endurance_max) / 256) * -2.5) - 60) * 256;
```
Vagabond, line 795: A comment explaining what this is doing and what the expected result will be, would be useful?

src/game/objects/har.c
```diff
@@ -1758,6 +1758,63 @@ static void har_palette_transform(damage_tracker *damage, vga_palette *pal, void
     }
 }
 
+void har_handle_stun(object *obj) {
+    har *h = object_get_userdata(obj);
+
+    if(h->health <= 0) {
+        h->endurance = h->endurance_max - 2;
```
Vagabond, line 1765: Comment here?

src/game/objects/har.c
```diff
@@ -1758,6 +1758,63 @@ static void har_palette_transform(damage_tracker *damage, vga_palette *pal, void
...
+    } else {
+        if(h->endurance < 1) { // We are currently dizzy, recover gradually
+            if(obj->cur_animation->id == ANIM_STUNNED) {
+                h->endurance += h->stun_factor * 1.4;
+            }
+        } else { // Not yet stunned, handle regular stun recovery
+            if((obj->cur_animation->id == ANIM_IDLE) || (obj->cur_animation->id == ANIM_CROUCHING) ||
+               (obj->cur_animation->id == ANIM_VICTORY)) {
+                float hpfactor = ((h->health_max * 5.0 / 9.0) + h->health) / h->health_max;
+                float stunfactor = 1.0 * h->stun_factor * h->endurance / h->endurance_max / 256.0;
```
Vagabond, line 1775: Stunfactor seems to always be the same in all these cases, so just define it once?

src/game/objects/har.c
```diff
@@ -1758,6 +1758,63 @@ static void har_palette_transform(damage_tracker *damage, vga_palette *pal, void
...
+
+                h->endurance -= ((hpfactor * stunfactor) + 18.0 / 250.0) * 256;
```
Vagabond, line 1777: Move this below the if/else block, since it's also common code?

Vagabond:
Overall this seems good, and seems to work.

src/game/objects/har.c
LegendaryBlueShirt: It's slightly different but I'll make it more obvious

Vagabond (dismissed):

PR #1118, opened 2025-04-03
Allow _setup_keyboard to specify the user and which config to user

Vagabond:
In netplay, we sometimes want to use the player 1 keyboard controls for player 2 (the challengee). This change allows the game to apply the player 1 controls to the player 2 player for this situation.

### DIFF
--- a/src/game/game_state.c
+++ b/src/game/game_state.c
@@ -1100,7 +1100,7 @@ int game_state_num_players(game_state *gs) {
     return N_ELEMENTS(gs->players);
 }
 
-void _setup_keyboard(game_state *gs, int player_id) {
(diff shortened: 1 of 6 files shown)


Nopey (approved):
Patch could touch less files if you implemented this via an "extended" call that's only used when needed:
```c
// in game_state.h..
void _setup_keyboard_ex(game_state *gs, int_player_id, int control_id);
inline void _setup_keyboard(game_state *gs, int_player_id) {
    _setup_keyboard_ex(gs, player_id, player_id);
}
```

but this is fine as is

Vagabond:
Yeah, we don't tend to do a lot of that in this codebase, so I was unsure if we wanted to start.

katajakasa (approved):

PR #1119, opened 2025-04-03
Rewrite dissolve effects noise function

Nopey:
after some poking around in ghidra, I found OMF's dissolve effect code.
Magic number 0x6B comes directly from that.

`tan(10*PHI*v.y)` is to offset each row, simulating how OMF's draw_sprite_row_dissolve would shift the pattern depending on where the row started.

<details>
<summary>Visual Comparison. OMF, OpenOMF (this PR), OpenOMF (master)</summary>

image
image
image

</details>


Vagabond:
This definitely looks more accurate

katajakasa (approved):

PR #1120, opened 2025-04-03
Fix some bugs with unranked challengers:

Vagabond:
* Only play ending cutscene after beating the real tournament champion, not any unranked challengers that trigger thereafter
* Prevent beating an unranked challenger from triggering another unranked challenger

### DIFF
--- a/src/game/scenes/newsroom.c
+++ b/src/game/scenes/newsroom.c
@@ -392,7 +392,8 @@ int newsroom_create(scene *scene) {
             local->challenger = fight_stats->challenger;
             p2->pilot = local->challenger;
             fight_stats->challenger = NULL;
-        } else if(p1->chr) {
+        } else if(p1->chr && p2->pilot && p2->pilot->rank != 0) {
+            // we didn't just beat an unranked challenger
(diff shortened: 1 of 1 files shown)


src/game/scenes/newsroom.c
```diff
@@ -392,7 +392,8 @@ int newsroom_create(scene *scene) {
             local->challenger = fight_stats->challenger;
             p2->pilot = local->challenger;
             fight_stats->challenger = NULL;
-        } else if(p1->chr) {
+        } else if(p1->chr && p2->pilot && p2->pilot->rank != 0) {
+            // we didn't just beat an unranked challenger
```
Nopey, line 396: that comment is a bit awkward. perhaps:
```suggestion
            // our match was against a ranked opponent
```

Vagabond:
> I don't believe this is the correct fix. An unconditional lockout of unranked challengers after beating an unranked challenger makes fighting Jazzy (pilot 32 in world tournament) impossible since his condition is beating pilot 33 (Eva earlong, also unranked)

Good point.

LegendaryBlueShirt (approved):

PR #1121, opened 2025-04-03
Fix another issue with rec playback with errant ticks

Vagabond:


src/controller/rec_controller.c
```diff
@@ -198,8 +199,10 @@ void rec_controller_create(controller *ctrl, int player, sd_rec_file *rec) {
             hashmap_put_int(&data->tick_lookup, (rec->moves[i].tick * 10) + j, &rec->moves[i], sizeof(sd_rec_move));
             last_tick = rec->moves[i].tick;
         }
+        if((rec->moves[i].lookup_id == 2 || rec->moves[i].lookup_id == 10) && rec->moves[i].tick > data->max_tick) {
+            data->max_tick = rec->moves[i].tick;
+        }
```
katajakasa, line 204: So what happens here ?

Vagabond: I'm just keeping the max_tick always increasing even if there's some weird event with a lower tick (which still happens sometimes)

katajakasa (approved):

PR #1122, opened 2025-04-04
Attempt to keep the netplay games more in-sync and perform less rewinds

Vagabond:


src/controller/net_controller.c
```diff
@@ -267,19 +273,19 @@ void send_events(wtf *data) {
...
         }
     }
 
     if(!events) {
         // nothing to send, so send a blank event
-        // serial_write_int16(&ser, 0);
-        // serial_write_int32(&ser, data->last_tick - data->local_proposal - 1);
-        // last_sent = data->last_tick - data->local_proposal - 1;
+        serial_write_int32(&ser, data->last_tick - data->local_proposal - 1);
+        serial_write_int8(&ser, 0);
```
katajakasa, line 284: I'm starting to wonder if we should put read/write function pairs for protocol messages somewhere separate.

Vagabond: I don't know what this would look like or why we'd do it.

katajakasa: Something like a function for that contains serialization for one protocol message. I haven't really thought about it any further, except that this way sort of makes it difficult to see wtf we are doing here. I mean, we write int32 and int8 -- what is this stuff ? Having it in a named function with actual args might make it easier to read.

Anyway, not for this PR.

katajakasa (approved):

PR #1123, opened 2025-04-04
Move arena debug layer to use the new text renderer

katajakasa:
- Use the new text renderer to handle debug layer
- Move debug text ticks to dyntick function


Vagabond:
Reordering the functions makes the diff hard to read, but this seems to work ok.

Vagabond (approved):

PR #1125, opened 2025-04-04
Tournament enemy count is a uword not a udword

Vagabond:
This allows some of the old fan made tournaments to load.

### DIFF
--- a/src/formats/tournament.c
+++ b/src/formats/tournament.c
@@ -151,10 +151,11 @@ int sd_tournament_load(sd_tournament_file *trn, const char *filename) {
     }
 
     // Read enemy count and make sure it seems somwhat correct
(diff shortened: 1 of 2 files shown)


src/formats/tournament.c
```diff
@@ -271,7 +273,8 @@ int sd_tournament_save(const sd_tournament_file *trn, const char *filename) {
     }
 
     // Header
-    sd_write_udword(w, trn->enemy_count);
+    sd_write_dword(w, trn->enemy_count);
+    sd_write_dword(w, trn->enemy_count);
```
mrannanj, line 277: This second write should maybe be the unknown_b

katajakasa (approved):

PR #1127, opened 2025-04-04
run_rectests.sh: Allow interrupting the test run

mrannanj:


run_rectests.sh
```diff
@@ -56,6 +56,12 @@ tests=(
 temp_dir=$(mktemp -d)
 trap 'rm -rf "$temp_dir"' EXIT
 
+interrupt() {
+    echo "Test run interrupted" >&2
```
Vagabond, line 60: does this also do the rm-rf above?

mrannanj: That is run yeah, no temp dir left behind.

Vagabond (approved):

PR #1128, opened 2025-04-05
Bug Fix: Enemies can recover from overhead throw

LegendaryBlueShirt:
Fixed air knockback code overriding M200 at the end of overhead throw


Nopey (approved):
this fixes the buggy rec I posted 👍

PR #1129, opened 2025-04-05
Case insensitive tournament

mrannanj:


Nopey:
Are we expecting to be `.CHR` save compatible with the original game? If so, we might need to emulate DOS filename truncation (e.g. `MYAWES~1.TRN`)

If save compatibility is not a goal, a simpler patch would be the to keep the first two commits and then just make sure we're preserving the case of filenames internally.

Also, this code does not compile on Windows-- you are directly using POSIX file APIs.

Nopey (changes requested):
please don't use argtable3's case insensitive string compare, if you want one stick it in our c_string_utils.
please don't add special filename-normalization logic to chr.c, if we need more lenient filename handling (e.g. "MYTOUR~1.PIC" being resolved to "MyTournament.Pic"), a helper method should be added somewhere central (where it can be used in multiple places)

src/utils/c_string_util.c
```diff
@@ -19,6 +25,22 @@ char *strncpy_or_abort(char *dest, const char *src, size_t n) {
...
+    return s;
+}
+
+char *omf_dirname(char *s) {
+    char *p = strrchr(s, PATH_DELIMITER);
+    if(p != NULL) {
+        *p = '\0';
+    }
+    return s;
+}
```
mrannanj, line 42: Looking at the man page, it has this example:
```
              path       dirname   basename
              /usr/lib   /usr      lib
              /usr/      /         usr
              usr        .         usr
              /          /         /
              .          .         .
              ..         .         ..
```
So my implementation is clearly not there yet.

mrannanj: Well, now these tests pass, but the code looks horrible and I will see if there is something more that could be done here.

src/formats/internal/reader.c
```diff
@@ -15,12 +17,40 @@ struct sd_reader {
...
 sd_reader *sd_reader_open(const char *file) {
-    sd_reader *reader = omf_calloc(1, sizeof(sd_reader));
+    char *path_dup = omf_strdup(file);
+    char *path_dup2 = omf_strdup(file);
+    char *fn = omf_basename(path_dup);
+    char *directory = omf_dirname(path_dup2);
+
+    list dir_list;
+    list_create(&dir_list);
+    if(scan_directory(&dir_list, directory) != 0) {
```
Vagabond, line 27: What if we used scan_directory only when we were looking for a file with a potentially screwy cased name?

src/utils/c_string_util.h
```diff
@@ -9,6 +9,10 @@ char *strncpy_or_truncate(char *dest, const char *src, size_t n);
 // strncpy() that guarantees null-termination by aborting on overflow
 char *strncpy_or_abort(char *dest, const char *src, size_t n);
 
+// libgen.h
+char *omf_basename(char *s);
```
Vagabond, line 13: These would probably go in like a fs_util.c?

src/formats/internal/reader.c
Vagabond: or if we had like a `scan_directory_for_file` function we could call?

Vagabond:
> Are we expecting to be `.CHR` save compatible with the original game? If so, we might need to emulate DOS filename truncation (e.g. `MYAWES~1.TRN`)

I don't think I've seen anything use filename truncation, and I don't think we should support it unless we do and it feels necessary.

src/formats/internal/reader.c
mrannanj: Yeah, could make the scan optional.

src/utils/c_string_util.h
mrannanj: Well, they are purely string operations, but maybe that's reasonable. Anyway, need to cleanup the implementation first ..

src/formats/internal/reader.c
mrannanj: scan_directory_for_file was added.

src/formats/internal/reader.c
```diff
@@ -15,10 +17,16 @@ struct sd_reader {
 };
 
 sd_reader *sd_reader_open(const char *file) {
+#if !defined(_WIN32) && !defined(WIN32)
+    char path_buf[256];
+    strncpy_or_abort(path_buf, file, sizeof(path_buf));
+    if(!scan_directory_for_file(path_buf, sizeof(path_buf))) {
```
Vagabond, line 23: Do we want all the sd_readers to be case insensitive, or should we only call this from the CHR loader?

mrannanj: I guess all loaders case insensitive is the correct default, but is there an example where you don't want this?

mrannanj: Maybe if there is some system where the operation is heavy? It could be added, yes.

Vagabond: I guess we can start like this and see if anything ever breaks. My main concern is save games.

src/utils/scandir.c
```diff
@@ -57,8 +46,43 @@ int scan_directory(list *dir_list, const char *dir) {
...
+    }
+    omf_free(path_dup);
+    omf_free(path_dup2);
+    list_free(&dir_list);
+    return found;
+#endif
+}
+
 int scan_directory_prefix(list *dir_list, const char *dir, const char *prefix) {
-    check_valid_directory(dir);
```
Vagabond, line 61: Why was this removed?

mrannanj: It is asserting that a directory ends in the path delimiter, but that is a wrong assertion, and no reason to crash the game. It was crashing the game for me on some path that I don't remember anymore.

mrannanj: So the point of this change is that sanitizing paths should not be an assert crash, it can be still fixed at this point.

src/formats/internal/reader.c
mrannanj: Well, save games are a problem mainly since the font does not let you see lowercase. So, not quite sure what to do with those.

Vagabond:
I can confirm this fixes tournaments on linux, both the original ones and the mantis project.

Vagabond (approved):

PR #1134, opened 2025-04-05
Cutscene font renderer + some fixes

katajakasa:
I /think/ I got the str_split right. Probably. Maybe.


LegendaryBlueShirt:
Are there any cases where the string might have more than two sections?  (More than two split characters)  If so, a test case with more than two sections would be good to have

katajakasa:
> Are there any cases where the string might have more than two sections? (More than two split characters) If so, a test case with more than two sections would be good to have

Sure. More tests added.

Nopey (approved):

LegendaryBlueShirt (approved):

PR #1137, opened 2025-04-06
Do the debug ticking in render function for now

katajakasa:
Just tick in render. It should be fine, since debug is for developing only.

### DIFF
--- a/src/game/scenes/arena.c
+++ b/src/game/scenes/arena.c
@@ -1155,54 +1155,6 @@ static void free_debug_data(scene *scene) {
     local->debug = NULL;
 }
 
-static void arena_tick_debug(scene *scene) {
-    if(scene->gs->hide_ui) {
-        return;
-    }
-    // Note! This will return cached object if it exists.
-    debug_data *d = create_debug_data(scene);
-
-    char buf[100];
-    game_player *player[2];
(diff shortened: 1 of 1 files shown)


src/game/utils/score.c
```diff
@@ -334,6 +334,7 @@ int chr_score_clone(chr_score *src, chr_score *dst) {
         score_text t2;
         memcpy(&t2, t, sizeof(score_text));
         t2.text = create_text_obj(text_c(t->text));
+        text_set_horizontal_align(t2.text, text_get_horizontal_align(t->text));
```
Nopey, line 337: chr_score_t's `text *total;` is still experiencing alignment issues.

katajakasa: possibly resolved.

Nopey (approved):
this fixes the bug (where rewinding would break arena score text alignment).

would be nice if there was a `text_clone` method in text.c that copied the alignment/font info/etc for us, but do as you will.

katajakasa:
> this fixes the bug (where rewinding would break arena score text alignment).
> 
> would be nice if there was a `text_clone` method in text.c that copied the alignment/font info/etc for us, but do as you will.

Agreed.
https://github.com/omf2097/openomf/issues/1152

Issue #1133, opened 2025-04-05
[GAMEPLAY] Unranked Challenger Fight Score Carries Over From Previous Fight


CaptainLucian:
**Steps to reproduce**
What were you doing?
Playing tournament mode in the Katushai Challenge, when getting challenged by an Unranked Challenger and accepting that fight, the initial score for the fight starts at the value the previous fight ended on.

What happened?
The score for Unranked Challenger fights start at the number the last fight ended at.

What did you expect to happen?
I expected the Unranked Challenger fight scores to start at 0.

**Documentation**
I hit Angel once here, and that shouldn't get me to 177k

[20250405_102034.zip](https://github.com/user-attachments/files/19616614/20250405_102034.zip)

Here's a cleaner pic showing 423k points with no damage done to Angel with the FIGHT message still on screen.
![Image](https://github.com/user-attachments/assets/88b0338e-c804-4e08-8e54-6b4372b45292)

**Platform (please complete the following information):**
 - OS: [e.g. Windows, MacOS, Linux] Windows
 - Version 1.8.1-RC1

**Additional context**
It's extremely low priority, but it did reset to 0 originally (Confirmed by fighting Christian in the North American Open)

![Image](https://github.com/user-attachments/assets/1e64b618-52cf-44c9-a7fe-d04ceb1423ab)

Vagabond:
Oh yeah, I bet I forgot to clear the score

LegendaryBlueShirt:
This is probably why unranked challengers are giving so much dang money!

Issue #1131, opened 2025-04-05
[GAMEPLAY] Scrap and Destruction Moves (at least Jaguar's) Do Not Increase Score In Tournament Mode


CaptainLucian:
**Steps to reproduce**
What were you doing?
Fighting in Tournament Mode with a Jaguar, doing the Scrap move (down down up punch) and the destruction move (down down up kick while in the middle of the scrap move) do not add the Destruction Bonus for the fight.

What happened?
No additional points were given for destroying the enemy HAR.

What did you expect to happen?
Additional points were given for performing the destruction move, originally it provided 100k points.

**Documentation**
This is the same set of attachments as my Unranked Challenger refight issue, but to do those refights you need to be doing destruction moves, so it should hopefully have the relevant information.
[screenshot_20250405_074508.zip](https://github.com/user-attachments/files/19615420/screenshot_20250405_074508.zip)

This screenshot is from the original running in DosBox

![Image](https://github.com/user-attachments/assets/45ecca1e-8c1f-47eb-9501-25bd0529c82f)

**Platform (please complete the following information):**
 - OS: Windows
 - Version 0.80

Vagabond:
@mrannanj might be the best to look at this, I think he did a lot of the tournament scoring

PR #1139, opened 2025-04-06
game: Reset score on arena creation in tournament mode

mrannanj:
#1133 and #1131

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -2482,9 +2482,10 @@ void har_finished(object *obj) {
             game_state_find_object(obj->gs, game_player_get_har_obj_id(game_state_get_player(obj->gs, !h->player_id)));
         har_face_enemy(obj, enemy_obj);
     } else if(h->state == STATE_SCRAP || h->state == STATE_DESTRUCTION) {
-        // play vistory animation again, but do not allow any more moves to be executed
+        // play victory animation again, but do not allow any more moves to be executed
         h->state = STATE_DONE;
         har_set_ani(obj, ANIM_VICTORY, 0);
+        har_event_done(h, ctrl);
     } else if(h->state == STATE_VICTORY || h->state == STATE_DONE) {
         // prevent object from being freed, hold last sprite of animation indefinitely
         obj->animation_state.finished = 0;
--- a/src/game/scenes/arena.c
+++ b/src/game/scenes/arena.c
@@ -1865,7 +1865,7 @@ int arena_create(scene *scene) {
     chr_score_set_tournament_mode(game_player_get_score(_player[1]), local->tournament);
 
     // Reset the score
-    chr_score_reset(game_player_get_score(_player[0]), !is_singleplayer(scene->gs));
+    chr_score_reset(game_player_get_score(_player[0]), is_tournament(scene->gs));
     chr_score_reset(game_player_get_score(_player[1]), 1);
 
     // Reset the win counter in single player mode
--- a/src/game/scenes/lobby.c
+++ b/src/game/scenes/lobby.c
@@ -1456,6 +1456,8 @@ int lobby_create(scene *scene) {
         // destroy any leftover controllers
         game_player *player = game_state_get_player(scene->gs, i);
         game_player_set_ctrl(player, NULL);
+        // reset any scores
+        chr_score_reset(game_player_get_score(game_state_get_player(scene->gs, i)), true);
     }
     reconfigure_controller(scene->gs);
 


src/game/scenes/arena.c
```diff
@@ -1870,7 +1870,7 @@ int arena_create(scene *scene) {
     chr_score_set_tournament_mode(game_player_get_score(_player[1]), local->tournament);
 
     // Reset the score
-    chr_score_reset(game_player_get_score(_player[0]), !is_singleplayer(scene->gs));
+    chr_score_reset(game_player_get_score(_player[0]), is_tournament(scene->gs));
```
Vagabond, line 1873: just checked, 2 player indeed does not reset between rounds.

Vagabond: Although it resets if you exit 2 player mode.

mrannanj: Well, do you think is the correct fix? I did check that it works in the case of unranked challenger.

Vagabond: I think this is fine, but we should also reset the score when the player goes back to the main menu and to the lobby. (so in the mainmenu and lobby create functions?)

src/game/objects/har.c
```diff
@@ -1887,6 +1887,12 @@ void har_tick(object *obj) {
     if(player_frame_isset(obj, "aa")) {
         h->air_attacked = 0; // This tag allows you to attack again
     }
+
+    if(h->state == STATE_DONE) {
+        // match is over
+        har_event_done(h, ctrl);
```
Vagabond, line 1893: Won't this get called a *lot*? Also state_done is only used if you can't do a scrap/destruction, IIRC?

mrannanj: This awards the scrap and destruction points, and they can be awarded only once. But not sure if this gets called a lot in the end of a fight.

src/game/scenes/arena.c
mrannanj: I'm not sure what problem does that solve?

Vagabond: So that the score is zeroed out if you go from singleplayer to 2 player, or vice versa? or if you return to the lobby from a fight?

src/game/objects/har.c
mrannanj: What I can say is that this does fix the issue of scrap and destruction points not being awarded.

src/game/scenes/arena.c
mrannanj: I guess the point of this code is that sometimes score is not to be cleared, and that would be the the non tournament single player mode, no idea when else you would want not to clear the score.

Vagabond: Right, but you've changed this to *only* clear the score in tournament mode now, you've now potentially broken the score clearing in 2 player (when you exit 2 player mode and go back to the menu), and netplay scoring, when you exit back to the lobby.

src/game/objects/har.c
Vagabond: Yes, but this might be more properly done via the arena functions that handle the end of the round/match?

src/game/scenes/arena.c
Vagabond: looks like mainmenu already resets the scores in mainmenu_create

src/game/objects/har.c
mrannanj: That's exactly what this does.

mrannanj: This used to be called here with he condition:
```
    if((h->state == STATE_DONE) && player_is_last_frame(obj) && obj->animation_
state.entered_frame == 1) {
        // match is over
        har_event_done(h, ctrl);
    }
```
But it was removed I think by accident in commit e3ffe261f14a413b891cfcf01211ffb97741bded

katajakasa: That player_is_last_frame call is probably invalid. It checks if the last frame is ongoing, not if the last frame has ended.

mrannanj: I don't know what the conditions there mean, but I will check if the version in this pull request gets called just once or many times.

mrannanj: It gets called a lot in the end with the current version, but doesn't make the scores be awarded multiple times.

Vagabond: I still think this is better removed as a HAR event and handled properly in the arena round end stuff. Just checking the last frame is not safe for several of the HARs, etc.

mrannanj: I put in this now:
```
    if(h->state == STATE_DONE && obj->animation_state.entered_frame == 1) {
```
This gets called about 10 times, which is a lot less than without the added condition ```obj->animation_state.entered_frame == 1```

Vagabond: call chr_score_done or whatever here: https://github.com/omf2097/openomf/blob/master/src/game/scenes/arena.c#L1272 just decide how many ticks you want to wait.

src/game/scenes/arena.c
mrannanj: I think this change is correct, but if not, can you propose a better one?

src/game/objects/har.c
mrannanj: That's a bit late to call it since it's immediately before scene switch, but I'll see if there is something nicer ..

mrannanj: Nah, i think this current change is the correct place to call this. The others are too near the scene change so you'd miss the floating score text animation.

LegendaryBlueShirt:
```if(player_frame_isset(obj_har[0], "be") || player_frame_isset(obj_har[1], "be") ||
                   chr_score_onscreen(s1) || chr_score_onscreen(s2)) {
                    local->ending_ticks = 50;
                }```
 This code should be resetting the scene transition countdown if you display any kind of score

Comment (mrannanj) [reply src/game/objects/har.c:1893]: Well, the code is designed so that what I have pushed here is the expected way. But I'll remove the "DONE!" debug print.

File: src/game/scenes/lobby.c [hunk 3]
```diff
@@ -1090,6 +1090,8 @@ void lobby_tick(scene *scene, int paused) {
 
                             game_player_set_selectable(challengee, 1);
 
+                            chr_score_reset(game_player_get_score(game_state_get_player(gs, 0)), true);
```
Vagabond, line 1093: There's a third place a game can launch, but I'd still recommend putting these in lobby_create for simplicity.

src/game/objects/har.c
Vagabond: Calling a har hook from a tick function still feels very wrong to me (and I wrote the har hook system originally). It feels like calling this from where the har gets set to DONE would be more appropriate (and I think that's in the arena now?)

mrannanj: Okay, moved this to har_finished().

Vagabond: If that still works, I like it much better.

mrannanj: It works, I tested it.

Vagabond (approved):

PR #1142, opened 2025-04-06
Added XOR check for facing and sprite flipping

### LINKED ISSUE #1090: Incorrect enemy HAR sprite for overhead throws
Gargoyle and Jaguar overhead throws still seem to select the wrong enemy HAR sprite position (sprite is now correct, position is still off)

Image
Image
Overhead throw shouldn't look like this.

Instead in omf2097:
Image
Image

LegendaryBlueShirt:
fixes #1090

### DIFF
--- a/src/game/protos/object.c
+++ b/src/game/protos/object.c
@@ -370,7 +370,7 @@ void object_render(object *obj) {
     }
 
     // Set X coord, take into account the HAR facing.
-    if(object_get_direction(obj) == OBJECT_FACE_LEFT) {
+    if((object_get_direction(obj) == OBJECT_FACE_LEFT) != ((rstate->flipmode & FLIP_HORIZONTAL) != 0)) {
         x = obj->pos.x - cur_sprite->pos.x + rstate->o_correction.x - object_get_size(obj).x;
     } else {
         x = obj->pos.x + cur_sprite->pos.x + rstate->o_correction.x;


src/game/protos/object.c
```diff
@@ -370,7 +370,7 @@ void object_render(object *obj) {
     }
 
     // Set X coord, take into account the HAR facing.
-    if(object_get_direction(obj) == OBJECT_FACE_LEFT) {
+    if((object_get_direction(obj) == OBJECT_FACE_LEFT) != ((rstate->flipmode & FLIP_HORIZONTAL) != 0)) {
```
katajakasa, line 373: aaa this makes my brain make flips. I think it makes sense though.

katajakasa (dismissed):

src/game/protos/object.c
Vagabond: I hate this

katajakasa: I was thinking something like
facing | flip | result
left | yes | 0
right | no | 1

etc.

Vagabond: can we make it an explicit xor?

LegendaryBlueShirt: Does C have that?

Vagabond: ^ is xor in C

Vagabond (approved):

PR #1145, opened 2025-04-06
Try to make it easier to debug NAT issues

Vagabond:
Also try to avoid a potential infinite loop

### DIFF
--- a/src/game/utils/nat.c
+++ b/src/game/utils/nat.c
@@ -230,9 +230,9 @@ void nat_try_pmp(nat_ctx *ctx) {
         sendpublicaddressrequest(&ctx->natpmp);
         int r = readpmpresponse(ctx, &response);
         if(r == 0) {
+            ctx->type = NAT_TYPE_PMP;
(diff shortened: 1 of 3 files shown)


katajakasa (approved):
Sure.

src/game/utils/nat.c
```diff
@@ -121,9 +121,12 @@ int readpmpresponse(nat_ctx *ctx, natpmpresp_t *response) {
 #else
         FD_SET(ctx->natpmp.s, &fds);
 #endif
-        getnatpmprequesttimeout(&ctx->natpmp, &timeout);
-        select(FD_SETSIZE, &fds, NULL, NULL, &timeout);
-        r = readnatpmpresponseorretry(&ctx->natpmp, response);
+        // getnatpmprequesttimeout(&ctx->natpmp, &timeout);
+        timeout.tv_sec = 3;
```
katajakasa, line 125: do we even need 3 seconds ?

katajakasa (approved):

src/game/utils/nat.c
Vagabond: yeah, probably, although some routers are *slow*

PR #1146, opened 2025-04-06
Allow setting logging level via commandline argument --log-level

katajakasa:
For future needs.

### DIFF
--- a/src/utils/log.c
+++ b/src/utils/log.c
@@ -4,11 +4,13 @@
 #include <assert.h>
 #include <stdarg.h>
 #include <stdio.h>
+#include <string.h>
 #include <time.h>
 
 #include "utils/allocator.h"
 
 #define MAX_TARGETS 3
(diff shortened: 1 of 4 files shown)


src/utils/log.c
```diff
@@ -50,6 +51,24 @@ void log_init(void) {
     state->target_count = 0;
 }
 
+log_level log_level_text_to_enum(const char *level, log_level default_value) {
+    for(int i = 0; i < LOG_LEVELS; i++) {
+        if(strncmp(level_names[i], level, strlen(level_names[i])) == 0) {
```
mrannanj, line 56: I guess this would let for example DEBUG12345 be DEBUG too. since comparing only up to DEBUG. Perhaps not a major problem.

katajakasa: Idea was to strip out any accidental whitespace garbage

Vagabond (approved):

Issue #1153, opened 2025-04-07
[GAMEPLAY] Crash on second match in netplay


LegendaryBlueShirt:
**Steps to reproduce**
Play a match online
Get booted back to the lobby < technically this is also a bug
Challenge again and attempt to start another match
Select a stage

What happened?
Crash!

What did you expect to happen?
The next fight

**Platform (please complete the following information):**
 - OS: Win 10
 - Version https://github.com/omf2097/openomf/suites/36805282009/artifacts/2895245708

PR #1154, opened 2025-04-07
Don't free recs when --rec is passed on cli, fix double-free in sd_rec_free

Vagabond:
This should help the netplay crash. #1153


katajakasa (approved):

mrannanj:
We tested this, and the double free in #1153 is not fixed. But maybe this fixes something else.

mrannanj (approved):

Issue #1159, opened 2025-04-07
Exception (Segment violation) on startup on Haiku OS


Peppersawce:
**Steps to reproduce**
What were you doing?
Running a compiled binary on [Haiku OS](https://www.haiku-os.org/)

What happened?
Binary starts running up until a blank 640x400 window is drawn then segfaults.
![Image](https://github.com/user-attachments/assets/a034e1d5-ce34-4e26-a8f3-1835642c3b8d)

What did you expect to happen?
To see the intro screen with lightning and thunder I guess.

**Platform:**
 - OS: Haiku x86_64
 - Version: latest development (hrev58793 Apr 6 2025)

**Additional context**
Building from the [latest source available](https://github.com/omf2097/openomf/tree/2aeba123ba3117770a33d6ce54a634d7a73cd78a) with a small patch to `src/resources/pathmanager.c` to set OS-specific paths.
The binary shows no issues with these paths, although it seems to not create its own settings folder without also patching that in. (but I might've missed something with my first patch)

Attaching both the game's log and the debug report generated by the OS.

[Openomf log](https://github.com/user-attachments/files/19639820/openomf.log)
[OS Debug report](https://github.com/user-attachments/files/19639832/OpenOMF-8845-debug-07-04-2025-21-35-47.report.log)

katajakasa:
Thanks for the report! The binary seems to fail when setting up opengl shaders. Unfortunately I don't have much experience in Haiku, but it seems you are using Mesa for video. Please try running the openomf binary with LIBGL_ALWAYS_SOFTWARE=1 environmental variable set -- that should ask mesa to run the game using llvmpipe. e.g. `LIBGL_ALWAYS_SOFTWARE=1 openomf`.

What sort of graphics hardware are you using by the way ? Device + driver ?

Peppersawce:
Thanks for the quick answer :)

Haiku should run on the "Intel Corporation HD Graphics 620", the integrated GPU, there's no hardware acceleration for it but it is recognized by the system and should have a... basic driver iirc.

I tried setting the variable but no dice, I think Mesa uses "llvmpipe" by default on here anyway.

For what is worth I run other sdl2 games that use Mesa and llvmpipe with no issues, the OS should be capable of running the game.

katajakasa:
[...]

PR #1162, opened 2025-04-08
Don't segfault on a missing shader

Vagabond:
Related to #1159

### DIFF
--- a/src/utils/log.c
+++ b/src/utils/log.c
@@ -12,6 +12,8 @@
 #define MAX_TARGETS 3
 #define LOG_LEVELS 4
 
+static char last_error[256];
+
 typedef struct log_target {
     FILE *fp;
     log_level level;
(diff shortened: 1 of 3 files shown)


katajakasa (approved):
lgtm

src/utils/log.c
```diff
@@ -160,6 +162,11 @@ void log_msg(log_level level, const char *fmt, ...) {
         va_start(args, fmt);
         vfprintf(target->fp, fmt, args);
         va_end(args);
+        if(level == LOG_ERROR) {
+            va_start(args, fmt);
```
turol, line 166: Calling `va_start` twice on the same argument is UB. You need to use `va_copy`

Vagabond: I thought va_start() reinitialized the list safely? Do you have a link to anything that discusses this? I'm having trouble finding anything.

Vagabond: The linux man page for va_start says: `Multiple traversals of the list, each bracketed by va_start() and va_end() are possible. va_end() may be a macro or a function.`

turol: Seems to have been changed recently then.

Vagabond: Happy to change it if you can find a reference that shows it might be UB. UBsan doesn't seem to flag it either?

turol: cppcheck at least used to.

Vagabond: I installed cppcheck and ran it on this file (and the whole project). It did find some other stuff, but nothing about va_args here.

katajakasa: I remember having to do va_copy at some point too, though it's been a while. Probably something that was changed. Or possibly mandela effect.

PR #1170, opened 2025-04-08
Sanitize new pilot .CHR filenames

mrannanj:


mrannanj:
This is draft because it's missing tests, but seems a thing that could be tested automatically quite easily.

src/utils/c_string_util.c
```diff
@@ -19,6 +19,13 @@ char *strncpy_or_abort(char *dest, const char *src, size_t n) {
     return dest;
 }
 
+char *strtoupper(char *s) {
+    for(size_t i = 0; s[i] != '\0'; ++i) {
+        s[i] = toupper(s[i]);
```
Nopey, line 24:
```suggestion
        s[i] = toupper((unsigned char)s[i]);
```
passing negative numbers to toupper is UB.
From cppreference.com:
> int ch - character to be converted. If the value of ch is not representable as unsigned char and does not equal EOF, the behavior is undefined.

src/resources/sgmanager.c
```diff
@@ -143,12 +131,28 @@ int sg_load(sd_chr_file *chr, const char *pilotname) {
...
+    char target_filename[1024];
+    char scan_filename[1024];
+
     const char *dirname = pm_get_local_path(SAVE_PATH);
-    snprintf(filename, sizeof(filename), "%s%s.CHR", dirname, chr->pilot.name);
-    int ret = sd_chr_save(chr, filename);
+    snprintf(target_filename, sizeof(target_filename), "%s%s.CHR", dirname, chr->pilot.name);
+    strncpy_or_abort(scan_filename, target_filename, sizeof(scan_filename));
+
+    // Try to make the save uppercase.
```
Nopey, line 141: This comment does not relate to the call to `scan_directory_for_file` and `rename` below.

The transformation we should be applying to get the savegame name (as per the original game):
- Any character a through z or A through Z, replace with a capital letter A-Z.
- Any other character, replace with an underscore `_`
- (Append .CHR)

As an example, my character whose name is `jo dU-lI` gets saved to filename `JO_DU_LI.CHR` in OMF 2097.

As for the scan_directory_for_file and rename call below.. Why do we need a case insensitive scan? We know the old (oomf 0.8.0 and prior) filename, it's just `name .. ".CHR"`. and we know the new filename, as described above. Check if the file exists with `fopen` (standard C) or `stat` (POSIX) before deciding to rename.

testing/test_trn.c
```diff
@@ -34,7 +34,7 @@ void test_sd_trn_roundtripping(void) {
 
     n_trn.enemies[0] = omf_calloc(1, sizeof(sd_pilot));
     n_trn.enemy_count = 1;
-    snprintf(n_trn.enemies[0]->name, 18, "test_pilot");
+    snprintf(n_trn.enemies[0]->name, 18, "TEST_PILOT");
```
Nopey, line 37: Why are you changing this test case?
image
The original game's enemies have mixed case.

mrannanj: This was because the forced rename was happening and test was failing, but it might well be the wrong change. Have to return to that.

src/utils/c_string_util.c
mrannanj: Dropped strtoupper()

src/formats/chr.c
```diff
@@ -285,6 +286,15 @@ int sd_chr_save(sd_chr_file *chr, const char *filename) {
     return SD_SUCCESS;
 }
 
+void sd_chr_filename(const sd_chr_file *chr, char filename[static 9]) {
+    size_t len = min2(strlen(chr->pilot.name), 8);
+    for(size_t i = 0; i < len; ++i) {
+        unsigned char c = chr->pilot.name[i];
+        filename[i] = isalnum(c) ? toupper(c) : '_';
```
Nopey, line 293: same critique as before with strtoupper. cast the potentially-signed `char` to `unsigned char` before passing to isalnum and toupper to avoid UB.

src/formats/chr.h
```diff
@@ -95,6 +95,14 @@ int sd_chr_load(sd_chr_file *chr, const char *filename);
  */
 int sd_chr_save(sd_chr_file *chr, const char *filename);
 
+/*! \brief Transform CHR pilot name to filename.
+ *
+ * \param chr CHR struct pointer to get pilot name from.
+ * \param filename Output parameter where the filename is written.
+ *                 This needs a buffer of size 9: 8 for letters and 1 for '\0'.
+ */
+void sd_chr_filename(const sd_chr_file *chr, char filename[static 9]);
```
Nopey, line 104: is there a reason sd_chr_filename doesn't append the `.CHR`?
and as per discussion in discord, let's simplify the `char filename[static 9]` parameter to something like a `char *filename` with a note in the doc comment stating how large a buffer is required (maybe provide a constant in the header, something like SD_CHR_FILENAME_LEN?).

mrannanj: It's because the calling function is doing join with dirname and .CHR already, and the rename from old to normalized follow that pattern, so it would be messier to have .CHR here. About the 8 length, sure could put it as some constant somewhere.

testing/test_trn.c
mrannanj: This change was dropped.

tools/pcxtool/main.c
```diff
@@ -91,14 +92,6 @@ static void show_pcx(pcx_file *pcx) {
     }
 }
 
-static int file_exists(char const *filename) {
```
Nopey, line 94: With the removal of file_exists, could you not remove the `<unistd.h>` and `<io.h>` includes?
the if-windows <errno.h> has always been redundant, as well.

mrannanj: Yeah, was able to remove some includes related to this here and in pathmanager.

src/resources/pathmanager.c
```diff
@@ -4,17 +4,17 @@
...
 #else
 #include <sys/stat.h> // mkdir
-#include <unistd.h>
 #endif
 
 #include "resources/pathmanager.h"
 #include "utils/allocator.h"
 #include "utils/c_string_util.h"
 #include "utils/crash.h"
+#include "utils/io.h"
```
Nopey, line 17: pathmanager.c could use file_exists to implement pm_in_portable_mode.
might remove the need for some of the above includes such as `<shlwapi.h>`

src/formats/chr.h
mrannanj: Yeah, need to have the old and new filename format have similar functions, so will most likely add the `.CHR` there, but I did add `MAX_CHR_BASENAME` in chr.h now. Anyway, this is currently broken if you for example create a pilot `A/B`, which becomes `A_B` and fails the first load on pilot creation.

src/resources/pathmanager.c
mrannanj: Okay, pm_in_portable_mode() uses file_exists() now.

src/formats/chr.h
Nopey: path injection in chr character names is less than ideal :sweat_smile:
perhaps skip the old-to-new-filename-renaming stuff if the character name contains `/` or `\`?

> Anyway, this is currently broken if you for example create a pilot `A/B`,

by currently, i assume you mean without this PR?

mrannanj: Well the pilot name `A/B` is currently problematic with and without this pr, but I'll make it work. I rewrote this now with str but it's still broken in a couple places.

PR #1171, opened 2025-04-09
Prevent any moves while stunned

LegendaryBlueShirt:
Return from the function right away after checking for button mashing to recover faster.

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -2235,7 +2235,7 @@ int har_act(object *obj, int act_type) {
     }
 
     if(h->endurance < 0) {
-        if(prefix != 1) { // Mash to recover from stun faster!
+        if(prefix == 'K' || prefix == 'P') { // Mash to recover from stun faster!
             h->endurance += 512;
         }
         return 0;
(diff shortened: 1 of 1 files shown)


src/game/objects/har.c
```diff
@@ -2229,15 +2229,18 @@ int har_act(object *obj, int act_type) {
...
-        h->endurance += 512;
-    }
-
     uint32_t input_staleness = obj->gs->tick - h->input_change_tick;
     if(input_changed) {
         h->input_change_tick = obj->gs->tick;
     }
 
+    if(h->endurance < 0) {
+        if(prefix != 1) { // Mash to recover from stun faster!
```
Nopey, line 2238:
```suggestion
        if(prefix == 'K' || prefix == 'P') { // Mash to recover from stun faster!
```
If I understand the significance of 'prefix' correctly, this is equivalent.
Unless prefix is sometimes equal to `\0` here?

LegendaryBlueShirt: I agree locality of logic makes sense here

katajakasa (approved):
Seems clear enough

PR #1174, opened 2025-04-09
Small changes

### LINKED ISSUE #1173: [GAMEPLAY] First Tick Skipped For Attacks (All animations?)
**Steps to reproduce**
Nova 4P

What happened?
Nova immediately goes from idle to 4P frame B

What did you expect to happen?
4P frame A should show for one tick

**Additional context**
This has sweeping consequences...

katajakasa:


Vagabond:
Congratulations for breaking so many rectests, that looks like a record.

LegendaryBlueShirt:
If it's any consolation, this does fix my issue with Nova's 4P

LegendaryBlueShirt (changes requested):
Currently Jaguar scrap, chronos matter phasing and katana wall spin are busted so I can't fix the rectests for those.

Jaguar Scrap - If the other HAR is too far away Jaguar will give up and resume victory pose
Chronos, Katana - Matter Phasing and Wall Spin do not flip around to go the other way

PR #1175, opened 2025-04-09
Redo HAR push logic

LegendaryBlueShirt:
Logic now follows what we found in the decomp

Pushing HARs while walking is much smoother than before


src/game/scenes/arena.c
```diff
@@ -1305,6 +1305,52 @@ void arena_static_tick(scene *scene, int paused) {
     gui_frame_tick(local->game_menu);
 }
 
+void push_players(scene *scene, game_player *p1, game_player *p2) {
+    object *obj_p1 = game_state_find_object(scene->gs, game_player_get_har_obj_id(p1));
+    object *obj_p2 = game_state_find_object(scene->gs, game_player_get_har_obj_id(p2));
+    har *h1 = obj_p1->userdata;
+    har *h2 = obj_p2->userdata;
+
+    if(!(obj_p1->pos.y == ARENA_FLOOR || obj_p2->pos.y == ARENA_FLOOR)) {
```
Vagabond, line 1314:
```suggestion
    // Check at least one HAR is on the floor
    if(!(obj_p1->pos.y == ARENA_FLOOR || obj_p2->pos.y == ARENA_FLOOR)) {
```

src/game/scenes/arena.c
```diff
@@ -1299,6 +1299,53 @@ void arena_static_tick(scene *scene, int paused) {
...
+        float p1x = obj_p1->pos.x;
+        float p2x = obj_p2->pos.x;
+        if(p1x < p2x) {
+            p1x -= 1;
+            p2x += 1;
+        } else {
+            p1x += 1;
+            p2x -= 1;
+        }
+        if(p1x < ARENA_LEFT_WALL) {
```
katajakasa, line 1332: clamp() function for these, probably.

src/game/scenes/arena.c
```diff
@@ -1299,6 +1299,53 @@ void arena_static_tick(scene *scene, int paused) {
     gui_frame_tick(local->game_menu);
 }
 
+void push_players(scene *scene, game_player *p1, game_player *p2) {
```
katajakasa, line 1302: Doxygen-style function comment with a short explanation of what happens here would be nice.

src/game/objects/har.h
```diff
@@ -146,14 +148,14 @@ typedef struct har_t {
 
     int in_stasis_ticks; // Handle stasis activator
     int throw_duration;
+    int height;
```
katajakasa, line 151: height of what ? Variable name could be move descriptive.

LegendaryBlueShirt: I'm not sure har.har_height would be more descriptive than it already is

katajakasa: I was thinking more in line of assumed_height or estimated_height or current_estimated_height or somesuch. Since this is not the height we get from sprite. Alternatively, code comment would also work.

LegendaryBlueShirt: I'm adding a comment for how it's used.

katajakasa (approved):

PR #1179, opened 2025-04-09
Stubbing out ideas for a path library

katajakasa:
Just throwing ideas at the wall

### DIFF
--- a/src/utils/path.c
+++ b/src/utils/path.c
@@ -1,0 +1,2 @@
+#include "utils/path.h"
+
--- a/src/utils/path.h
+++ b/src/utils/path.h
@@ -11,7 +11,8 @@ typedef struct path {
 } path;
 
 // Normalize separator to "/"
-void path_create(path *path, const char *str);
+void path_from_c(path *path, const char *str);
+void path_from_parts(path *path, ...);
 void path_preferences_dir(path *dst);  // Init with preferences directory (users writeable directory)
 void path_resources_dir(path *dst); // Init with resources directory (read-only game data)
 
@@ -31,7 +32,7 @@ bool path_rmdir(const path *path);
 void path_glob_begin(const path *path, iterator *iter);
 void path_glob_iter_begin(const path *path, const char *pattern, iterator *iter);
 
-void path_append(path *path, const char *str); // Append new element
+void path_append(path *path, ...); // Append new elements
 void path_parent(path *path);  // Drop last element
 void path_absolute(path *path);
 
--- a/src/utils/path.h
+++ b/src/utils/path.h
@@ -36,7 +36,6 @@ void path_append(path *path, ...); // Append new elements
 void path_parent(path *path);  // Drop last element
 void path_absolute(path *path);
 
-bool path_file_length(path *path, size_t *length);
-bool path_read_file(path *path, char *buffer, size_t length);
+FILE *path_open(const path *path, const char *mode);
 
 #endif // PATH_H
--- a/src/utils/path.h
+++ b/src/utils/path.h
@@ -1,0 +1,41 @@
+#ifndef PATH_H
+#define PATH_H
+
+#include <stdbool.h>
+#include <stdio.h>
+
+#include "utils/iterator.h"
+
+typedef struct path {
+    char path[1024]; // overkill
+} path;
+
+// Normalize separator to "/"
+void path_create(path *path, const char *str);
+void path_preferences_dir(path *dst);  // Init with preferences directory (users writeable directory)
+void path_resources_dir(path *dst); // Init with resources directory (read-only game data)
+
+bool path_is_directory(const path *path);
+bool path_is_file(const path *path);
+bool path_exists(const path *path);
+
+const char *path_ext(const path *path);
+const char *path_basename(const path *path);
+const char *path_dirname(const path *path);
+
+bool path_touch(const path *path);
+bool path_unlink(const path *path);
+bool path_mkdir(const path *path);
+bool path_rmdir(const path *path);
+
+void path_glob_begin(const path *path, iterator *iter);
+void path_glob_iter_begin(const path *path, const char *pattern, iterator *iter);
+
+void path_append(path *path, const char *str); // Append new element
+void path_parent(path *path);  // Drop last element
+void path_absolute(path *path);
+
+bool path_file_length(path *path, size_t *length);
+bool path_read_file(path *path, char *buffer, size_t length);
+
+#endif // PATH_H


src/utils/path.c
```diff
@@ -0,0 +1,389 @@
...
+#include <shlobj.h>
+#include <shlwapi.h>
+#include <windows.h>
+#else
+#include <dirent.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#endif
+
+#define ENSURE_ZERO(str) str[PATH_MAX_LENGTH - 1] = '0'
```
mrannanj: Is this missing backslash from the zero? Like, `'\0'`, not `'0'`

src/utils/blob.h
```diff
@@ -0,0 +1,43 @@
+/**
+ * @brief Just a dumb buffer object
```
Nopey: blob is unused and has no tests.

src/utils/blob.h
```diff
@@ -0,0 +1,43 @@
...
+
+#ifndef BLOB_H
+#define BLOB_H
+
+#include <stddef.h>
+
+typedef struct blob {
+    unsigned char *data;
+    size_t size;
+} blob;
```
Nopey: Are these members private, or are users of blob allowed to reach in and muck around?

This header provides no getters for data or size.

src/utils/blob.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Allocate the blob object and initialize its contents.
+ * @return Mew initialized blob object
+ * @param size How much space to preallocate for the data
+ */
+blob *blob_create(size_t size);
```
Nopey: Why is the 16-or-so-byte blob struct heap allocated? If it was to make `blob*` an opaque pointer, we shouldn't have the blob struct defined in header.

Our str, vector, and hashtable containers all let the caller determine where the struct is stored.

Nopey: Behavior if `size` is zero is undescribed; and current implementation calls `omf_malloc(0)` in that case.

src/utils/blob.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Resize the blob object contents to the new size
+ * @param b Existing blob object
+ * @param size New size of the buffer
+ */
+void blob_resize(blob *b, size_t size);
```
Nopey: I think it's worth stating explicitly that this method (and blob_set) are both *exact* resizes-- they are just as willing to shrink the buffer as grow it, and no "spare capacity" is kept around (unlike e.g. C++ std::vector::resize or Rust std::vec::Vec::resize).

Behavior if size is 0 is again unspecified.

src/utils/blob.h
```diff
@@ -0,0 +1,43 @@
...
+ * @param src Data to set
+ * @param len Length of data
+ */
+void blob_set(blob *b, const char *src, size_t len);
+
+/**
+ * Free the blob object and its contents
+ * @param b Allocated blob object
+ */
+void blob_free(blob *b);
```
Nopey: Worth mentioning that calling this method repeatedly triggers a use-after-free (unlike most of our other free methods, such as vector_free which is idempotent).

src/utils/path.h
```diff
@@ -0,0 +1,198 @@
...
+#include "utils/iterator.h"
+#include "utils/str.h"
+
+#define PATH_MAX_LENGTH 1024
+
+#define PATH_NARGS(...) (sizeof((const char *[]){__VA_ARGS__}) / sizeof(const char *))
+
+typedef struct path {
+    char buf[PATH_MAX_LENGTH];
+} path;
```
Nopey, line 17: this struct is huge. stack-allocating large structs can trigger stack overflow.

I think it's worth mentioning what bytes can and cannot appear in a path:
My understanding is that paths can be any non-NUL sequence of bytes, and we are treating both `/` and `\` as directory seperators (if someone on linux installs the game to `/home/Nopey/fun\\times/openomf` it's going to misbehave and access nonexistent path `/home/Nopey/fun/times/openomf`-- and this misbehavior is known, expected, and accepted.

Might be worth mentioning that users of this struct should not access field buf directly, but instead use `path_c`.

src/utils/str.h
```diff
@@ -330,6 +330,14 @@ const char *str_c(const str *src);
...
 void str_split(vector *dst, const str *src, char ch);
 
+/**
+ * Join string pieces into one
+ * @param dst Destination string to be filled with the joined pieces (Do not preallocate)
+ * @param src Source vector full of initialized str's
+ * @param ch Join point character
+ */
+void str_join(str *dst, const vector *src, char ch);
+
```
Nopey: unused, no tests.

I assume this was written with the thought that `path` might be implemented in terms of `str`?

src/utils/path.h
Nopey: Could we define path as containing a `str` instead?
In path.c, we're already expecting the bytes contained in `path`s to be valid `str`s.

src/utils/str.h
```diff
@@ -338,4 +346,37 @@ void str_split(vector *dst, const str *src, char ch);
...
+bool str_match(const str *test, const char *pattern);
+
+/**
+ * Check if str matches a given glob pattern. Note! Only a single \* supported!
+ * This is the case-insensitive version of str_match().
+ * @param test String to test
+ * @param pattern Glob pattern, e.g. 'path/\*.CHR'
+ * @return True if matched, false if not.
+ */
+bool str_imatch(const str *test, const char *pattern);
```
Nopey, line 372: Please specify what kind of case insensitivity this function performs.
This method's behavior currently changes with the locale.

Is the str and pattern expected to be UTF-8?
Does [0xb4 == 0xb8](https://en.cppreference.com/w/c/string/byte/tolower#Example)?

Nopey: On a tangent, we should probably document what bytes our `str` is expected to contain.

I believe the current situation is that they're sometimes UTF-8, sometimes CP437, and perfectly capable of handling arbitrary non-NUL byte sequences (e.g. file paths).

src/utils/path.h
```diff
@@ -0,0 +1,198 @@
...
+ * @param p Directory path to delete
+ * @return True if operation succeeded, false if not.
+ */
+bool path_rmdir(const path *p);
+
+/**
+ * Find files from the path using a glob pattern
+ * Note! Only a single '*' supported!
+ * @param dir Directory path to scan
+ * @param results Result list. Note! This must be pre-allocated using list_create() !
```
Nopey: Does it append to results, or is the list cleared (but not created)?

src/utils/path.c
katajakasa: Yep, fixed

src/utils/blob.h
katajakasa: Right. I forgot to remove it. I had some plans on using it, but ended up taking a completely different path. Removed it now.

src/utils/blob.h
katajakasa: Removed completely.

src/utils/blob.h
katajakasa: Removed completely.

src/utils/blob.h
katajakasa: Removed completely.

src/utils/blob.h
katajakasa: Removed completely.

src/utils/path.h
katajakasa: We can probably reduce the size of the path to 256, actually. I did consider using str, but I specifically wanted to have something that we don't need to deallocate (since we are already doing path operations with snprintf on large stack buffers)

src/utils/str.h
katajakasa: Oh right, forgot to remove it. Yeah, I had plans for it originally, and then those did not end up happening :) Removed now.

src/utils/str.h
katajakasa: Ideally, str should be able to handle utf-8. Currently I just avoid the whole question here, since it will require some considerable effort to do properly. So probably something for the future. Right now I just expect all paths to be 8-bit encoding and if they're not, then we just hope stuff accidentally works.

src/utils/path.h
katajakasa: Clarified.

src/utils/path.h
katajakasa: I should clarify -- I am fine with using str in the path and hiding the contents and returning pointers in path initializers, if people are fine with handling deallocation constantly :)

Nopey: "256 bytes should be long enough for anyone's openomf installation/config/etc path"
you can quote me in 18 months when this breaks

katajakasa: Windows path limit is 256, so I sorta picked it up from there.

Still UTF-8 was mentioned... would it just be better to mask str as a path instead? Would it be ergonomic to use ? It /would/ give us UTF-8 when we implement it.

src/utils/str.h
Nopey: If we're keeping track somewhere (doc comment in path.h?), our filepaths are UTF-8 on Windows 10+ (because we specify compile option `/utf-8` in CMakeLists.txt).

I don't think we care about Windows 8.1 & earlier, so I guess all we need to watch out for is unusual Linux installs and paths?
Having the program error out if a path is not UTF-8 seems acceptable to me, as long as we display a meaningful error message to the user (call err_msgbox? if we want to include the faulty path, replace control codes and invalid utf-8 with substitution character U+FFFD).

src/utils/path.c
```diff
@@ -0,0 +1,389 @@
...
+#define ENSURE_ZERO(str) str[PATH_MAX_LENGTH - 1] = '\0'
+
+static void normalize_slashes(path *p) {
+    char *str = p->buf;
+    while(*str != '\0') {
+        if(*str == '\\') {
+            *str = '/';
+        }
+        str++;
+    }
```
Nopey, line 29: Let's not convert `\` to `/` on Linux.
Backslashes are valid within path components there.

As prior work, std::filesystem::path::make_preferred will convert `/` to `\` on Windows, but will not vice-versa on Linux.

Converting `/` to `\` on Windows is largely unnecessary, but it's an aesthetic touch if we need to display a path to the user (I'm neutral on it)

CMakeLists.txt
```diff
@@ -48,6 +48,8 @@ if(MSVC)
     # remove cmake's default /W3 to avoid warning D9025
     string(REGEX REPLACE "/W3" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /permissive- /W4 /WX")
+    # Add full path to __FILE__ macro (for unit-tests)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /FC")
```
Nopey: Please provide the unit-tests with the needed filepath via a macro in cmake instead of passing /FC:
```cmake
set(MY_TESTING_DIR testing)
cmake_path(ABSOLUTE_PATH MY_TESTING_DIR NORMALIZE OUTPUT_VARIABLE MY_TESTING_DIR)
target_compile_definitions(whatever_target PRIVATE "MY_TESTING_DIR=${MY_TESTING_DIR}")
```

I'd rather diagnostic messages be more brief and not specific to the build environment-- consider a message from `src\utils\path.c` vs `D:\a\openomf\openomf\src\utils\path.c`.

src/utils/path.c
```diff
@@ -0,0 +1,389 @@
...
+}
+
+bool path_touch(const path *path) {
+    FILE *fp = fopen(path->buf, "w");
+    if(fp == NULL) {
+        return false;
+    }
+    fclose(fp);
+    return true;
+}
```
Nopey: This `path_touch` is very unlike touch(1):
- It empties the file
- It cannot operate on directories.

path_touch is currently unused, and barely tested.
Perhaps it should be removed or renamed?

src/utils/path.c
```diff
@@ -0,0 +1,389 @@
...
+
+bool path_unlink(const path *p) {
+#if defined(_WIN32) || defined(WIN32)
+    const path tmp = convert_to_windows_slashes(p);
+    return DeleteFile(tmp.buf) != 0;
+#else
+    return unlink(p->buf) == 0;
+#endif
+}
```
Nopey: I think it's worth documenting that path_unlink will refuse to remove a directory or folder, and that path_rmdir should be used for that instead.

src/utils/path.h
```diff
@@ -0,0 +1,197 @@
...
+ * Note! Only a single '*' supported!
+ * @param dir Directory path to scan
+ * @param results Result list. This must be preallocated, and results are appended to an existing list.
+ * @param pattern Pattern to search for, e.g. '*.TXT'.
+ * @return True if operation succeeded, false if not.
+ */
+bool path_glob(const path *dir, list *results, const char *pattern);
+
+/**
+ * Set new file extension for the file path. Old file extension (if it exists) will be replaced.
```
Nopey, line 155: What happens if a file has multiple file extensions?
e.g. `my-file.tar.gz` or `openomf.bk.backup`

katajakasa: We always get the last one. I added comments and test to cement that.

src/utils/path.c
katajakasa: I specifically mentioned "file" in the docs, but I can add reference to rmdir too.

src/utils/path.c
katajakasa: It's used in tests, which is why it was added here... It might be better to just move it as a utility function in the test file.

src/utils/str.h
katajakasa: I added the whole UTF-8 thing to my improvements ticket at https://github.com/omf2097/openomf/issues/1293 -- I think I won't touch the UTF-8 stuff here for now, since the path library should work about the same as our old stuff. But it is definitely something that needs to be looked into, both in the path side and possible translation support.

src/utils/path.c
katajakasa: Hmmh, we would need to #ifdef windows path unit-tests too.

katajakasa: I originally added the conversion because there were big phat warnings in stackoverflow that not all windows APIs support the forward slashes. But looking into the docs, it seems all the basic APIs use the same normalization steps, which means the forward slashes should be supported.

I don't know, it might be just easier to just normalize to forward slashes and then use those everywhere :/

CMakeLists.txt
katajakasa: Sure, I don't mind. I don't use MSVC anyway :D

katajakasa: Heh, looks like we /already/ had TESTS_ROOT_DIR; @mrannanj has added it at some point :P

src/utils/path.c
katajakasa: Right, renamed to fake_touch and moved it to the path test file.

src/utils/path.c
katajakasa: Decided to drop the windows slash conversion. It's a minor performance optimization to leave it as it is :)

.github/workflows/functional.yml
```diff
@@ -46,6 +46,9 @@ jobs:
     - name: Extract openomf assets
       run: unzip -j openomf-assets.zip -d build-test/resources
 
+    - name: Copy new assets
+      run: cp resources/openomf.bk build-test/resources/ && ls -la build-test/resources
+
```
mrannanj: This change is not needed, cmake already does this, see https://github.com/omf2097/openomf/pull/1298

src/formats/chr.c
```diff
@@ -71,34 +68,33 @@ int sd_chr_load(sd_chr_file *chr, const char *filename) {
...
-            trn_loaded = trn_load(&trn, chr->pilot.trn_name) == 0;
-        }
+    // Load PIC file and make a surface
+    const path players_path = get_resource_filename("PLAYERS.PIC");
+    // log_debug("trying to load players.pic from %s", path_c(&players_path));
+    sd_pic_create(&players);
+    const int ret = sd_pic_load(&players, path_c(&players_path));
+    if(ret == SD_SUCCESS) {
+        // Load player gender from PLAYERS.PIC
+        // log_debug("loading %d from players.pic", chr->pilot.photo_id);
```
mrannanj: Would be good to remove the leftover commented out log_debugs

src/formats/pilot.c
```diff
@@ -340,17 +340,15 @@ void sd_pilot_set_player_color(sd_pilot *pilot, player_color index, uint8_t colo
...
+        // log_debug("trying to load players.pic from %s", path_c(&players_filename));
+        //  Load PIC file and make a surface
         sd_pic_file players;
         sd_pic_create(&players);
-        int ret = sd_pic_load(&players, players_filename);
+        int ret = sd_pic_load(&players, path_c(&players_filename));
         if(ret == SD_SUCCESS) {
             // Load player palette from PLAYERS.PIC
-            log_debug("loading %d from players.pic", pilot->photo_id);
+            // log_debug("loading %d from players.pic", pilot->photo_id);
```
mrannanj: Here too, commented out log_debugs

.github/workflows/functional.yml
katajakasa: Dropped

testing/test_path.c
```diff
@@ -0,0 +1,469 @@
...
+    path p;
+    path_from_c(&p, "/home/openomf");
+    path_append(&p, ".local", "share", "test.ogg");
+    CU_ASSERT_STRING_EQUAL(p.buf, "/home/openomf/.local/share/test.ogg");
+}
+
+#define ADD_TEST(desc, fn)                                                                                             \
+    if(CU_add_test(suite, desc, fn) == NULL) {                                                                         \
+        return;                                                                                                        \
+    }
```
mrannanj: This seems a macro that could be used more widely, so could move it somewhere where it can be included by other tests.

src/utils/path.c
```diff
@@ -0,0 +1,368 @@
...
+    path_from_str(path, &tmp);
+    ENSURE_ZERO(path->buf);
+    str_free(&tmp);
+}
+
+void path_dossify_filename(path *path) {
+    str tmp;
+    size_t pos = 0;
+    str_from_c(&tmp, path->buf);
+    str_last_of(&tmp, '/', &pos);
```
mrannanj, line 205: If '/' is not found in there, this would return false and pos would be uninitialized, right? So could add assert or something.

katajakasa: str_last_of does not change pos if character is not found, therefore pos will remain 0, and we assume the path is a filename without the parent component. Therefore we transform the whole path.

mrannanj: Oh, okay, fine.

mrannanj (dismissed):

Nopey (approved):

PR #1180, opened 2025-04-10
Fix versioncmake

Nopey:
i broke it, working to fix it in this pr.

image

i thought i had VERSION_REQUIRED on in CI, but didn't-- hence why this failed silently.

I'm guessing that it's failing to fetch the tags?
EDIT: Yes, failing to fetch the tags.. by thinking it already had them! the third commit is the actual fix.


Nopey:
Merging without approval to fix the "latest" release versioning.

PR #1182, opened 2025-04-11
Fixed Fire Pit Orb

### LINKED ISSUE #1163: [GAMEPLAY] Fire Pit Orb not hittable
**Steps to reproduce**
- Try to hit the fire pit orb
- Nothing will happen

With the recent change to make Pyros's flames not hittable, any hit detection involving palette indices not in HAR range stopped working.  May need to check whether the object being hit is a fire pit orb or not

LegendaryBlueShirt:
- Orb movement is now source accurate
- fixes #1163 
- Removed bespoke orbit logic


src/game/protos/player.c
```diff
@@ -653,6 +647,28 @@ void player_run(object *obj) {
         }
     }
 
+    // Orb meandering.  Only the fire pit orb should have this tag set.
+    if(sd_script_isset(frame, "as")) {
+        double t = sinf(((obj->orb_val & 7) + 8.f) * ((abs(obj->orb_val * 4) + obj->gs->tick) * 0.003574533) +
```
katajakasa, line 652: formatter really mangles this -- could you split these into a couple of lines with a helper variable or something ? Would be easier to read.

katajakasa: Where do the magic values come from, btw ?

LegendaryBlueShirt: I wouldn't be surprised if there's some mathematics book somewhere in Rob's attic with this curve formula.

LegendaryBlueShirt: https://github.com/omf2097/omfsrc/blob/d3e60c230fd9c7e4902cadfc3735f4d1b396c9fa/backdrop.cpp#L880

katajakasa (approved):
I see more red then green, I like.

Squash!

PR #1191, opened 2025-04-13
Fix the second shadow clone's position during scrap

LegendaryBlueShirt:


src/game/protos/player.c
```diff
@@ -409,12 +410,6 @@ void player_run(object *obj) {
             float vx = 0;
             float vy = 0;
```
katajakasa, line 412: Does anything still need the shadow_corner_hack  ?

LegendaryBlueShirt: Yes, it primarily drives correct behavior when Shadow is propped up against the wall since I have not yet rewritten the way the `am` tag works.

Vagabond (approved):

PR #1193, opened 2025-04-14
Fixes for Pyros's Destruction

LegendaryBlueShirt:
- Implemented the `cl` tag
- `h` tag now actually applies every frame in order to zero out the velocity when it expires
- Alternate gravity now applies to both hars when the round is ending
- Locked out all moves from normal execution when the constrain bit is active, no longer excluding scrap and destruction


src/game/objects/har.c
```diff
@@ -582,136 +580,151 @@ void har_move(object *obj) {
...
-            har_event_land(h, ctrl);
-            har_floor_landing_effects(obj, true);
-        } else if(h->state == STATE_JUMPING && enemy_har->is_grabbed == 0) {
-            // Change animation from jump to walk or idle,
-            // depending on held inputs
-            if(last_input == '6') {
-                h->state = STATE_WALKTO;
-                har_set_ani(obj, ANIM_WALKING, 1);
+            char last_input = get_last_input(h);
+            if(player_frame_isset(obj, "cl")) {
```
Vagabond, line 590: So does CL mean "chain to next animation on landing"?

src/game/objects/har.c
```diff
@@ -1960,12 +1973,12 @@ bool is_move_chain_allowed(object *obj, af_move *move) {
                 }
                 break;
             case CAT_SCRAP:
-                if(player_frame_isset(obj, "jf") && h->state != STATE_DONE) {
+                if(player_frame_isset(obj, "jf") && h->state != STATE_DONE && allowed_in_idle) {
```
Vagabond, line 1976: Can you elaborate on how/why this helps?

LegendaryBlueShirt: Without this restriction the second stage of Pyros's scrap can be performed immediately from the victory pose, which we don't want.

Vagabond: So the JF tag is set on the frame, but we;re not supposed to let it chain yet?

Vagabond (approved):

PR #1196, opened 2025-04-14
Reimplement Air Turnaround

LegendaryBlueShirt:


LegendaryBlueShirt:
I'm in a bind now because the KICK-NOTURN test cannot possibly succeed even if I modify it due to how our input interpreter works.

Vagabond:
> I'm in a bind now because the KICK-NOTURN test cannot possibly succeed even if I modify it due to how our input interpreter works.

We probably need to fix this sooner rather than later anyway, I think it's the cause of at least a couple bugs.

PR #1200, opened 2025-04-15
Release log level

mrannanj:


src/main.c
```diff
@@ -189,6 +189,7 @@ int main(int argc, char *argv[]) {
     log_add_stderr(LOG_DEBUG, true);
     log_set_level(LOG_DEBUG);
 #else
+    log_add_stdout(LOG_INFO, true);
```
katajakasa, line 192: stdout is for data output, stderr is for logging messages.

https://en.wikipedia.org/wiki/Standard_streams

mrannanj: Okay, I didn't realize that release build already logs to the log file.

PR #1205, opened 2025-04-17
Netcode refactoring and improvement

Vagabond:
This PR does several things:

* Reworks the rewind/replay loop to be tick based, not event based (which allows for the code to be simpler)
* Avoids conditions where the remote tick could get ahead of the local tick and cause desyncs
* Sends the tick of the cloned game state over the wire (mostly for lobby visibility)
* Remove logspam


src/controller/net_controller.c
```diff
@@ -793,7 +762,7 @@ int net_controller_tick(controller *ctrl, uint32_t ticks0, ctrl_event **ev) {
                             }
                         }
 
-                        for(size_t i = 18; i < event.packet->dataLength;) {
+                        for(size_t i = 18 + 4; i < event.packet->dataLength;) {
```
katajakasa, line 765: What is this + 4 ?

src/controller/net_controller.c
```diff
@@ -761,7 +731,6 @@ int net_controller_tick(controller *ctrl, uint32_t ticks0, ctrl_event **ev) {
         list_create(&data->transcript);
     }
 
-    int last_received = 0;
     int has_received = 0;
```
katajakasa, line 734: bool ?

src/controller/net_controller.c
```diff
@@ -1156,11 +1135,14 @@ void menu_controller_hook(controller *ctrl, int action) {
         ENetPacket *packet;
         serial_create(&ser);
         serial_write_int8(&ser, EVENT_TYPE_ACTION);
+        // blank header
```
katajakasa, line 1138: Issue with this style is that we have no field names. What is this value normally ?

src/controller/net_controller.c
Vagabond: I added a uint32 field to the header

src/controller/net_controller.c
Vagabond: look at the other places ACTION is encoded?

src/controller/net_controller.c
```diff
@@ -382,131 +374,115 @@ int rewind_and_replay(wtf *data, game_state *gs_current) {
...
-                    }
+                    event_names(buf0, ev->events[0]);
+                    event_names(buf1, ev->events[1]);
 
-                    sd_rec_insert_action(gs->rec, gs->rec->move_count, &move);
-                    k++;
+                    int sz = snprintf(buf, sizeof(buf),
+                                      "tick %d -- player 1 %s (%d) -- player 2 %s (%d) -- hash %" PRIu32 "\n", ev->tick,
+                                      buf0, ev->events[0][0], buf1, ev->events[1][0], arena_hash);
+                    SDL_RWwrite(data->trace_file, buf, sz, 1);
```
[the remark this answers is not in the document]
Vagabond: I plan to keep it around a while, although I should check it still works reliably after all this refactoring

src/controller/net_controller.c
Vagabond: This could probably be the serial reader's position instead.

Vagabond: Done.

src/controller/net_controller.c
Vagabond: Done.

src/controller/net_controller.c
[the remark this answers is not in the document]
Vagabond: Done, updated and confirmed working.

src/controller/net_controller.c
katajakasa: Sure. But my main point is that it would be nice if we had some sort of natural documentation for these. I wonder if there could be some sort of 
```
send_action(action_dto dto) {
    serial ser;
    serial_write_xxx(ser, dto->field1);
    ....
    send(ser);
}
```

... or something.

katajakasa (dismissed):

src/controller/net_controller.c
Vagabond: I should really just have 2 kinds of event messages.

PR #1212, opened 2025-04-19
Projectile behavior fixes

LegendaryBlueShirt:
- Fix shredder hands not doing multiple hits
- Fix shadow punch freezing the opponent before it hits
- Fix electra's projectile not being controllable

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -536,8 +536,8 @@ void har_move(object *obj) {
         return;
     }
 
-    obj->pos.x += obj->vel.x;
-    obj->pos.y += obj->vel.y;
+    obj->pos.x += (obj->vel.x + obj->cvel.x);
+    obj->pos.y += (obj->vel.y + obj->cvel.y);
 
     object *enemy_obj =
         game_state_find_object(obj->gs, game_player_get_har_obj_id(game_state_get_player(obj->gs, !h->player_id)));
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -1348,12 +1348,6 @@ void har_collide_with_projectile(object *o_har, object *o_pjt) {
         return;
     }
 
-    // rehit mode is off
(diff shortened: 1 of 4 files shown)


src/game/objects/har.c
```diff
@@ -1368,7 +1362,13 @@ void har_collide_with_projectile(object *o_har, object *o_pjt) {
     // Check for collisions by sprite collision points
     int level = 2;
     vec2i hit_coord;
+    if(o_pjt->can_hit) {
```
Vagabond, line 1365: I know this is already done in HAR collisions, but what is is actually doing?

src/game/objects/har.c
```diff
@@ -536,8 +536,8 @@ void har_move(object *obj) {
         return;
     }
 
-    obj->pos.x += obj->vel.x;
-    obj->pos.y += obj->vel.y;
+    obj->pos.x += (obj->vel.x + obj->cvel.x);
```
Vagabond, line 539: When does the HAR have cx/cy tags set on itself?

Vagabond:
Shadow slides still seem to have trouble hitting with this branch? it makes the rectest I have even less likely to hit?

src/game/objects/har.c
LegendaryBlueShirt: Flail's spinning chains

src/game/objects/har.c
LegendaryBlueShirt: I am copying the q tag handling

Vagabond:
Perhaps we should merge this as it is an improvement in several ways?

LegendaryBlueShirt:
> Perhaps we should merge this as it is an improvement in several ways?

I would like to.  I don't know that any of the changes I made affect shadow slide, which is weird, but I can't discount it.

Vagabond:
My proposal is we merge this and address anything else as a separate PR

Vagabond (approved):

PR #1213, opened 2025-04-19
Implement Scrape Rendering

### LINKED ISSUE #778: Plugs complaint scene does not show scrapes and dents on HAR
If you get beaten, your HAR should show it.

LegendaryBlueShirt:
fixes #778

### DIFF
--- a/src/game/protos/object.h
+++ b/src/game/protos/object.h
@@ -110,6 +110,7 @@ struct object_t {
     int cur_sprite_id;
     const char *sound_translation_table;
     uint8_t sprite_override; //< Tells whether cur_sprite should be kept constant regardless of anim string.
+    bool surface_override; // Tell if we've messed with the sprite and shouldn't touch it
 
     // 0 if this object is not attached to any other objects
     // non zero object id if it is. In this case, velocity and direction will be matched.
--- a/src/game/protos/object.h
+++ b/src/game/protos/object.h
@@ -110,7 +110,6 @@ struct object_t {
     int cur_sprite_id;
     const char *sound_translation_table;
     uint8_t sprite_override; //< Tells whether cur_sprite should be kept constant regardless of anim string.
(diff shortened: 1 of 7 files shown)


src/game/protos/object.h
```diff
@@ -110,6 +110,7 @@ struct object_t {
     int cur_sprite_id;
     const char *sound_translation_table;
     uint8_t sprite_override; //< Tells whether cur_sprite should be kept constant regardless of anim string.
+    bool surface_override; // Tell if we've messed with the sprite and shouldn't touch it
```
Vagabond, line 113:
```suggestion
    bool surface_override; //< Tell if we've messed with the sprite and shouldn't touch it
```

LegendaryBlueShirt: This doesn't work anyway, will need a new strategy to implement this feature

src/game/scenes/vs.c
```diff
@@ -624,6 +624,19 @@ int vs_create(scene *scene) {
...
+        ani = &bk_get_info(scene->bk_data, 9)->ani;
+
+        int n_scrapes = ((fight_stats->max_hp - fight_stats->hp) * 170) / fight_stats->max_hp;
+        log_debug("Creating %d scrapes", n_scrapes);
+        int scrape_no;
+        for(int n = 0; n < n_scrapes; n++) {
+            scrape_no = rand_int(animation_get_sprite_count(ani));
+            surface_multiply_decal(har_sprite->data, animation_get_sprite(ani, scrape_no)->data, rand_int(160), rand_int(120));
+        }
+        player1_har->cur_surface = har_sprite->data;
```
katajakasa, line 639: This is the wrong way to do it. Rendering should be all done in renderer.

katajakasa: As mentioned in the chat -- it would be preferred if this was done in the renderer. However, since this is done only once when loading the HAR, it's probably fine.

katajakasa (approved):

PR #1218, opened 2025-04-20
Fix version.cmake non-prerelease tag matching

Nopey:
As Vagabond wrote in Discord:
>  I think the regexp on line 154 of version,cmake is ignoring tags without a label

Now it successfully matches `0.8.1` (and thus labels `master` as a 0.8.2 prerelease)


Nopey:
Closed in favor of #1217

PR #1226, opened 2025-04-20
Create a cache for the 'shadow trail' animation effects

Vagabond:
Currently we are creating a new animation and sprite for every single frame that needs a trail effect. Netcode rollbacks can cause this to overflow the atlas. This patch changes things so that each potential sprite that needs the trail effect only creates one animation, that is cached and reused (along with the associated sprite).


LegendaryBlueShirt (approved):
Works under stress test conditions offline.

PR #1227, opened 2025-04-21
Fallback to doing git config without --comment for compatability

Vagabond:


cmake-scripts/version.cmake
```diff
@@ -110,8 +110,17 @@ endif()
 execute_process(
     COMMAND "${GIT_EXECUTABLE}" config "--comment" "set by version.cmake, to sort prerelease versions as older than releases." "versionsort.suffix" "-"
     OUTPUT_QUIET ERROR_QUIET
+    RESULT_VARIABLE "GIT_ERRORCODE"
 )
 
+if(GIT_ERRORCODE)
+    # retry without the --comment argument, which was added in git 2.45, released april 2024
```
mrannanj, line 117: Why not make this the only way, if the git needed is that new?

Nopey: the purpose of the comment is to direct a human towards this file, in case anyone ever wonders "where did this git config entry come from? i don't remember adding it!"

removing the comment from the git config.. removes the breadcrumb trail

mrannanj: Well, okay. Could maybe put the capitals correct in that comment, like start Retry and April in big letter.

mrannanj (approved):

PR #1229, opened 2025-04-22
Spectator mode

Vagabond:
Still has some problems

- [x]  Wrong pilot portraits shown on VS screen
- [x] Random seed is not sent to the spectator
- [x] No way to exit spectator mode yet

### DIFF
--- a/src/controller/controller.h
+++ b/src/controller/controller.h
@@ -29,7 +29,8 @@ enum
     CTRL_TYPE_GAMEPAD,
     CTRL_TYPE_NETWORK,
     CTRL_TYPE_AI,
-    CTRL_TYPE_REC
+    CTRL_TYPE_REC,
+    CTRL_TYPE_SPECTATOR
 };
 
 // TODO this should be reduced to only ACTION and CLOSE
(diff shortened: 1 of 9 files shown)


src/formats/script.c
```diff
@@ -488,8 +488,8 @@ const sd_script_tag *sd_script_get_tag(const sd_script_frame *frame, const char
     }
     iterator it;
     sd_script_tag *now;
-    vector_iter_begin(&frame->tags, &it);
-    foreach(it, now) {
+    vector_iter_end(&frame->tags, &it);
```
Vagabond, line 491: This was added accidentally, remove.

src/utils/iterator.h
```diff
@@ -18,5 +18,6 @@ void *iter_peek(iterator *iterator);
 void *iter_prev(iterator *iterator);
 
 #define foreach(iterator, item) while((item = iter_next(&iterator)) != NULL)
+#define foreach_reverse(iterator, item) while((item = iter_prev(&iterator)) != NULL)
```
Vagabond, line 21: Added accidentally.

PR #1234, opened 2025-04-24
Implemented K Tag

LegendaryBlueShirt:
- Removed damage pre-calculation since K needs to be evaluated first
- Rectest pending another anim string bugfix


Vagabond:
Having a rectest added here would help develop the fix.

Vagabond (approved):

PR #1239, opened 2025-04-26
Fix Turnaround Glitch

LegendaryBlueShirt:
Also getting some cycles back by skipping the first 14 animations

KICK-NOTURN is testing a particular input quirk that we are not emulating from the original game yet.  It was passing before for bad reasons that are now clear to me.


src/game/objects/har.c
```diff
@@ -2101,7 +2101,7 @@ af_move *match_move(object *obj, char prefix, char *inputs) {
     af_move *move = NULL;
     size_t len;
 
-    for(int i = 0; i < 70; i++) {
+    for(int i = 15; i < 70; i++) {
```
Vagabond, line 2104: Can you use ANIM_SCREW or something here?

LegendaryBlueShirt: Sure can!

LegendaryBlueShirt:
Opened issue #1240 to track a future fix

Vagabond (approved):

PR #1241, opened 2025-04-26
Don't flip inputs in the input buffer, only when being checked

Vagabond:


LegendaryBlueShirt (approved):
Works for me, crossing my fingers that this fixes some netplay issues as well.

PR #1247, opened 2025-04-27
Fix fire orb random to not desync netplay

Vagabond:
Also, this adds the random seed to the arena hash so it's easier to detect a desync like this. Also adds some logging improvements for debugging.

### DIFF
--- a/src/game/game_state.c
+++ b/src/game/game_state.c
@@ -1011,10 +1011,6 @@ void game_state_static_tick(game_state *gs, bool replay) {
 // This function is called when the game speed requires it
 void game_state_dynamic_tick(game_state *gs, bool replay) {
 
-    if(gs->hit_pause > 0) {
-        gs->hit_pause--;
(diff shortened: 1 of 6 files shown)


LegendaryBlueShirt (approved):

src/game/game_state.c
```diff
@@ -1049,8 +1045,12 @@ void game_state_dynamic_tick(game_state *gs, bool replay) {
         game_state_dyntick_controllers(gs);
     }
 
-    // Tick scene
-    scene_dynamic_tick(gs->sc, game_state_is_paused(gs));
+    if(gs->hit_pause > 0) {
+        gs->hit_pause--;
+    } else {
+        // Tick scene
+        scene_dynamic_tick(gs->sc, game_state_is_paused(gs));
```
katajakasa, line 1052: should this check be in game_State_is_paused instead ?

PR #1250, opened 2025-04-27
Select unranked challenger with a weighted sort

mrannanj:


src/game/scenes/newsroom.c
```diff
@@ -346,6 +346,24 @@ void newsroom_startup(scene *scene, int id, int *m_load, int *m_repeat) {
     }
 }
 
+static int32_t challenger_score(sd_pilot *p, game_player *p1, game_player *p2, fight_stats *fight_stats) {
```
katajakasa, line 349: Could you add a comment explaining what this function does ? The bitshift logic and checks are not obvious, I think.

mrannanj: Yeah, I can add a comment, it's making a bitset so the conditions can be sorted for importance.

Vagabond (approved):

PR #1253, opened 2025-04-28
Rehit Mode

LegendaryBlueShirt:
- Y velocity on air hits seems wrong, will investigate later
- As of yet haven't validated the rehit combo damage scaling
- This logic also takes care of invulnerability when recovering from air hits in non-rehit mode.  As long as a player doesn't press any buttons they should be able to land safely.


src/game/objects/har.c
```diff
@@ -1271,27 +1277,23 @@ int har_collide_with_har(object *obj_a, object *obj_b, int loop) {
...
+                return 0;
+            }
 
-        // rehit mode is on, but the opponent isn't airborne or stunned
-        if(obj_b->gs->match_settings.rehit && b->state == STATE_RECOIL &&
-           (!object_is_airborne(obj_b) || b->endurance < 0)) {
-            return 0;
+            // check this mode hasn't already rehit
+            if(strchr(b->rehits, move->id)) {
+                log_debug("move %d has already done a rehit");
```
Vagabond, line 1288: Missing printf argument here

Vagabond (approved):

PR #1256, opened 2025-04-29
Throws can no longer be started against opponents getting up

LegendaryBlueShirt:


Vagabond:
Feels like a rectest here would be wise

PR #1257, opened 2025-04-29
Players cannot block while attacking

LegendaryBlueShirt:
Apparently players who are holding down/back are still counted as crouchblocking when attacking


Vagabond:
The war on har->state continues

PR #1258, opened 2025-04-29
Hit mechanics overhaul

LegendaryBlueShirt:
Using a launcher move (ai tag) to stun an enemy shouldn't cause them to get up while airborne anymore.
Launcher hits don't apply rehit damage scaling anymore
Reversed the rehit damage scaling, turns out rehits do more damage not less
Hit velocities now calculated using move base damage, not adjusted damage

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -1064,19 +1064,31 @@ void har_take_damage(object *obj, af_move *move) {
 // for scrap, nuts, bolts, and sparks (aka burning oil)
 static vec2f har_debris_random_vel(object *har_obj, bool is_destruction) {
     vec2f vel;
-    vel.x = rand_float() * 4.0f - 2.0f;
-    vel.y = rand_float() * 4.0f - 2.0f;
+    vel.x = (rand_float() * 4.0f) - 2.0f;
+    vel.y = (rand_float() * 4.0f) - 2.0f;
 
     // FIXME: burning oil/sparks in destruction and scrap animations
     // FIXME: how does the original game handle scrap velocity in destruction anims?
     if(is_destruction) {
+        int x = rand_int(1);
+        // 50/50 directional split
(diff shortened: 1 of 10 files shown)


src/game/objects/har.c
```diff
@@ -941,6 +941,12 @@ void har_take_damage(object *obj, const str *string, float damage, float stun) {
         if(h->throw_duration) {
             // No special handling
             object_set_stride(obj, 1);
+        } else if(player_frame_isset(other_har, "ai")) {
```
Vagabond, line 944: What happens if the HAR was hit by a projectile while the opponent missed an attack with this flag?

LegendaryBlueShirt: I need to propagate the ai flag some other way then, you're right it would cause an issue

LegendaryBlueShirt:
Bump for visibility, there are drastic changes since the last comment

src/game/objects/har.c
```diff
@@ -937,28 +1004,29 @@ void har_take_damage(object *obj, const str *string, float damage, float stun) {
...
             log_debug("airborne knockback");
-            // append the 'airborne knockback' string to the hit string, replacing the final frame
-            size_t last_line = 0;
-            if(!str_last_of(string, '-', &last_line)) {
-                last_line = 0;
-            }
-
-            str_from_slice(&custom, string, 0, last_line);
+            // append the 'airborne knockback' string to the hit string
+            str_from(&custom, string);
```
Vagabond, line 1016: This replaces, not appends, right?

LegendaryBlueShirt: The previous behavior was looking at the last frame of the hitstring and replacing it with a small suffix.  I have removed that logic and the suffix is simply appended, but that's further down.  There's an existing comment that I'm amending here

Vagabond (approved):

PR #1263, opened 2025-04-30
Fix Enhancement String Selector

LegendaryBlueShirt:
- Enhancements now work in all game modes
- Enhancements won't override hyper/normal mode selection

### DIFF
--- a/src/game/objects/har.c
+++ b/src/game/objects/har.c
@@ -2797,49 +2797,41 @@ int har_create(object *obj, af *af_data, int dir, int har_id, int pilot_id, int 
     for(int i = 0; i < MAX_AF_MOVES; i++) {
         move = af_get_move(af_data, i);
         if(move != NULL) {
-            if(!is_tournament(obj->gs)) {
-                // projectiles have hyper mode, but may have extra_string_selector of 0
-                if(move->ani.extra_string_count > 0 && move->extra_string_selector != 1 &&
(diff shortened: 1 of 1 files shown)


src/game/objects/har.c
```diff
@@ -2797,49 +2797,41 @@ int har_create(object *obj, af *af_data, int dir, int har_id, int pilot_id, int
...
+               move->extra_string_selector != 2) {
+                // if you have 1 enhancement choose extra string 2
+                // if you have 2 enhancements choose extra string 3
+                // if you have 3 enhancements choose extra string 4
+                // if that string doesn't exist, pick the highest one
+                if(pilot->enhancements[har_id] > 0) {
+                    // find the last populated enhancement index, or 1 (hyper)
+                    int enhancement_str = min2(1 + pilot->enhancements[har_id], move->ani.extra_string_count - 1);
+                    // don't override the hyper mode switch
+                    if(enhancement_str >= 2) {
```
Vagabond, line 2814: What if hyper mode is off, but I have a level 1 enhancement?

LegendaryBlueShirt: That would be string 2 as noted in the comment above these lines

Vagabond: ah, I see.

Vagabond (approved):

PR #1271, opened 2025-08-31
Add windows unit-tests

katajakasa:
Stab at msvc so that we can run our unit-test also on windows-platform. This is so that we can test windows AND linux code-paths.

### DIFF
--- a/.github/actions/build_msvc/action.yaml
+++ b/.github/actions/build_msvc/action.yaml
@@ -13,32 +13,25 @@ runs:
     - name: Get OpenOMF Version
       uses: ./.github/actions/version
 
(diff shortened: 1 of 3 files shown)


Nopey (approved):
lgtm.

i pushed a [few minor changes](https://github.com/omf2097/openomf/compare/e618a892d5c0e2ec1a6720a30b3e15ab5c402816..e5b8cfb082275ceff2273d64a13b5557d0044c98)

acknowledging that vcpkg caching is still broken, i'll try to fix it in another PR. EDIT: PR #1273 is now open, based on this PR, and fixes the vcpkg binary caching.

PR #1272, opened 2025-08-31
Fix MacOS CI

Nopey:


Nopey:
we have cmake 3.31 installed from github's image, so no need to update (despite the availability of cmake 4.1.1, as we don't need it)

katajakasa (approved):

PR #1273, opened 2025-09-01
Fix VCPKG binary caching

Nopey:
The johnwason/vcpkg-action@v7.0.1 action is now used, as it supports caching with reasonably few changes on our end.

RIP `VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"`, microsoft removed it.


katajakasa (approved):
nice!

PR #1276, opened 2025-09-04
Implement Lobby Match Settings

LegendaryBlueShirt:
Added a dialog to show match settings before challenging
Propagate match settings to the challenger and spectators

I made a hacky version of the show dialog function to extend the size to fit what I needed, there is obviously a better way.

The game now uses lobby names instead of pilot names during netplay

### DIFF
--- a/src/game/scenes/lobby.c
+++ b/src/game/scenes/lobby.c
@@ -130,6 +130,7 @@ typedef struct lobby_user {
     uint8_t wins;
     uint8_t losses;
     uint8_t status;
+    uint32_t opponent_id;
     match_settings match_settings;
 
     text *name_text;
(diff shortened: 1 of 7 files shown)


src/game/scenes/lobby.c
```diff
@@ -516,51 +530,91 @@ void lobby_cancel_challenge(component *c, void *userdata) {
...
+    lobby_local *local = scene_get_userdata(s);
+    if(result == DIALOG_RESULT_NO) {
+        local->mode = LOBBY_MAIN;
+    } else if(result == DIALOG_RESULT_YES_OK) {
+        lobby_do_spectate(NULL, s);
+    }
+}
 
+char *challengeStr = "Challenge %s?\n\nRounds %s\nHazards %s\nFight Mode %s\nRehit Mode %s";
+char *spectateStr = "Spectate %s?\n\nRounds %s\nHazards %s\nFight Mode %s\nRehit Mode %s";
```
Vagabond: The spectator doesn't really care what the match settings are, right?

src/game/scenes/lobby.c
```diff
@@ -1491,6 +1548,7 @@ void lobby_tick(scene *scene, int paused) {
 
                                 // force the speed to 3
                                 game_state_set_speed(gs, 10);
+                                game_state_copy_match_settings(gs, &local->opponent->match_settings);
```
Vagabond: We should do this differently, by having the lobby pack the challengee's match settings into the spectate packet stream, I think.

src/game/scenes/lobby.c
LegendaryBlueShirt: Not really, but here would be a fantastic place to show "Spectate %s vs %s?"

Vagabond:
I think the main thing I'd prefer to see is the settings that differ from your own, like if some idiot has jump height set to 400%

LegendaryBlueShirt:
Superseded by #1284

PR #1278, opened 2025-09-06
Control Chronos Teleport

LegendaryBlueShirt:
Holding left/right now teleports to the wall as expected.  If the opponent is already in the corner, whether you push them out or not is random.


Vagabond:
another case where a rectest would be good?

LegendaryBlueShirt:
> another case where a rectest would be good?

Yeah turns out random where I placed it doesn't sync and I would probably break netplay if I merged this as is so back to the drafts it goes.

src/game/protos/player.c
```diff
@@ -386,13 +386,20 @@ void player_run(object *obj) {
...
-        // set the object's X position to be behind the opponent
-
-        if(obj->pos.x > enemy->pos.x) { // From right to left
-            obj->pos.x = enemy->pos.x - object_get_size(obj).x / 2;
-        } else { // From left to right
-            obj->pos.x = enemy->pos.x + object_get_size(enemy).x / 2;
+        har *har = object_get_userdata(obj);
+        switch(har->inputs[0]) {
+            case '6':
+                obj->pos.x = ARENA_RIGHT_WALL - rand_int(30);
```
Vagabond: I believe you need to use the seeded rand: `random_int(&obj->gs->rand, 30)`

PR #1279, opened 2025-09-07
Katana Turnaround

LegendaryBlueShirt:
Katana can now turn in the air after heel stomp or wall spin hits as intended


src/game/protos/player.c
```diff
@@ -385,6 +385,28 @@ void player_run(object *obj) {
         obj->pos.y = ARENA_FLOOR;
     }
 
+    if(sd_script_isset(frame, "ad")) {
+        har *har = object_get_userdata(obj);
+        int new_facing = obj->direction;
+        switch(har->inputs[0]) {
+            case '4':
```
Vagabond, line 392: 7 and 9 won't do it too?

LegendaryBlueShirt: You are correct after testing it in DOS, will adjust.

PR #1287, opened 2025-09-10
Tune down NAT timeouts and correct some log messages

Vagabond:


LegendaryBlueShirt (approved):

LegendaryBlueShirt:
Anecdotally, it does connect much faster now.

PR #1289, opened 2025-09-11
Throw Range Bug

LegendaryBlueShirt:
Throw followups tend to have lower range than the triggering move, the smaller range is now checked.
Throw range is now measured from the thrower and not the throwee
Removed the code preventing throw start from KOing opponents in order to better follow the original game


src/game/objects/har.c
```diff
@@ -1318,7 +1327,7 @@ int har_collide_with_har(object *obj_a, object *obj_b, int loop) {
         }
     }
 
-    if(!is_in_range(obj_b, move)) { // Won't get hit by throws out of range
+    if(!is_in_range(obj_a, move)) { // Won't get hit by throws out of range
```
Vagabond: Why did this change to object a? the move is also sourced from object a?

src/game/objects/har.c
```diff
@@ -1437,12 +1446,6 @@ int har_collide_with_har(object *obj_a, object *obj_b, int loop) {
         disable_rehit(b, move);
 
         if(move->next_move) {
-            af_move *next_move = af_get_move(a->af_data, move->next_move);
```
Vagabond, line 1440: Why is this no longer necessary?

src/game/objects/har.c
LegendaryBlueShirt: I've changed the function signature now because we need both obj_a and obj_b to measure distance.  The function was previously fetching the other object which is redundant because we already had a handle for that data.

src/game/objects/har.c
LegendaryBlueShirt: Preventing the start of a throw from KOing the opponent is not accurate to the source game so I removed it.

Vagabond: a rectest for that might be wise?

PR #1298, opened 2025-09-14
run_pytest.sh: Drop OPENOMF_RESOURCE_PATH

mrannanj:
For testing what breaks without this.

### DIFF
--- a/.github/workflows/functional.yml
+++ b/.github/workflows/functional.yml
@@ -46,9 +46,6 @@ jobs:
     - name: Extract openomf assets
       run: unzip -j openomf-assets.zip -d build-test/resources
 
-    - name: Copy new assets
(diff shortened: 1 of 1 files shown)


src/resources/resource_paths.c
```diff
@@ -136,6 +136,13 @@ static bool find_resource_path(path *dst) {
         }
         str_free(&tmp);
     }
+    str_from_c(&tmp, ".");
+    // Check cwd.
+    if(scan_potential_resource_dirs(dst, &tmp, "openomf/resources/openomf.bk", "openomf/")) {
+        log_debug("Resources found in system path: %s", path_c(dst));
+        goto ok;
+    }
+    str_free(&tmp);
```
katajakasa: I don't think this belongs to the game binary -- it is only used by tests, afaiks there is no other purpose for this. I think the envvar is the way to go here.

PR #1300, opened 2025-09-15
Implement path_resolve() for resolving relative paths to absolute.

katajakasa:


src/utils/path.c
```diff
@@ -116,6 +116,27 @@ const char *path_c(const path *path) {
...
+        normalize_slashes(p);
+        return true;
+    }
+    return false;
+#else
+    char *resolved = realpath(p->buf, NULL);
+    if(resolved == NULL) {
+        return false;
+    }
+    strncpy(p->buf, resolved, PATH_MAX_LENGTH);
```
Nopey: This is a misuse of strncpy, and does not guarantee null termination.

Use one of the openomf functions, like strncpy_or_abort.
Or, calculate the length of resolved, check that it fits in PATH_MAX_LENGTH (return false otherwise), and memcpy.

src/resources/resource_paths.c
```diff
@@ -91,9 +91,14 @@ static bool scan_potential_resource_dirs(path *result, const str *src, const cha
...
     foreach(it, slice) {
         path_from_parts(&test, str_c(slice), find);
+        if(!path_resolve(&test)) {
+            log_warn("Unable to resolve %s; does it exist?", path_c(&test));
+            continue;
+        }
         log_debug("Looking for resources in %s ...", path_c(&test));
         if(path_is_file(&test)) {
             path_from_parts(result, str_c(slice), append);
+            path_resolve(result);
```
Nopey: The two calls to path_resolve is a (probably innocuous) TOCTOU bug.

If we make `find` a relative path from `append`, then we can call path_resolve once on "slice..find" and check for the existence of openomf.bk inside of that canonicalized path.

Nopey (changes requested):

src/utils/path.c
katajakasa: We could probably just use strlen chec + strcpy here, since we know resolved is always null terminated.

Nopey: that could overflow the destination buffer.

PATH_MAX is quite large, i think around 4096 bytes on common Linux configurations?

katajakasa: if(strlen(src) < MAX) strcpy() else abort()

Nopey: that's what our strncpy_or_abort function boils down to, just without calculating the string length twice.
it's defined in our `c_string_util.h`.
edit: i may not have read the "strlen +" part of two messages ago properly. still, use the c_string_util function, don't implement it ad-hoc inline

katajakasa: Oh, sure. I was just musing that there is actually a case where strcpy could be used safely.

Isn't strncpy_or_abort exactly the same though ? It does strlen and then does the copy; strlen and strcpy would do exactly the same.

katajakasa: I did edit my previous post immediately after posting it to clarify it, so you may have read the original one :)

katajakasa: Anyways, swapped to strncpy_or_abort() now.

src/resources/resource_paths.c
katajakasa: Yep. I reworked the resolve function a bit in general, should be better.

src/utils/path.c
Nopey: > Isn't strncpy_or_abort exactly the same though ? It does strlen and then does the copy; strlen and strcpy would do exactly the same.

Yes and no: Yes, the result is the same. No, it is not exactly the same. strcpy has a memory-dependency where the moves depend on the values loaded, whereas in memcpy the memory-dependency is on the copy-size argument-- and if we need to debug a failure, having a central "strncpy_or_abort" implementation to add spew, breakpoints, see in stacktraces,.. is a boon.

Nopey (dismissed):

src/utils/path.c
katajakasa: Yes, but I mean the amount of strlens. To my knowledge, in both cases we strlen once and copy once.

EDIT: In reference to "that's what our strncpy_or_abort function boils down to, just without calculating the string length twice."

PR #1302, opened 2025-09-17
More path fixes and improvements

katajakasa:
Minor cleanups and windows bug fix


LegendaryBlueShirt:
On line 97

`if(!path_resolve(base)) {`

The value of `base` is clobbered on Windows 10

PR #1307, opened 2025-09-20
Prefer crash() instead of abort()

katajakasa:
This might be nicer to use instead of abort() in some cases.

### DIFF
--- a/src/controller/rec_controller.c
+++ b/src/controller/rec_controller.c
@@ -5,6 +5,7 @@
 #include "game/game_state_type.h"
 #include "game/protos/scene.h"
 #include "utils/allocator.h"
+#include "utils/crash.h"
 #include "utils/log.h"
 #include <inttypes.h>
 
@@ -87,7 +88,7 @@ int rec_controller_poll(controller *ctrl, ctrl_event **ev) {
                 if(parse_assertion(buf, &ass)) {
                     log_assertion(&ass);
                     if(!game_state_check_assertion_is_met(&ass, ctrl->gs)) {
-                        abort();
+                        crash("REC file assert failed!");
                     }
                 }
             } else if(move->lookup_id == 96) {
--- a/src/formats/error.c
+++ b/src/formats/error.c
@@ -1,5 +1,6 @@
 #include "formats/error.h"
-#include <stdlib.h>
+
+#include <assert.h>
 
 const char *sd_get_error(enum SD_ERRORCODE error_code) {
     switch(error_code) {
@@ -30,5 +31,6 @@ const char *sd_get_error(enum SD_ERRORCODE error_code) {
         case SD_FAILURE:
             return "General failure";
     }
-    abort();
+    assert(false && "Unknown error type");
+    return "Unknown error type";
 }
--- a/src/game/game_state.c
+++ b/src/game/game_state.c
@@ -30,6 +30,7 @@
 #include "utils/allocator.h"
 #include "utils/c_array_util.h"
 #include "utils/c_string_util.h"
+#include "utils/crash.h"
 #include "utils/log.h"
 #include "utils/miscmath.h"
 #include "video/vga_state.h"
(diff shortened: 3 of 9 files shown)


src/utils/crash.h
```diff
@@ -1,7 +1,11 @@
 #ifndef CRASH_H
 #define CRASH_H
 
-void _crash(const char *message, const char *function, const char *file, int line);
+#if defined(_WIN32) || defined(WIN32)
+__declspec(noreturn) void _crash(const char *message, const char *function, const char *file, int line);
+#else
+void _crash(const char *message, const char *function, const char *file, int line) __attribute__((__noreturn__)) __COLD;
+#endif
```
Nopey: C11 has standardized noreturn, please use it instead of these pre-standardization decls.

https://en.cppreference.com/w/c/language/noreturn.html

then use an IF-GCC or similar, as __COLD has nothing to do with MSVC nor standard C.

EDIT: this will fix the MacOS CI, too-- where __COLD is currently causing trouble.

src/utils/allocator_default.h
```diff
@@ -24,7 +26,7 @@ static inline void *omf_calloc_real(size_t nmemb, size_t size, const char *file,
     if(ret != NULL)
         return ret;
     fprintf(stderr, _text_calloc_error, nmemb, size, file, line);
-    abort();
+    crash("Allocation failure");
```
Nopey: the fprintfs are already reporting the allocation failure, and **caller** file/line.
Perhaps these allocation failures could call _crash directly, and pass file and line from the caller?

might be tidy if they could pass the format string & format args to some _crash_vargs or whatever but that might be too much.

src/controller/rec_controller.c
```diff
@@ -87,7 +88,7 @@ int rec_controller_poll(controller *ctrl, ctrl_event **ev) {
                 if(parse_assertion(buf, &ass)) {
                     log_assertion(&ass);
                     if(!game_state_check_assertion_is_met(&ass, ctrl->gs)) {
-                        abort();
+                        crash("Game state check failed!");
```
Nopey: These are "REC TEST assertions", and should probably mention that instead of "game state check".

the function name isn't terribly fortunate, but it's that way for the namespacing.

src/game/game_state.c
```diff
@@ -118,39 +119,40 @@ void game_state_set_pilot_name(game_state *gs, int pilot_id, const char *pilot_n
...
+            object *obj_opp = game_state_find_object(
+                gs, game_player_get_har_obj_id(game_state_get_player(gs, abs(op->value.attr.har_id - 1))));
+            return fabsf(obj->pos.x - obj_opp->pos.x);
+            case ATTR_DIRECTION:
+                return object_get_direction(obj);
         }
+        default:
+            break;
     }
+    crash("Unknown attribute");
```
Nopey: "Unknown ATTR in REC TEST assertion", perhaps-- keeping in line with them being referred to as "ATTR"s in the help string for the ASSERT console command (i don't remember how the rec cli presents them).

src/formats/error.c
```diff
@@ -30,5 +31,5 @@ const char *sd_get_error(enum SD_ERRORCODE error_code) {
         case SD_FAILURE:
             return "General failure";
     }
-    abort();
+    crash("Unknown error type");
```
Nopey: I think this function should not crash, and should instead `assert(0)` (because there's been a programmer bug) before returning "Unknown error type" as the human-readable error name.

Nopey (changes requested):

src/utils/crash.h
katajakasa: So C11 has _NoReturn, which is superceded by noreturn attr in C23. Now I'm wondering if we should just switch to C23 directly.

src/utils/allocator_default.h
katajakasa: I was also thinking that it might make sense to pass varargs. I'll probably test it later.

src/formats/error.c
katajakasa: Yeah. I didn't have time to consider this yet, I just did a search and replace -- I was thinking that we don't really need the error types; it might be easier to just crash() in the parsers on error, and return bool false on general failure from the functions (just like the rest of openomf).

PR #1309, opened 2025-09-22
Do Not Iterate Multiple Inputs In Mechlab

### LINKED ISSUE #1303: Game crashes to desktop when entering arena.
Can reproduce quite reliably.
To reproduce:
- go to tournament play mode
- select enter tournament button
- press SHIFT and ENTER at the same time
- game crashes

LegendaryBlueShirt:
fixes #1303

### DIFF
--- a/src/game/scenes/mechlab.c
+++ b/src/game/scenes/mechlab.c
@@ -530,60 +530,59 @@ void mechlab_input_tick(scene *scene) {
     if(i == NULL) {
         return;
     }
-    do {
-        if(i->type == EVENT_TYPE_ACTION) {
-            if(local->popup) {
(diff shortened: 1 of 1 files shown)


src/game/scenes/mechlab.c
```diff
@@ -530,60 +530,59 @@ void mechlab_input_tick(scene *scene) {
     if(i == NULL) {
         return;
     }
-    do {
```
Vagabond, line 534: You could also make this a while() where we check the scene isn't changing?

Vagabond:
Perhaps a better way is to prevent switching scenes while we're switching scenes?

PR #1311, opened 2025-09-22
Fix Scoreboard

LegendaryBlueShirt:


src/game/protos/scene.c
```diff
@@ -40,10 +40,10 @@ int scene_create(scene *scene, game_state *gs, int scene_id) {
             bk_filename = get_resource_filename(get_resource_file(BK_INTRO + (scene_id - 1)));
             break;
         case SCENE_SCOREBOARD:
-            bk_filename = get_resource_filename("MENU.BK");
+            bk_filename = get_resource_filename(get_resource_file(BK_MENU));
             break;
         case SCENE_LOBBY:
-            bk_filename = get_resource_filename("NETARENA.PCX");
+            bk_filename = get_resource_filename(get_resource_file(PCX_NETARENA));
```
katajakasa: Why does this fix things ? I am specifically attempting to get rid of the get_resource_file() API.

katajakasa: Right, the filename was wrong. Anyway, I'm trying to just use static names, since the whole get_resource_file is just pointless misdirection now. So let's not use it.

LegendaryBlueShirt: okie dokie

PR #1312, opened 2025-09-22
Allow resizing intermediate framebuffer

katajakasa:
WIP

PR #1313, opened 2025-09-22
Mod support

Vagabond:
Things to do:

- [x] Require a mod manifest in each zip file
- [x] Either do not load mods in netplay mode, or load only cosmetic changes (would require retaining original sprites for compatible collision detection)
- [x] Fix all the memory leaks
- [x] Support for sprite information like collision coordinates, offsets and index reuse
- [x] Support for tournament pilot photos/logos to be higher resolution
- [x] Switch from file IDs to filenames (eg scene_id 8 -> "ARENA0", har_id 1 -> "FIGHTR1")
- [x] unify with https://www.omf2097.com/wiki/doku.php?id=openomf:openomf_mod_folder_structure

Depends on #1312 to properly display high-res assets.


src/resources/modmanager.h
```diff
@@ -0,0 +1,24 @@
+#ifndef MODMANAGER_H
+#define MODMANAGER_H
```
katajakasa, line 2: modmanager seems unnecessarily long. Maybe just prefix with "mods_" ?

src/resources/modmanager.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include "formats/chr.h"
+#include "resource_files.h"
+#include "resources/af.h"
+#include "resources/af_move.h"
+#include "resources/animation.h"
+#include "resources/bk_info.h"
+#include "utils/hashmap.h"
```
katajakasa: hashmap not used in header btw

src/resources/modmanager.c
```diff
@@ -93,6 +97,15 @@ bool modmanager_init(void) {
                                 log_warn("failed to load sprite %s", str_c(&filename));
                             }
                         }
+#ifdef OPUSFILE_FOUND
+                    } else if(strcmp(".ogg", str_c(&ext)) == 0) {
```
katajakasa: str_equals(&ext, ".ogg")

src/utils/png_reader.c
```diff
@@ -37,14 +37,14 @@ static FILE *open_and_check(const char *filename) {
 }
 
 static void read_png_data(unsigned char *dst, png_structp png_ptr, png_infop info_ptr, int w, int h) {
-    png_bytep *row_pointers = png_malloc(png_ptr, h * png_get_rowbytes(png_ptr, info_ptr));
-    png_read_image(png_ptr, row_pointers);
+    png_uint_32 row_bytes = png_get_rowbytes(png_ptr, info_ptr);
```
katajakasa: row_bytes is unused

src/game/common_defines.c
```diff
@@ -55,7 +56,7 @@ int har_get_id(const char *name) {
     int num_hars = sizeof(har_type_names) / sizeof(har_type_names[0]);
 
     for(int id = 0; id < num_hars; id++) {
-        if(strcasecmp(name, har_type_names[id]) == 0) {
+        if(omf_strncasecmp(name, har_type_names[id], strlen(har_type_names[id])) == 0) {
```
Nopey:
```diff
-        if(strcasecmp(name, har_type_names[id]) == 0) {
+        if(omf_strncasecmp(name, har_type_names[id], strlen(har_type_names[id])) == 0) {
```
these lines lines are not equivalent, as the latter is a "is `har_type_names[id]` a suffix of `name`?" test, wheres the former is an equality test.

Do not pass strlen into the length argument of strncasecmp, this is an anti-pattern.

Perhaps: `omf_strncasecmp(name, har_type_names[id], SIZE_MAX)`, if you want to compare the whole strings? You could add this as a helper method, call it "omf_strcasecmp."

Nopey: correction: someone already added the omf_strcasecmp helper method, please use it. https://github.com/omf2097/openomf/blob/5cf380f41370e8913096445b970ec8290b19cf9b/src/utils/c_string_util.c#L55

src/resources/modmanager.h
Vagabond: I kind of like it longer, as it's very explicit what is happening?

Vagabond:
I'm going to start prepping this for review, I am going to punt all the remaining tasks to future PRs, the only bug remaining I intend to fix is that there's a problem with pilot portrait alignment in tournament matches.

src/resources/modmanager.h
katajakasa: Ok.

katajakasa:
"Prevent most mod assets loading during netplay" -- most. What is allowed ?

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+            log_info("Ignoring mod %s v%s in favor of existing v%s", name, version, existing->version);
+            cfg_free(cfg);
+            return false;
+        }
+    }
+
+    // Create manifest entry
+    mod_manifest manifest;
+    memset(&manifest, 0, sizeof(mod_manifest));
+    manifest.name = strdup(name);
```
katajakasa: omf_strdup()

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+            log_warn("Unknown language '%s' or index out of bounds", lang_name);
+            continue;
+        }
+
+        char *quote = cfg_getstr(lang, "quote");
+        if(quote) {
+            if(pilot->quotes[lang_index]) {
+                free(pilot->quotes[lang_index]);
+            }
+            pilot->quotes[lang_index] = strdup(quote);
```
katajakasa: omf_strdup

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+        }
+
+        // Update name
+        char *name = cfg_getstr(lang, "name");
+        if(name) {
+            log_info("previous title was %s", locale->title);
+            if(locale->title) {
+                free(locale->title);
+            }
+            locale->title = strdup(name);
```
katajakasa: omf_strdup

you get the idea :P

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+                        // this was loaded before
+                        str_free(&filename);
+                        zip_entry_close(zip);
+                        continue;
+                    }
+
+                    unsigned long long entry_size = zip_entry_uncomp_size(zip);
+                    void *entry_buf = omf_calloc(entry_size, 1);
+                    if(zip_entry_noallocread(zip, entry_buf, entry_size) < 0) {
+                        log_warn("failed to load %s into memory", zip_entry_name(zip));
```
katajakasa, line 390: continue;

?

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+bool modmanager_get_tournament_mod(const char *tournament_name, sd_tournament_file *tourn_data) {
+    if(!mods_allowed) {
+        return false;
+    }
+    if(!tournament_name || !tourn_data) {
+        return false;
+    }
+
+    str filename;
+    str_from_format(&filename, "tournaments/%s/tournament.ini", tournament_name);
```
katajakasa, line 1608: str_free() seems to be missing before the next call to str_from_format

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+
+    str_free(&filename);
+
+    if(!result && strncmp("arena", str_c(name), 5) == 0 &&
+       ((anim_id >= 6 && anim_id <= 11) || (anim_id >= 24 && anim_id <= 27))) {
+        // TODO make sure this is an arena
+        // For arenas, check for 'common' for anim_ids 6 (round), 7 (number), 8 (you lose), 9 (you win), 10 (fight),
+        // 11 (ready), 24 (dust 1), 25 (dust 2), 26 (dust 3), 27 (match counters)
+
+        str_from_format(&filename, "scenes/common/%d/animdata.ini", anim_id);
```
katajakasa, line 1064: matching str_free() missing

src/utils/png_reader.c
```diff
@@ -87,9 +86,120 @@ bool read_paletted_png(const path *filename, unsigned char *dst) {
...
+    }
+
+    *w = png_get_image_width(png_ptr, info_ptr);
+    *h = png_get_image_height(png_ptr, info_ptr);
+    png_byte color_type = png_get_color_type(png_ptr, info_ptr);
+    png_byte bit_depth = png_get_bit_depth(png_ptr, info_ptr);
+
+    if(!allow_transparency && trans) {
+        log_error("PNG contained transparent pixels");
+        return false;
```
katajakasa, line 148: I think this leaks png_ptr and info_ptr

png_destroy_read_struct(&png_ptr, &info_ptr, NULL);

src/resources/af_loader.c
```diff
@@ -20,7 +23,7 @@ int load_af_file(af *a, int id) {
     }
 
     // Convert
-    af_create(a, &tmp);
+    af_create(a, &tmp, &fn);
```
katajakasa, line 28: str_free(fn)

src/resources/bk_loader.c
```diff
@@ -13,8 +13,11 @@ int load_bk_file(bk *b, const path *filename) {
         return 1;
     }
 
+    str fn;
+    path_stem(filename, &fn);
+
     // Convert
-    bk_create(b, &tmp);
+    bk_create(b, &tmp, &fn);
```
katajakasa, line 20: str_free(fn)

src/audio/audio.c
```diff
@@ -245,12 +239,28 @@ void audio_play_music(resource_id id) {
...
-                load_xmp_music(path_c(&music));
-                break;
-            case MUSIC_FILE_TYPE_OGG:
-                load_opus_music(path_c(&music));
+            case MUSIC_FILE_TYPE_PSM: {
+                str fn;
+                unsigned char *buf;
+                size_t len;
+                // check the modmanager here for a music mod
+                path_stem(&music, &fn);
```
katajakasa, line 247: str_free(fn)

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+}
+
+bool modmanager_parse_photo_mod(const char *buf, sd_pic_photo *photo) {
+    if(!buf || !photo) {
+        return false;
+    }
+
+    // Options for main photo settings
+    cfg_opt_t photo_opts[] = {CFG_STR("gender", NULL, CFGF_NONE), CFG_INT("width", photo->sprite->width, CFGF_NONE),
+                              CFG_INT("height", photo->sprite->width, CFGF_NONE), CFG_END()};
```
katajakasa: "height", photo->sprite->width,

copy-paste-typo ?

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+    list *l;
+    unsigned int len = 0;
+
+    bool found = false;
+
+    if(!hashmap_get_str(&mod_resources, str_c(&filename), (void **)&l, &len)) {
+        unsigned int count = list_size(l);
+        log_info("found %d music files for %s", count, name);
+        if(index >= count) {
+            log_warn("requested index %s into list of %d members", index, count);
```
katajakasa: ... index %d into list ...

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+#include <string.h>
+
+#ifdef OPUSFILE_FOUND
+#include <opusfile.h>
+#endif
+
+#include <confuse.h>
+
+#define UPDATE_FIELD_INT(field_name, struct_field, new_value)                                                          \
+    if(new_value != 0 && struct_field != new_value) {                                                                  \
```
katajakasa, line 25: btw, I guess this means that mods cannot set field values intentionally to 0, ever ? Is that okay ?

katajakasa: Same for floats, I suppose

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+            resolution_scale = 4;
+            break;
+        default:
+            resolution_scale = 8;
+            break;
+    }
+    return min2(resolution_scale, setting->video.fb_scale);
+}
+
+hashmap mod_resources;
```
katajakasa: static ?

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+        if(locale_index >= MAX_TRN_LOCALES) {
+            log_warn("Locale index %d out of bounds for language '%s'", locale_index, lang_name);
+            continue;
+        }
+
+        // Get or create the locale
+        sd_tournament_locale *locale = tourn->locales[locale_index];
+        if(!locale) {
+            // Allocate new locale if it doesn't exist
+            locale = calloc(1, sizeof(sd_tournament_locale));
```
katajakasa: omf_calloc

src/resources/modmanager.c
```diff
@@ -0,0 +1,1674 @@
...
+    str filename;
+    unsigned int len = 0;
+
+    str_from_format(&filename, "tournaments/%s/pilots/%d/pilot.png", trn_name, pilot_id);
+    str_tolower(&filename);
+    mod_asset *obuf;
+    if(!hashmap_get_str(&mod_resources, str_c(&filename), (void **)&obuf, &len)) {
+        assert(obuf->type == MOD_SPRITE);
+        log_info("found portrait for pilot %d in %s", pilot_id, trn_name);
+        // omf_free(pilot_data->photo);
```
katajakasa: leak ?

src/game/game_state.c
```diff
@@ -741,6 +742,12 @@ int game_load_new(game_state *gs, int scene_id) {
     gs->this_id = scene_id;
     gs->next_id = scene_id;
 
+    // Disable gameplay mods during netplay arena to prevent desyncs
+    if(scene_id >= SCENE_ARENA0 && scene_id <= SCENE_ARENA4) {
+        modmanager_set_allowed(game_state_get_player(gs, 0)->ctrl->type != CTRL_TYPE_NETWORK &&
+                               game_state_get_player(gs, 1)->ctrl->type != CTRL_TYPE_NETWORK);
+    }
```
katajakasa, line 751: modmanager_set_allowed(true) on else case ?

Vagabond: it would be true, the goal is to prevent mods affecting netplay

katajakasa: If we come out from the netarena, the mods are still disallowed until the first arena ?

Vagabond:
> "Prevent most mod assets loading during netplay" -- most. What is allowed ?

music and arena backgrounds right now

src/game/game_state.c
Vagabond: fixed

src/resources/modmanager.c
```diff
@@ -0,0 +1,1679 @@
...
+
+        if(lang_index == -1 || lang_index >= 10) {
+            log_warn("Unknown language '%s' or index out of bounds", lang_name);
+            continue;
+        }
+
+        char *quote = cfg_getstr(lang, "quote");
+        if(quote) {
+            if(pilot->quotes[lang_index]) {
+                free(pilot->quotes[lang_index]);
```
katajakasa: omf_free()

src/resources/modmanager.c
```diff
@@ -0,0 +1,1679 @@
...
+            tourn->locales[locale_index] = locale;
+            log_info("Created new locale for %s at index %d", lang_name, locale_index);
+        }
+
+        // Update name
+        char *name = cfg_getstr(lang, "name");
+        if(name) {
+            log_info("previous title was %s", locale->title);
+            if(locale->title) {
+                free(locale->title);
```
katajakasa: omf_free()

src/resources/modmanager.c
```diff
@@ -0,0 +1,1679 @@
...
+            }
+            locale->title = omf_strdup(name);
+            log_info("setting %s name to '%s'", lang_name, name);
+        }
+
+        // Update description
+        char *description = cfg_getstr(lang, "description");
+        if(description) {
+            if(locale->description) {
+                free(locale->description);
```
katajakasa: omf_free()

src/resources/modmanager.c
```diff
@@ -0,0 +1,1679 @@
...
+
+            // Update ending pages
+            for(int page = 0; page < 10; page++) {
+                char page_key[10];
+                snprintf(page_key, sizeof(page_key), "page%d", page + 1);
+
+                char *page_text = cfg_getstr(ending, page_key);
+                if(page_text) {
+                    if(locale->end_texts[ending_index][page]) {
+                        free(locale->end_texts[ending_index][page]);
```
katajakasa: omf_free()

src/audio/sources/opus_source.c
```diff
@@ -88,6 +88,31 @@ bool opus_load(music_source *src, int channels, int sample_rate, const char *fil
...
+    src->context = context;
+    src->set_volume = opus_set_volume;
+    src->render = opus_render;
+    src->close = opus_close;
+    return true;
+
+exit_1:
+    op_free(context->handle);
+exit_0:
+    omf_free(context);
```
katajakasa, line 113: also rb_free() here

src/utils/png_reader.c
```diff
@@ -87,9 +86,121 @@ bool read_paletted_png(const path *filename, unsigned char *dst) {
...
+
 #else // PNG_FOUND
 
-bool png_read_paletted(const char *filename, unsigned char *dst) {
+bool read_paletted_png(const path *filename, unsigned char *dst) {
+    PERROR("PNG reading is not supported in current build!");
+    return false;
+}
+
+bool read_paletted_png_from_memory(const char *filename, unsigned char *dst, int *w, int *h, bool allow_transparency,
```
katajakasa: different header, size missing

src/audio/sources/opus_source.c
katajakasa, line 195: rb_free() missing in exit_0 case (before omf_free(context) )

src/resources/animation.c
```diff
@@ -59,6 +71,42 @@ void animation_create(animation *ani, array *sprites, void *src, int id) {
...
+        vector new_coords;
+        vec2i origin;
+        vec2i sprite_offset = vec2i_create(sdani->sprites[i]->pos_x, sdani->sprites[i]->pos_y);
+        vector_create_with_size(&new_coords, sizeof(collision_coord), 0);
+        if(modmanager_get_hitcoords(type, name, id, i, &new_coords, &origin, &sprite_offset)) {
+            if(vector_size(&new_coords) > 0) {
+                iterator it;
+                int to_add = vector_size(&new_coords);
+                // Remove all original coords for this frame.
+                vector_clear(&ani->collision_coords);
```
katajakasa: this does not clear per-frame, it clears per-animation.

src/resources/modmanager.c
```diff
@@ -0,0 +1,1794 @@
...
+    unsigned int len;
+    if(hashmap_get_str(mod_registry, name, (void **)&existing, &len) == 0) {
+        // Found a mod with the same name, compare versions
+        int compare = compare_versions(version, existing->version);
+        if(compare > 0) {
+            // New mod has a higher version, update the existing entry
+            log_info("Replacing mod %s v%s with v%s", name, existing->version, version);
+
+            // Update the existing entry with new information
+            existing->load_order = load_order;
```
katajakasa, line 215: Should also update the other fields too ?

katajakasa:
## Code Review

-- By claude --

Findings beyond what's already in the inline comments. Headline: **the AF-move, BK-animation, and fighter-header INI override paths are non-functional today** — they hit a stack of three independent bugs (#1–#3 below).

### Bugs that break entire mod features

#### 1. `foreach` is iterating with the wrong type in three INI parsers

`modmanager_get_af_move`, `modmanager_get_bk_animation`, and `modmanager_get_fighter_header` declare `char *buf` and pass it straight into the parser:

```c
char *buf;
foreach(it, buf) {
    result |= modmanager_parse_af_move_mod(buf, move_data);
}
```

But the list nodes hold `mod_asset` structs (`list_append(l, buf, sizeof(mod_asset))` at lines 470 / 491). `buf` therefore points to a `mod_asset`, and the parser reads the struct's binary layout as INI text — `cfg_parse_buf` fails immediately. Compare with the correct pattern in `modmanager_get_pilot_mod` at line 1411:

```c
mod_asset *obuf;
foreach(it, obuf) {
    assert(obuf->type == MOD_BUFFER);
    result |= modmanager_parse_pilot_mod((char *)obuf->buf, pilot_data);
}
```

Affected loops: `modmanager.c:1001-1003, 1019-1022, 1051-1053, 1069-1072, 1147-1150`.

#### 2. Missing `str_tolower(&filename)` before lookup

Hashmap keys are normalized to lowercase at insertion (`modmanager_init`, line ~378). Almost every getter lowercases the lookup key — these don't:

- `modmanager_get_af_move`: lines 991, 1013
- `modmanager_get_bk_animation`: lines 1041, 1064
- `modmanager_get_fighter_header`: line 1138
- `modmanager_get_pilot_mod` (`har_color.png` only): line 1392

Names from `path_stem` are uppercase (`FIGHTR1`, `ARENA0`) and `trn_name` likewise, so those lookups silently miss every time.

#### 3. `modmanager.c:991` — `str *` passed to `%s`

```c
str_from_format(&filename, "fighters/%s/%d/animdata.ini", name, move_id);
```

`name` is `str *`; needs `str_c(name)`. The BK twin at `:1041` does it correctly.

Combined with #1 and #2, AF-move INI overrides hit three independent failures.

#### 4. `chr.c:130` — modded HAR-color palette is overwritten

`modmanager_get_pilot_mod` applies a tournament-pilot HAR color via `palette_copy(&pilot_data->palette, obuf->pal, 0, 48)` at `:1397`. chr.c then does:

```c
memcpy(&chr->enemies[i]->pilot.palette,
       &pic.photos[trn.enemies[i]->photo_id]->pal,
       sizeof(vga_palette));
```

…which blows the modded palette away with the unmodded tournament-PIC palette. Either copy from `trn.enemies[i]->palette` (which already absorbed the mod) or move this memcpy before `trn_load`.

#### 5. `menu_audio.c:124-125` — concatenated help text missing a space

```c
"Which type of game music to use."
"Whenever more than one track is available for a given scene a random one will be chosen."
```

Adjacent string literals concatenate without a separator → `"…use.Whenever…"`.

### Memory / robustness

#### 6. Single-asset `hashmap_put_str` silently overwrites without freeing

`mod_resources` was created with `hashmap_create` (no free callback). Verified in `hashmap.c:154-159` — on key collision, `hashmap_put` reallocs and memcpys over the existing value:

```c
seek->pair.value = omf_realloc(seek->pair.value, value_len);
memcpy(seek->pair.value, val, value_len);
```

The previous `mod_asset`'s inner pointers (`img.data` / `spr.data` / `pal`) are never freed. So when two mods override the same single-asset key (`scenes/arena0/background.png`, sprite PNG, hitcoord PNG), the older entry leaks. Affected `hashmap_put_str` sites: `:411, :428, :444`. The list-based paths (`.ini` / `.ogg`) avoid this by checking-then-appending; the single-asset paths don't.

Either check existence first and free/skip, or give the hashmap a free callback.

#### 7. `sd_sprite_copy` over an already-loaded sprite leaks the previous data

Latent in `sprite.c` — does `memset(dst, 0, …)` without freeing `dst->data` first. The PR adds two new triggers where the destination is already populated:

- `modmanager_get_player_pics:1758` — `players->photos[i]->sprite` already has data from `sd_pic_load`
- `modmanager_get_tournament_mod:1457` — `tourn_data->locales[0]->logo` already populated

Fix `sd_sprite_copy` to free existing `dst->data` first, or `sd_sprite_free` the destination before each copy.

#### 8. `zip_open` return is never checked (`:257, :368`)

If a zip is corrupt, removed between glob and open, or unreadable, every subsequent `zip_entry_*` and `zip_close` call dereferences NULL.

#### 9. `omf_calloc(entry_size, 1)` asserts on zero-size entries (`:388`)

`omf_calloc_real` in `allocator_default.h:46` does `assert(nmemb > 0)`. Any zero-byte zip entry — or `zip_entry_uncomp_size` returning the error sentinel cast to a huge unsigned — crashes the assert. Add a guard, or skip empty entries.

### Smaller issues

#### 10. `modmanager_init` always returns `true`

…despite the `bool` return type. `engine.c`'s `exit_7` cleanup branch is unreachable. Either return `false` on real failures, or change the signature to `void` and drop the dead `goto exit_7`.

#### 11. `modmanager_get_sprite` does an extra empty-key lookup

After `str_free(&filename)` at `:586`, if neither common-replacement condition matches, neither branch rebuilds `filename`. Line 612 then calls `hashmap_get_str(…, "")`. Not a crash (`str_c` on a freed str returns the zeroed `ssmall` buffer = `""`), but a guaranteed miss with log churn. Add an `else` early-return or move the second lookup inside the matched branches.

#### 12. `modmanager.c:393` — `path path;` shadows the type name

Compiles, but reads oddly. Rename to `p` or similar.

#### 13. `modmanager.c:694-697` — dead `else if` branch

```c
if(origin_count == 0) { … return false; }
else if(origin_count != 1) { … return false; }
```

The `> 1` case already returned at `:669`. By line 691, `origin_count` is 0 or 1; the `!= 1` branch can only fire when `== 0`, which the prior `if` already handled.

src/audio/audio.c
Vagabond: It's freed before the break below

PR #1318, opened 2025-10-02
Clean up REC actions & extra_data

Nopey:
- sd_action now uses the OMF bits directly (without translation), obsoleting the `raw_action` field on sd_rec_move.
- sd_rec_move uses a new smallbuffer utility class, allowing up to 10 bytes of extra_data to be stored in-line.
- field `action` is now part of `extra_data`, simplifying the implementation of packets that aren't actions (e.g. the openomf assert, lookup id 10).
- Added the remaining lookup_id's to the sd_rec_extra_len function.

### DIFF
--- a/src/controller/rec_controller.c
+++ b/src/controller/rec_controller.c
@@ -92,7 +92,7 @@ int rec_controller_poll(controller *ctrl, ctrl_event **ev) {
         int j = 0;
         while(hashmap_get_int(&data->tick_lookup, (ticks * 10) + j, (void **)(&move), &len) == 0) {
             char *extra_data = sd_rec_get_extra_data(move);
-            if(move->lookup_id == 10 && extra_data[0] == 'A') {
+            if(move->lookup_id == 10 && extra_data[0] == REC_LOOKUP10_ASSERT_BYTE) {
                 rec_assertion ass;
                 if(parse_assertion((uint8_t const *)sd_rec_get_extra_data(move), &ass)) {
                     log_assertion(&ass);
@@ -100,7 +100,7 @@ int rec_controller_poll(controller *ctrl, ctrl_event **ev) {
                         crash("REC file assert failed!");
                     }
                 }
-            } else if(move->lookup_id == 10 && extra_data[0] == 4) {
+            } else if(move->lookup_id == 10 && extra_data[0] == REC_LOOKUP10_SETRANDOM_BYTE) {
                 uint32_t seed;
                 memcpy(&seed, extra_data + 4, sizeof(seed));
                 log_debug("setting random seed to %d from REC file", seed);
--- a/src/game/scenes/arena.c
+++ b/src/game/scenes/arena.c
@@ -2107,7 +2107,7 @@ int arena_create(scene *scene) {
         char *extra_data = sd_rec_set_lookup_id(&mv, 10);
         mv.tick = 1;
         uint32_t seed = random_get_seed(&scene->gs->rand);
-        extra_data[0] = 4;
+        extra_data[0] = REC_LOOKUP10_SETRANDOM_BYTE;
         memcpy(extra_data + 4, &seed, sizeof(seed));
         sd_rec_insert_action_at_tick(scene->gs->rec, &mv);
     }
--- a/src/formats/rec.h
+++ b/src/formats/rec.h
@@ -102,6 +102,16 @@ typedef struct {
     sd_rec_move *moves;      ///< REC event records list
 } sd_rec_file;
 
+// Moves with lookup ID 10 are differentiated further by their first byte
+enum
+{
+    REC_LOOKUP10_UNK1_BYTE = 1,
+    REC_LOOKUP10_UNK2_BYTE = 2,
+    REC_LOOKUP10_UNK3_BYTE = 3,
+    REC_LOOKUP10_SETRANDOM_BYTE = 4,
+    REC_LOOKUP10_ASSERT_BYTE = 'A',
+};
+
 /*! \brief Initialize REC file structure
  *
  * Initializes the REC file structure with empty values.
--- a/src/formats/rec_assertion.c
+++ b/src/formats/rec_assertion.c
@@ -1,6 +1,7 @@
 
 
 #include "formats/rec_assertion.h"
+#include "formats/rec.h"
 #include "utils/log.h"
 #include "utils/str.h"
 
@@ -51,8 +52,7 @@ bool parse_assertion(const uint8_t *data, rec_assertion *out) {
     bit_reader reader = {data, 0};
 
     // packet 10 is differentiated by its first byte.
-    // openomf defines 'A' as an Assert.
-    if(read_bits(&reader, 8) != 'A') {
+    if(read_bits(&reader, 8) != REC_LOOKUP10_ASSERT_BYTE) {
         return false;
     }
 
@@ -126,8 +126,8 @@ bool encode_assertion(const rec_assertion *assertion, uint8_t *buffer) {
     memset(buffer, 0, 8);
     bit_writer writer = {buffer, 0};
 
-    // write packet sub-type 'A' for an Assert.
-    write_bits(&writer, 'A', 8);
+    // write packet sub-type for an Assert.
+    write_bits(&writer, REC_LOOKUP10_ASSERT_BYTE, 8);
 
     // Write header (3b op + 1b types)
     write_bits(&writer, assertion->op, 3);
--- a/tools/rectool/main.c
+++ b/tools/rectool/main.c
@@ -140,7 +140,7 @@ void print_rec_root_info(sd_rec_file *rec) {
             printf("%6u %10u %5u %6u %6u %22s", i, rec->moves[i].tick, rec->moves[i].lookup_id, rec->moves[i].player_id,
                    extra_len, tmp);
 
-            if(rec->moves[i].lookup_id == 10 && extra_data[0] == 'A') {
+            if(rec->moves[i].lookup_id == 10 && extra_data[0] == REC_LOOKUP10_ASSERT_BYTE) {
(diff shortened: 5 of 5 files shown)


katajakasa (approved):
Codewise this seems fine. I don't know much about the RECs, so someone else should also review.

src/controller/rec_controller.c
```diff
@@ -74,30 +86,26 @@ int rec_controller_poll(controller *ctrl, ctrl_event **ev) {
...
+                if(parse_assertion((uint8_t const *)sd_rec_get_extra_data(move), &ass)) {
                     log_assertion(&ass);
                     if(!game_state_check_assertion_is_met(&ass, ctrl->gs)) {
                         crash("REC file assert failed!");
                     }
                 }
             } else if(move->lookup_id == 96) {
                 uint32_t seed;
-                memcpy(&seed, move->extra_data, sizeof(seed));
+                memcpy(&seed, sd_rec_get_extra_data(move) + 1, sizeof(seed));
```
Vagabond: Comment here to indicate we ignore the first byte to avoid confusing DOS?

src/formats/actions.h
```diff
@@ -13,16 +13,19 @@
...
-    SD_ACT_RIGHT = 0x20, ///< Move right (walk)
-} sd_action;
+    SD_ACT_NONE = 0x00, ///< No action
+
+    SD_ACT_PUNCH = 0x01, ///< Punch/select
+    SD_ACT_KICK = 0x02,  ///< Kick/select
 
-#define SD_MOVE_MASK (SD_ACT_UP | SD_ACT_DOWN | SD_ACT_LEFT | SD_ACT_RIGHT) ///< Mask of all movement keys
-#define SD_ACT_VALID_BITS 0x3Fu                                             ///< Mask of all valid bits of an sd_action
+    SD_ACT_UPUP = 0x10,
```
Vagabond, line 21: I'm a bit confused here, why did these change, don't they reflect the values in the REC?

src/utils/smallbuffer.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <stddef.h>
+#include <string.h>
+
+typedef struct smallbuffer {
+    // private fields, do not access.
+    ptrdiff_t priv_cap;
+    union {
+        char *priv_alloc;
+        char priv_inline[sizeof(char *)];
```
Vagabond, line 20: Why is this `sizeof(char *)`, basically the size of a pointer, and not something like `8`?

testing/test_rec.c
```diff
@@ -22,9 +22,9 @@ void test_sd_rec_create(void) {
     for(int i = 0; i < 10; i++) {
         sd_rec_move mv;
         fill((char *)&mv, sizeof(sd_rec_move));
-        mv.lookup_id = 2;
-        mv.action = SD_ACT_KICK | SD_ACT_PUNCH | SD_ACT_LEFT | SD_ACT_DOWN;
-        mv.extra_data = NULL;
+        smallbuffer_create(&mv.extra_data);
```
Vagabond, line 25: isn't this smallbuffer_create unnecessary?

src/utils/smallbuffer.h
Vagabond: Ah, I see, it's so the union is always the same size.

src/game/scenes/arena.c
```diff
@@ -929,42 +929,56 @@ void write_rec_move(scene *scene, game_player *player, int action) {
...
-    }
-
-    if(action & ACT_DOWN) {
-        move.action |= SD_ACT_DOWN;
-    }
-
-    if(action & ACT_LEFT) {
-        move.action |= SD_ACT_LEFT;
+    switch(action & ACT_Mask_Dirs) {
+        case ACT_UP:
```
Vagabond, line 949: Not a blocker, but should we plan to unify the ACT_ and SD_ACT enums, so we don't have to convert between them?

testing/test_smallbuffer.c
```diff
@@ -0,0 +1,63 @@
...
+    "ipsum. Sed eveniet repudiandae voluptates inventore quidem at similique.\n";
+static char const oops_all_nul_bytes[400] = "";
+
+void test_smallbuffer(void) {
+    smallbuffer sb;
+    smallbuffer_create(&sb);
+    CU_ASSERT_EQUAL(smallbuffer_capacity(&sb), 0);
+
+    smallbuffer_resize(&sb, 4);
+    CU_ASSERT_EQUAL(smallbuffer_capacity(&sb), 4);
```
Vagabond, line 19: Shouldn't capacity be a minimum of `sizeof(char *)`?

src/formats/actions.h
Nopey: The old shadowdive directions do not match the REC bytes, in fact.

That is why there was both raw_action and action-- `action` was the sd_action, `raw_action` was the untranslated OMF action byte.

OMF directions go from 0x10 to 0x80 in increments of 0x10.
Unlike OpenOMF & old-ShadowDive directions, OMF directions cannot be OR'd together to create inbetweens.
`UPUP | RIGHTRIGHT` does not equal `UPRIGHT` in OMF directions-- that's why I renamed the directions from SD_ACT_UP to SD_ACT_UPUP.

in exchange for not being as bit-fiddle-able, the OMF style directions do make it easier to use the directions as an index into an array, as the 9 valid directions are all together.

testing/test_rec.c
Nopey: Yes, because we eventually pass the `rec` into `sd_rec_free`, which calls smallbuffer_free.

`fill` is not `memset(0)`-- it fills the struct with ascending bytes 1, 2, 3..

src/game/scenes/arena.c
Nopey: `SD_ACT_UPUP | SD_ACT_RIGHTRIGHT` does not equal `SD_ACT_UPRIGHT`.

I don't know how important being able to bit twiddle the openomf ACT_'s are to our game code offhand.

Anyhoo, unifying ACT_ & SD_ACT_ seems out of scope.

testing/test_smallbuffer.c
Nopey: _resize sets the capacity. smallbuffers track their capacity, and the capacity can be less than the inline capacity.

if we want we could have a `smallbuffer_resize_to_at_least` which grows to *at least* the specified size.

src/controller/rec_controller.c
Nopey: lookup_id 96 is being removed in PR 1319, so I'd rather not.

(That PR targets this PR. will merge once it's reviewed)

testing/test_smallbuffer.c
Vagabond: but a smallbuffer always has at least the sizeof(char *) capacity, no? it may contain a smaller object, but it has *room* for at least `sizeof(char*)`? Maybe the function should be called smallbuffer_size, and there should be a separate capacity?

src/controller/rec_controller.c
Vagabond: ok

src/formats/actions.h
Vagabond: I see, so this actually surfaces the *actual* values in the REC, which is what the SD stuff generally is supposed to do. ok.

testing/test_smallbuffer.c
Nopey: renaming the getter smallbuffer_size is reasonable, but I don't want two getters-- then we'd have to check each caller is using the correct one.

Hence the two methods: smallbuffer_resize and smallbuffer_resize_to_at_least-- we want the allocation policy of smallbuffer to be transparent to the caller.

I also thought about having some way for the caller to indicate whether they want existing bytes to be preserved.

Maybe one `smallbuffer_resize` with some flags in an option?

Vagabond: I think the current API is fine for now, because the REC usecase is pretty constrained, but if we start to use smallbuffer more generally I think this limited API may lead to trouble. If I want to copy or write out the bytes from a smallbuffer, I want the size, if I want to write into it, I want to know the capacity, no? Don't a lot of container types in C++ or Rust have both size and capacity getters?

Vagabond: Maybe that's where the `at least` resize comes in, we'd just be careful to always resize to at least the size we want to write?

Nopey: smallbuffer only tracks capacity, it does not track used-size (as std::vec::Vec or std::vector do)
This means that every resize is prone to reallocating, and so should not be called on a whim.

smallbuffer is intended to be useful as a building block for someone implementing a container like vector, where it would be tracking the allocated capacity (and vector would be tracking the used-size).

Nopey: as an alternative to renaming `smallbuffer_capacity` to `smallbuffer_size`,
we could rename `smallbuffer_resize` to something like `smallbuffer_realloc`

Vagabond: I guess I don't understand the resistance to tracking size AND capacity. Especially if this is running on a 32 bit machine, we can't make assumptions about what will/won't fit in the smallbuffer, so we *need* to know the capacity if we want to write into it. doing an `at least` realloc would minimize allocs if the existing smalbuffer can hold the proposed realloc size in the inline or allocated buffer, no?

Nopey: Tracking size AND capacity requires two fields.
I do not want smallbuffer to be larger than it must be.

Some users, such as REC, do not need size AND capacity. User should track used-size themselves.

Nopey: I think i've clarified the API's intent by renaming resize to realloc

and, we'll figure out what the API for alloc-at-least will be when we need it.

Vagabond (approved):
let's land this

PR #1319, opened 2025-10-02
Fix REC DOS backwards Compatibility

Nopey:
Our assertions (lookup_id 10) and setrandom (lookup_id 96) were implemented in such a way as to confuse DOS.

The assertion packets were mostly harmless, but could trigger random codepaths we don't understand in the DOS game.
The setrandom 96 packets would cause a parser desync, garbaging subsequent moves.


Vagabond:
My only concern here is the magic numbers of 'A' and 4, could we define these in a common header somewhere? like in rec.h?

PR #1366, opened 2026-02-12
Try to fix #1363

katajakasa:
Attempt to fix #1363. Honestly, it might be better to just add some #ifdefs and generate multiple shaders and use the one we want.


katajakasa:
This is already in master now.

PR #1367, opened 2026-03-02
Fix two player joystick gameplay

xlarger:
Better share than sit on it - even though I've cobbled this together with AI: Joysticks enumerate - both of them now, with id preferred instead of names. Fallback to keyboard.

### DIFF
--- a/src/controller/joystick.c
+++ b/src/controller/joystick.c
@@ -11,10 +11,15 @@
 
 void joystick_free(controller *ctrl) {
     joystick *k = ctrl->data;
+    if(!k) {
+        return;
+    }
     if(k->haptic) {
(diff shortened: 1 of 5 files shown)


katajakasa:
Hello. SDL_JoystickID is not stable across sessions; they are only valid while joystick is connected and then may completely change on reconnect. See https://wiki.libsdl.org/SDL2/SDL_JoystickID

katajakasa:
Closing as this is a wrong approach; I would recommend looking into GUIDs for stable identifiers -- see https://wiki.libsdl.org/SDL2/SDL_JoystickGetGUID

xlarger:
Fair enough, thanks for checking!

PR #1370, opened 2026-03-28
Double iterator issue

katajakasa:
There may be cases where we have two iterators at the same time pointing to the same vector, so try to resolve that by marking animations done instead of removing them.

This is a bit of a shot in the dark; I could only repeat this once and not sure if this fixes things.


Vagabond (approved):
Seems harmless.

PR #1371, opened 2026-04-07
video: Add borderless fullscreen window mode

maininformer:
## Summary
- Add a new `borderless` config option for borderless fullscreen (`SDL_WINDOW_FULLSCREEN_DESKTOP`)
- `fullscreen` remains a bool — fully backward compatible with existing configs
- In-game video menu gets a separate BORDERLESS ON/OFF toggle
- Borderless fullscreen allows easy Alt-Tab/Cmd-Tab without hijacking the display

## Config example
```
fullscreen=true
borderless=true
```

## Test plan
- [x] `fullscreen=false` — game runs windowed
- [x] `fullscreen=true, borderless=false` — exclusive fullscreen (existing behavior)
- [x] `fullscreen=true, borderless=true` — borderless fullscreen, can Alt-Tab away
- [x] Old configs with just `fullscreen=true` still work (borderless defaults to false)
- [x] Toggle both options via in-game VIDEO settings menu

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/game/scenes/mainmenu/menu_video.c
+++ b/src/game/scenes/mainmenu/menu_video.c
@@ -169,8 +169,7 @@ void menu_video_done(component *c, void *u) {
                    v->fb_scale, v->scaling_mode);
         menu_set_submenu(c->parent, menu_video_confirm_create(s, &local->old_video_settings));
     } else if(local->old_video_settings.screen_w != v->screen_w || local->old_video_settings.screen_h != v->screen_h ||
-              local->old_video_settings.window_mode != v->window_mode ||
-              local->old_video_settings.vsync != v->vsync ||
+              local->old_video_settings.window_mode != v->window_mode || local->old_video_settings.vsync != v->vsync ||
               local->old_video_settings.aspect != v->aspect ||
               local->old_video_settings.framerate_limit != v->framerate_limit ||
               local->old_video_settings.fb_scale != v->fb_scale ||
@@ -305,10 +304,9 @@ component *menu_video_create(scene *s) {
     menu_attach(menu, textselector_create_bind_opts("ASPECT", "Video aspect ratio. Original game is 4:3.", NULL, NULL,
                                                     &setting->video.aspect, aspect_opts, 2));
     const char *window_mode_opts[] = {"OFF", "WINDOWED", "EXCLUSIVE"};
-    menu_attach(menu,
-                textselector_create_bind_opts("FS",
-                                              "OFF=windowed, WINDOWED=borderless fullscreen, EXCLUSIVE=true fullscreen.",
-                                              NULL, NULL, &setting->video.window_mode, window_mode_opts, 3));
+    menu_attach(menu, textselector_create_bind_opts(
+                          "FS", "OFF=windowed, WINDOWED=borderless fullscreen, EXCLUSIVE=true fullscreen.", NULL, NULL,
+                          &setting->video.window_mode, window_mode_opts, 3));
 
     // Done button
     menu_attach(menu, button_create("DONE", "Return to the main menu.", false, false, menu_video_done, s));
--- a/src/game/scenes/mainmenu/menu_video.c
+++ b/src/game/scenes/mainmenu/menu_video.c
@@ -165,16 +165,17 @@ void menu_video_done(component *c, void *u) {
     bool render_plugin_changed = strcmp(v->renderer, local->old_video_settings.renderer) != 0;
     if(render_plugin_changed) {
         video_close();
-        video_init(v->renderer, v->screen_w, v->screen_h, v->fullscreen, v->vsync, v->aspect, v->framerate_limit,
+        video_init(v->renderer, v->screen_w, v->screen_h, v->window_mode, v->vsync, v->aspect, v->framerate_limit,
                    v->fb_scale, v->scaling_mode);
         menu_set_submenu(c->parent, menu_video_confirm_create(s, &local->old_video_settings));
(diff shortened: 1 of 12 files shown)


maininformer:
Closing for now — still WIP. Will reopen when backward compatibility with `fullscreen=true` configs and menu alignment are addressed.

src/video/renderers/opengl3/sdl_window.c
```diff
@@ -96,10 +96,11 @@ bool create_window(SDL_Window **window, int width, int height, bool fullscreen)
     }
 
     if(fullscreen) {
-        if(SDL_SetWindowFullscreen(w, SDL_WINDOW_FULLSCREEN) != 0) {
+        Uint32 fs_flag = borderless ? SDL_WINDOW_FULLSCREEN_DESKTOP : SDL_WINDOW_FULLSCREEN;
```
katajakasa: We don't use the SDL int types.

src/game/scenes/mainmenu/menu_video.c
```diff
@@ -305,6 +306,9 @@ component *menu_video_create(scene *s) {
                                                     &setting->video.aspect, aspect_opts, 2));
     menu_attach(menu, textselector_create_bind_opts("FULLSCREEN", "Run the game in a fullscreen window.", NULL, NULL,
                                                     &setting->video.fullscreen, offon_opts, 2));
+    menu_attach(menu,
+                textselector_create_bind_opts("BORDERLESS", "Use borderless fullscreen for easy Alt-Tab switching.",
```
katajakasa: due to how little room we have in the UI, the fullscreen option should be multi-option toggleable. Also, the code could use enum for the variable type. To avoid compatibility issues, the option in configs should be called something new (e.g. window_mode or somesuch).

EDIT: clarified a bit

katajakasa:
Note that since we now started receiving AI code PR's, we are currently considering the AI policy. This PR is probably going to have to wait until we have some idea how we want to handle them.

maininformer:
Thank you.

> Note that since we now started receiving AI code PR's, we are currently considering the AI policy. This PR is probably going to have to wait until we have some idea how we want to handle them.

This sounds good to me; please let me know if I can be of help.

I Will address the comments.

One question, should the borderless be the default behavior of fullscreen? Adding a new enum works. The issue is many players may not know what "borderless" means. Plus the word "borderless" is too long for the UI and mangles it. Can try shorter words but that would be even more mysterious.

katajakasa:
> Thank you.
> 
> > Note that since we now started receiving AI code PR's, we are currently considering the AI policy. This PR is probably going to have to wait until we have some idea how we want to handle them.
> 
> This sounds good to me; please let me know if I can be of help.
> 
> I Will address the comments.
> 
> One question, should the borderless be the default behavior of fullscreen? Adding a new enum works. The issue is many players may not know what "borderless" means. Plus the word "borderless" is too long for the UI and mangles it. Can try shorter words but that would be even more mysterious.

I would just use:
FULLSCREEN: OFF
FULLSCREEN: WINDOWED
FULLSCREEN: EXCLUSIVE

and then explain the modes in description. If there is need to shorten, then "FULLSCREEN" can just be "FS" or "FS MODE" or somesuch (with description explanation).

src/game/utils/settings.h
```diff
@@ -1,6 +1,13 @@
 #ifndef SETTINGS_H
 #define SETTINGS_H
 
+typedef enum
+{
+    WINDOW_MODE_WINDOWED,
+    WINDOW_MODE_BORDERLESS,
+    WINDOW_MODE_FULLSCREEN
+} window_mode;
```
katajakasa, line 9: there should also be a typedef for this, and that should be used as a function argument type, if possible.

katajakasa:
Right, the AI policy is now at https://github.com/omf2097/openomf/blob/master/CONTRIBUTING.md#ai--coding-assistant-policy -- please check it out.

maininformer:
Addressed.

src/video/renderers/opengl3/sdl_window.c
```diff
@@ -73,7 +74,7 @@ bool has_gl_available(int version_major, int version_minor) {
     return ret;
 }
 
-bool create_window(SDL_Window **window, int width, int height, bool fullscreen) {
+bool create_window(SDL_Window **window, int width, int height, int window_mode) {
```
katajakasa, line 77: You still need to use the new type. For example this would be "window_mode window_mode" instead of "int window_mode". Idea is that we declare by type that this variable can only contain the three types defined by the typedef. This goes for all functions where this is used.

maininformer:
Updated to use the `window_mode` typedef instead of `int` in all function signatures, function pointer typedefs, and local variables. The `settings_video` struct field remains `int` because the GUI textselector widget binds to it as `int *`.

Also updated the commit trailers to use `Assisted-by` per the new AI policy.

katajakasa:
Issue: Switching to fullscreen exclusive seems to break 4:3 aspect ratio with black sides.

maininformer:
@katajakasa cannot reproduce this. Could you please share an image? and is this behavior absent on master?

katajakasa:
> @katajakasa cannot reproduce this. Could you please share an image? and is this behavior absent on master?

Sorry, took me a while to get back to this. So after some poking, it seems that the fullscreen aspect ratio stuff is broken in master also. What /should/ happen is that when 4:3 is selected, and fullscreen is enabled (any fullscreen mode), and a 16:9 or 16:10 resolution is selected, there should be black bars in the right and the left of the screen. I think we can consider that a separate issue though.

Edit: Otherwise I think this seems fine now. One more squash so that things are in a single commit, and I think we can merge.

PR #1373, opened 2026-04-11
Extended palette

katajakasa:
Extends palette from 256 colors to 1024.


src/video/vga_palette.c
```diff
@@ -9,49 +9,45 @@ void vga_palette_init(vga_palette *palette) {
 }
 
 void vga_palette_tint_range(vga_palette *pal, vga_index ref_index, vga_index start, vga_index end, uint8_t step) {
-    // NOTE: These functions are required to treat a range of 1,0 as starting at 1 and ending at 255.
```
Vagabond, line 12: Does this no longer apply?

katajakasa: I haven't tested yet :D I didn't quite remember why they were there and just dropped them for now; need to still test them out. There's a bunch of small details like these in there still (like how the games internal screenshots do not yet work, we need to extend the screen capture to use 16bit ranges when taking the newsroom shots, etc.)

katajakasa: Right, I think the old fix was because our palette index variable was 8bit and it was rolling over. It's now signed 16bit int, so this should be unnecessary.

shaders/debug_atlas.frag
```diff
@@ -0,0 +1,15 @@
+#version 330 core
+
+in vec2 tex_coord;
+layout (std140) uniform palette {
+    vec4 colors[1024];
```
Nopey: Does the palette need 16 bytes per color swatch? 16KiB for 1024 colors seems like a waste of upload bandwidth.

```suggestion
    uint32 colors[1024];
```

Dropping the alpha channel would result in an additional 25% saving, but is more cumbersome and requires multiple stabs to sample a single color.

src/video/vga_remap.h
```diff
@@ -7,16 +7,17 @@
...
-static_assert(256 == sizeof(vga_remap_table), "vga_palette should pack into 256 bytes");
+static_assert((2 * VGA_PALETTE_SIZE) == sizeof(vga_remap_table), "vga_remap_table should pack properly");
 
 typedef struct vga_remap_tables {
     vga_remap_table tables[VGA_REMAP_COUNT];
 } vga_remap_tables;
 
-static_assert(4864 == sizeof(vga_remap_tables), "vga_remap_tables should pack into 4864 bytes");
+static_assert((2 * VGA_REMAP_COUNT * VGA_PALETTE_SIZE) == sizeof(vga_remap_tables),
+              "vga_remap_tables should pack properly");
```
Nopey:
```suggestion
static_assert((sizeof(vga_index) * VGA_REMAP_COUNT * VGA_PALETTE_SIZE) == sizeof(vga_remap_tables),
              "vga_remap_tables should pack properly");
```

Nopey (commented):
I would suggest gating this entire change behind a compile-time flag.

Lower-end targets such as Nintendo 64 and DOS will not want a 16-bit vga_index.

shaders/debug_atlas.frag
katajakasa: What I'd prefer to do here is to move the whole palette from UBO to an 1D texture, which would then let us drop the whole UBO code, and manage the palette size by variable (256 or 1024 sized).

katajakasa:
> I would suggest gating this entire change behind a compile-time flag.
> 
> Lower-end targets such as Nintendo 64 and DOS will not want a 16-bit vga_index.

Yep. Ideally, such platforms would just change the VGA_PALETTE_SIZE from VGA_EXTENDED_PALETTE_SIZE to VGA_STANDARD_PALETTE_SIZE and that's that. I still need to fix the remaps to work with that though (plus the gpu uploads).

src/video/vga_remap.h
katajakasa: Fixed!

shaders/debug_atlas.frag
katajakasa: There we go, that should be a bit cleaner I think.

katajakasa:
WIP: It might be a good idea to make a separate cmake flag for testing this (as @Nopey suggested)

katajakasa:
> WIP: It might be a good idea to make a separate cmake flag for testing this (as @Nopey suggested)

And done.

PR #1374, opened 2026-04-11
Add AI policy

katajakasa:
Initial AI policy. I think we will need to clarify and modify in the future, but at least it now exists :P


Vagabond:
I think this is about the best we can hope for here, as I suspect explicitly banning it will just lead to people not disclosing their use of a LLM.

PR #1376, opened 2026-04-15
16 bit internal surface

katajakasa:
This makes the engine itself 16bit image capable; the 8bit to 16bit interface moves to the loader.

note -- I pulled this from git stash (and fixed a couple of bugs), not properly tested yet. Also contains some const correctness stuff which I didn't bother separating to another commit.


katajakasa:
Tested a bit more, seems to work. Let's merge so that modding can keep going.

PR #1378, opened 2026-04-15
video: Handle surface 8bit/16bit mode better

katajakasa:
Resize the surface data depending on whether we are in extended palette mode or not. Should save memory on lower-end platforms.

### DIFF
--- a/src/video/vga_palette.h
+++ b/src/video/vga_palette.h
@@ -12,11 +12,15 @@
 // USE_EXTENDED_PALETTE is set by cmake
 #ifdef USE_EXTENDED_PALETTE
 #define VGA_PALETTE_SIZE VGA_EXTENDED_PALETTE_SIZE
+typedef uint16_t vga_pixel;
 #else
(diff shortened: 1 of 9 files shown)


src/video/vga_palette.h
```diff
@@ -12,11 +12,15 @@
...
+typedef uint16_t vga_pixel;
 #else
 #define VGA_PALETTE_SIZE VGA_STANDARD_PALETTE_SIZE
+typedef uint8_t vga_pixel;
 #endif
 
-typedef int16_t vga_index;
+// This is what we use for iterating and assigning values. It covers the uint16_t and uint8_t vga_pixel regions.
+// Assert should be used in handlers to ensure we stay withing the [0 ... n ... VGA_PALETTE_SIZE] region.
+typedef int32_t vga_index;
```
Vagabond, line 23: wouldn't int16 suffice here?

katajakasa: for our use cases yes. However, if the pixel is uint16_t, its range can be 0-65535, and int16_t would cover only 0-32767. Technically does not matter, but might as well use an int for correctness.

PR #1391, opened 2026-05-10
Improve Windows build process and documentation

miguelmartins1987:
* **What my change is:** An attempt to facilitate building on Windows, using either MSVC (VS2022/2026) or Clang for Windows
* **What it does:** Essentially, it gets rid of the following two build errors when using Clang on Windows:

1. Signed vs unsigned comparisons in `keyboard_binds_key` (keyboard.c)
2. Usage of the `PRIu32` macro in `arena_state_dump` (arena.c) by relying on a transitive include (`#include <inttypes.h>` was added to the top of the file for this purpose)

It also documents a new building procedure on Windows, allowing Windows users/developers to choose between MSVC or Clang (see: BUILD.md)

* **Why it should be merged:** I believe this should be merged to not only provide users with two ways to compile on Windows, but also to correct the build errors stated above.

This is my very first PR; constructive feedback is, of course, welcome and appreciated.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -2,7 +2,7 @@
 
 ### Dependencies
 
-Use at least GCC 9 or Clang 10. MSVC is not supported.
+Use at least GCC 9 or Clang 10. MSVC is also supported, namely the version that ships with VS2022's (or higher) Build Tools.
 
 Required:
 * SDL2 (>=2.0.16): http://www.libsdl.org/download-2.0.php
@@ -14,6 +14,14 @@ Required:
 * libxmp: https://github.com/cmatsuoka/libxmp
 * opusfile: https://github.com/xiph/opusfile
 
+
+On Windows, a `vcpkg` installation is required. In turn, `vcpkg` can be used in Manifest mode to pull the required dependencies.
+Replace `C:\vcpkg` with a directory of your choice, if needed.
+
+```
+git clone https://github.com/microsoft/vcpkg.git C:\vcpkg
+C:\vcpkg\bootstrap-vcpkg.bat
+```
 
 On Ubuntu, it is possible to pull the libraries using apt-get.
 ```
@@ -39,7 +47,14 @@ project in GitHub.
 
 ### Compiling
 
-To compile (change install prefix if necessary):
+To compile on Windows, use the "x64 Native Tools Command Prompt for VS", located on your Start Menu:
+
+```
+cmake -B build -S . --toolchain C:\vcpkg\scripts\buildsystems\vcpkg.cmake
+cmake --build build
+```
+
+To compile on other operating systems (change install prefix if necessary):
 
 ```
 $ mkdir -p build
--- a/BUILD.md
+++ b/BUILD.md
@@ -47,9 +47,10 @@ project in GitHub.
 
 ### Compiling
 
-To compile on Windows, use the "x64 Native Tools Command Prompt for VS", located on your Start Menu:
+To compile on Windows, use the "x64 Native Tools Command Prompt for VS":
 
 ```
+cd C:\path\to\openomf
 cmake -B build -S . --toolchain C:\vcpkg\scripts\buildsystems\vcpkg.cmake
 cmake --build build
 ```
--- a/src/game/scenes/arena.c
+++ b/src/game/scenes/arena.c
@@ -1,4 +1,5 @@
 #include <SDL.h>
+#include <inttypes.h>
 #include <math.h>
 #include <stdbool.h>
 #include <stdio.h>
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,6 +1,5 @@
 {
   "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
-  "builtin-baseline": "082b32b6c5de20fd01f61dfaa9856fcba849e73e",
   "dependencies": [
     {
       "name": "sdl2",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
+  "builtin-baseline": "082b32b6c5de20fd01f61dfaa9856fcba849e73e",
   "dependencies": [
     {
       "name": "sdl2",


BUILD.md
```diff
@@ -2,7 +2,7 @@
 
 ### Dependencies
 
-Use at least GCC 9 or Clang 10. MSVC is not supported.
+Use at least GCC 9 or Clang 10. MSVC is also supported; it was tested with the version that ships with VS2022's Build Tools. Using VS2026's Build Tools is equally supported.
```
Nopey: it *is* tested with VS 2022 Build Tools' MSVC on every commit in CI (not was).

Could you write fewer words here?
Please do not mention that MSVC was tested in this sentence-- notice how GCC and Clang are listed as supported, but there is no mention of what distro or installation method was used for testing them.

CMakePresetsTemplate.json
```diff
@@ -0,0 +1,78 @@
+{
+    "version": 3,
+    "configurePresets": [
+        {
+            "name": "base-win",
+            "hidden": true,
+            "generator": "Ninja",
```
Nopey: Is there a reason you're recommending all windows developers use Ninja?
I use the "Visual Studio 17 2022" generator, which is the default in 2022 build tools.
It's also a multi-config generator, so I can switch between Debug and Release builds without reconfiguring CMake.

There also exists "Ninja Multi-Config".

CMakePresetsTemplate.json
```diff
@@ -0,0 +1,78 @@
...
+            "binaryDir": "${sourceDir}/build",
+            "cacheVariables": {
+                "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
+                "CMAKE_TOOLCHAIN_FILE": "{{REPLACE_ME}}",
```
Nopey: I think `"{{REPLACE_ME}}"` is unfortunate, and is why I have not added a CMakePresets.json.

win-build-setup.ps1
```diff
@@ -0,0 +1,3 @@
+Copy-Item .\CMakePresetsTemplate.json .\CMakePresets.json
+$VCPKG_CMAKE_PATH = ($env:VSINSTALLDIR -replace '\\', '/') + "VC/vcpkg/scripts/buildsystems/vcpkg.cmake"
+(Get-Content -Path ".\CMakePresets.json") -replace '\{\{REPLACE_ME\}\}', $VCPKG_CMAKE_PATH | Set-Content -Path ".\CMakePresets.json"
```
Nopey: I would suggest shelling out to cmake for the configure directly, rather than writing a CMakePresets.json.

```
cmake -B build -S . --toolchain $VCPKG_CMAKE_PATH
```
is really all you need.

Are you using the CMake integration in VSCode?
It's the first tool that comes to mind that leans so heavily on CMakePresets.json.

Nopey:
I didn't read the PR description before reading the commits, so one more thing!

>    What it does: Essentially, it gets rid of the following two build errors when using Clang on Windows:
>
>    Signed vs unsigned comparisons in keyboard_binds_key (keyboard.c)
>    Usage of the PRIu32 macro in arena_state_dump (arena.c) by relying on a transitive include (#include <inttypes.h> was added to the top of the file for this purpose)

This kind of explanation of "why" is always great, and I love seeing it.
I generally [prefer it be in a commit message body](https://martijnhols.nl/blog/how-to-write-a-good-git-commit-message), rather than the PR description, so that future generations who are looking at `git log` or `git blame` can see why a commit makes its changes.

BUILD.md
miguelmartins1987: I can write fewer words here, indeed. How's this?

`Use at least GCC 9 or Clang 10. MSVC is also supported, namely the version that ships with VS2022's (or higher) Build Tools.`

CMakePresetsTemplate.json
miguelmartins1987: On second thought, yes, it is probably better to let the Windows developers choose their preferred generator. But to answer your question: no, there is no particular reason.

I might drop this `CMakePresetsTemplate.json` file in favor of `cmake -B build -S . --toolchain $VCPKG_CMAKE_PATH`, as you suggested in another comment (see below).

CMakePresetsTemplate.json
miguelmartins1987: Fair enough. Perhaps `"{{VAR_TOOLCHAIN_FILE}}"` would better convey the intent here? But that would be under the assumption that I keep this file (which I might not, see below).

win-build-setup.ps1
miguelmartins1987: I am indeed using the CMake integration in VSCode. My intent here was to try and make the build process simple, but I suppose I might have over-engineered a bit. I propose telling the developer to open a powershell session inside the "x64 Native Tools Command Prompt for VS" command prompt (by simply typing `powershell`), then executing the following commands:

```
$VCPKG_CMAKE_PATH = ($env:VSINSTALLDIR -replace '\\', '/') + "VC/vcpkg/scripts/buildsystems/vcpkg.cmake"
cmake -B build -S . --toolchain "$VCPKG_CMAKE_PATH"
cmake --build build
```

Then, no `win-build-setup.ps1` or `CMakePresetsTemplate.json` files would be needed. What do you think?

Nopey: Those commands should work.

I think `$VCPKG_CMAKE_PATH = ($env:VSINSTALLDIR -replace '\\', '/') + "VC/vcpkg/scripts/buildsystems/vcpkg.cmake"` is too arcane for BUILD.md, please simplify these instructions

Conversion to forward slashes doesn't seem necessary anymore, now that we're no longer injecting it into JSON.

VCPKG is an optional component of Visual Studio, and so users may encounter issues where it is not present in VSINSTALLDIR :
image

Please provide some guidance in BUILD.md on vcpkg installation-- I prefer to check the microsoft/vcpkg git repo out myself (as the same install process can be done on Linux and MacOS) rather than using VS Installer, but either method is acceptable.

Please document the VCPKG installation process in the Dependencies section, near the `apt-get` and `brew install` code blocks.

BUILD.md
Nopey: That sounds good, feel free to use what you've suggested above. It could also be as short as: `Use at least GCC 9, Clang 10, or Visual Studio 2022's MSVC.`, but it doesn't need to be so terse.

BUILD.md
```diff
@@ -39,7 +47,14 @@ project in GitHub.
 
 ### Compiling
 
-To compile (change install prefix if necessary):
+To compile on Windows, use the "x64 Native Tools Command Prompt for VS", located on your Start Menu:
+
+```
+cmake -B build -S . --toolchain C:\vcpkg\scripts\buildsystems\vcpkg.cmake
```
Nopey: missing step: `cd C:\path\to\openomf` (since the native tools command prompt starts in the VS install directory)

or, have more faith in the user & don't mention that the command prompt is in the Start Menu.

Nopey:
Squashed the first two commits into a5b8ad19706c951ab1e35cfeb28b601d9f245ccb, and provided the relevant part of your PR's description as the commit message.

Pushed a commit reenabling msvc warnings C4389 and C4459; you just fixed the C4389, C4459 was fixed at some previous time.

Nopey:
The CI pipeline failed, as it checked out the most recent tagged version of VCPKG, [2026.04.27](https://github.com/microsoft/vcpkg/tree/2026.04.27) (aka 56bb2411609227288b70117ead2c47585ba07713), which is older than the baseline you've specified.

BUILD.md
miguelmartins1987: I opted for both including `cd C:\path\to\openomf` and removing the "located on your Start Menu" excerpt. I believe that should make it easier and cleaner for any users who wish to compile.

PR #1394, opened 2026-05-13
Miscellaneous cleanups

katajakasa:
Clang-tidy and manual cleanups stuff from my todo text file


Nopey (approved):
Good catch on the str_format bug.

Nit: I think it strange, the use of "const correctness" to describe adding const to local variables with no references or pointers involved.

katajakasa:
> Good catch on the str_format bug.
> 
> Nit: I think it strange, the use of "const correctness" to describe adding const to local variables with no references or pointers involved.

Heh, it's const correctness from CLion and clang-tidy point of view :P

PR #1396, opened 2026-05-29
Audio improvements part II

katajakasa:
Add the more correct audio handling from the original. Echo still missing, as are other more minor details (there are exceptions for different sound ranges on what the defaults are).

Note that tests are Assisted-by claude. I couldn't be arsed to write them manually. I did tinker with them, but the main plan there was to just nail down the implementation so I don't cause breakage later, so whatever.

### DIFF
--- a/src/game/game_state.c
+++ b/src/game/game_state.c
@@ -1,12 +1,11 @@
 #include "game/game_state.h"
-#include "audio/audio.h"
-#include "audio/sound_sources/dat_source.h"
 #include "console/console.h"
 #include "controller/joystick.h"
 #include "controller/keyboard.h"
 #include "controller/rec_controller.h"
 #include "formats/error.h"
 #include "formats/pilot.h"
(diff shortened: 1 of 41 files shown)


src/game/game_state.c
```diff
@@ -926,98 +925,14 @@ void game_state_call_tick(game_state *gs, int mode) {
...
-            vector_delete(&gs->sounds, &it);
-        }
+    if(mode == TICK_DYNAMIC) {
+        const int delta = game_state_ms_per_dyntick(gs);
+        sound_tracker_tick(&gs->tracker, delta, sound_pan_lookup_object, gs);
     }
 }
 
 void game_state_merge_sounds(game_state *old, game_state *new) {
-    // We need to do several things here:
```
Vagabond, line 944: This comment got lost moving to sound_tracker?

katajakasa: Yeah, re-added.

PR #1404, opened 2026-06-14
Formats cleanups, part 3

katajakasa:
Use our own types in places where they make sense. Also some other tweaks and cleanups included.


src/controller/net_controller.c
```diff
@@ -327,8 +327,7 @@ void send_game_information(wtf *data) {
     serial_write_int8(&ser, sd_pilot_get_player_color(player->pilot, PRIMARY));
     serial_write_int8(&ser, sd_pilot_get_player_color(player->pilot, SECONDARY));
     serial_write_int8(&ser, sd_pilot_get_player_color(player->pilot, TERTIARY));
-    serial_write_int8(&ser, strlen(player->pilot->name));
```
Vagabond, line 330: you dropped the length prefixing? why?

src/game/utils/serial.c
```diff
@@ -114,6 +114,25 @@ void serial_read(serial *s, char *buf, size_t len) {
     s->rpos += len;
 }
 
+void serial_write_str(serial *s, const str *src) {
+    if(str_size(src) > 127) {
+        crash("Invalid serial write -- string must be shorter than 128 bytes!");
```
Vagabond: why not treat it as unsigned, a negative string length makes no sense

src/controller/net_controller.c
Vagabond: oh, serial_write_str does it, fine.

src/game/utils/serial.c
katajakasa: I could, but that would change the protocol. If that is fine, then sure.

katajakasa: never mind, I misunderstood the original source. Fixed.

PR #1412, opened 2026-07-02
Bundle soundtracks to releases

katajakasa:
Bundle soundtrack mods to releases. This does increase the zip size somewhat.

### DIFF
--- a/.github/actions/fetch_mods/action.yaml
+++ b/.github/actions/fetch_mods/action.yaml
@@ -1,0 +1,24 @@
+name: Fetch mods
+description: Fetch new soundtracks
+inputs:
+  mods_url:
+    description: 'Soundtrack mod ZIP package URL'
+    required: false
+    default: 'https://github.com/omf2097/openomf-music-mod/releases/latest/download/openomf-mods.zip'
+  filename:
+    description: 'Downloaded soundtrack bundle filename'
+    required: false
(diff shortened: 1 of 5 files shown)


.github/actions/fetch_mods/action.yaml
```diff
@@ -0,0 +1,24 @@
...
+  mods_url:
+    description: 'Soundtrack mod ZIP package URL'
+    required: false
+    default: 'https://github.com/omf2097/openomf-music-mod/releases/latest/download/openomf-mods.zip'
+  filename:
+    description: 'Downloaded soundtrack bundle filename'
+    required: false
+    default: 'openomf-mods.zip'
+runs:
+  # NOTE! No caching, since the soundtrack bundle file might change!
```
mrannanj, line 13: Description says there would be caching though? So should make them consistent, whatever the choice.

mrannanj (changes requested):
Would the new AppImage build want this too?

.github/actions/fetch_mods/action.yaml
katajakasa: Fixed. That was a remnant from copypasting from the assets download action.

katajakasa:
> Would the new AppImage build want this too?

yes, rebased and added.

mrannanj (approved):

PR #1423, opened 2026-07-19
Harden controller action list handling

katajakasa:
Hardens how the controllers read and write per-tick action lists. Parsing moves into shared serial helpers and all scan loops get proper bounds, so malformed packets can no longer overflow the fixed-size buffers. Includes a couple of small cleanups.

### DIFF
--- a/src/controller/net_controller.c
+++ b/src/controller/net_controller.c
@@ -1272,14 +1272,14 @@ int net_controller_poll(controller *ctrl, ctrl_event **ev) {
         if(e->events[id][0] != 0 && e->tick == current_tick) {
             // events for the current tick, send em all
             int i = 0;
-            while(e->events[id][i]) {
+            while(i < MAX_EVENTS_PER_TICK && e->events[id][i]) {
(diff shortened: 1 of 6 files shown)


src/controller/net_controller.c
```diff
@@ -4,6 +4,8 @@
 #include <stdio.h>
 #include <time.h>
 
+#include <SDL.h>
```
LegendaryBlueShirt, line 7: Why?

katajakasa: I moved it from the header to here.

LegendaryBlueShirt: But this file doesn't use it...

LegendaryBlueShirt: I see now, net_controller.h didn't need SDL.h in the first place.

katajakasa: Yeah. So I figured I could probably remove it, but since it's one of those headers you tend to need sooner or later for some type or something, I decided to leave it to the C file. We can just prune includes after 1.0 :P

LegendaryBlueShirt (approved):

PR #1426, opened 2026-07-19
Add 3 tick buffer for punch and kick inputs

LegendaryBlueShirt:
Should make combos easier to perform in general, matches behavior of DOS OMF.

PR Drafted until verified working in netplay.


Vagabond (approved):
I think this looks safe

PR #1428, opened 2026-07-20
Package OpenOMF for MacOS 26

mrannanj:


CMakeLists.txt
```diff
@@ -498,3 +516,54 @@ else()
     )
     install(DIRECTORY shaders/ DESTINATION share/games/openomf/shaders COMPONENT Data)
 endif()
+
+if(APPLE)
+    find_package(SDL2 REQUIRED)
+    find_package(SDL3 REQUIRED)
```
Nopey, line 522: Why are we finding and installing SDL3 here?
We use SDL2.

CMakeLists.txt
```diff
@@ -498,3 +516,54 @@ else()
...
+                 DESTINATION \"\${BUNDLE_FRAMEWORKS_DIR}\"
+                 RENAME \"libSDL3.dylib\"
+            )
+        endif()
+
+        set(EMBEDDED_SDL3 \"\${BUNDLE_FRAMEWORKS_DIR}/${SDL3_FILENAME}\")
+        set(SEARCH_DIRS \"/opt/homebrew/lib\" \"/usr/local/lib\")
+
+        set(BU_CHMOD_BUNDLE_ITEMS ON)
+        fixup_bundle(\"\${APP_PATH}\" \"\${EMBEDDED_SDL3}\" \"\${SEARCH_DIRS}\")
```
Nopey: fixup_bundle should be doing the heavy lifting here, no need to spoon feed it SDL3's location and searchpath imo.

```
fixup_bundle(\"\${APP_PATH}\" \"\" \"\")
```

Which should be sufficient to copy over all our runtime deps (sdl2, sdl2_mixer, libnatpmp, etc..)

CMakeLists.txt
```diff
@@ -498,3 +516,54 @@ else()
...
+
+        # 2. Synchronize signatures across all active variations
+        execute_process(COMMAND codesign --force --sign - \"\${BUNDLE_FRAMEWORKS_DIR}/libSDL3.dylib\" OUTPUT_QUIET ERROR_QUIET)
+        execute_process(COMMAND codesign --force --sign - \"\${EMBEDDED_SDL3}\" OUTPUT_QUIET ERROR_QUIET)
```
Nopey: Why are we signing SDL3, but not our other dylibs?
If signing is required for all our shipped dylibs, we could do an ugly for loop over the directory listing & sign them; but I suspect they're already signed, having come from brew.

Image

CMakeLists.txt
```diff
@@ -498,3 +516,54 @@ else()
...
+        get_filename_component(REAL_NAME \"${SDL3_REAL_SOURCE}\" NAME)
+        if(NOT \"\${REAL_NAME}\" STREQUAL \"${SDL3_FILENAME}\")
+            file(RENAME \"\${BUNDLE_FRAMEWORKS_DIR}/\${REAL_NAME}\" \"\${BUNDLE_FRAMEWORKS_DIR}/${SDL3_FILENAME}\")
+        endif()
+        if(NOT \"${SDL3_FILENAME}\" STREQUAL \"libSDL3.dylib\")
+            file(INSTALL \"${SDL3_REAL_SOURCE}\"
+                 DESTINATION \"\${BUNDLE_FRAMEWORKS_DIR}\"
+                 RENAME \"libSDL3.dylib\"
+            )
+        endif()
```
Nopey: redundantly file copying SDL3 here (twice!). we do so below in the call to fixup_bundle

.github/actions/build_macos_arm/action.yaml
```diff
@@ -10,13 +10,15 @@ runs:
...
+        brew install confuse cunit enet libepoxy libnatpmp libxmp miniupnpc \
+          opusfile sdl2-compat sdl2_mixer sdl3
+        brew install coreutils create-dmg
 
     - name: Generate Release
       shell: bash
       run: |
         mkdir build-release && cd build-release
-        cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=release/usr ..
+        cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=. ..
```
Nopey, line 21: before merging, I imagine you'll want to set the build type back to Release

.github/actions/build_macos_arm/action.yaml
```diff
@@ -25,26 +27,63 @@ runs:
...
+        if ! grep -q "$OPENOMF_VERSION" "$RUNNER_TEMP/version"; then
+          echo "$RUNNER_TEMP/version does not include $OPENOMF_VERSION"
+          exit 1
+        fi
+
+    - name: Run REC tests
+      shell: bash
+      env:
+        OPENOMF_RESOURCE_PATH: /Applications/openomf.app/Contents
+      run: ./run_rectests.sh build-release
```
Nopey, line 95: Let's not run the rec tests more times than we have to, as they take 6 minutes in CI.
They're run as part of the "Functional CI / Run rec tests and menu tests" action.

CMakeLists.txt
mrannanj: The Apple build will be using SDL3 through SDL2 compatibility interface.

CMakeLists.txt
mrannanj: Yeah, this manual SDL3 handling is because it did not get copied automatically.

.github/actions/build_macos_arm/action.yaml
mrannanj: Probably, or at least making a different run for the rec tests.

CMakeLists.txt
Nopey: ah, yes it is.
libSDL2-2.0.0.dylib has strings like `sdl2-compat  : SDL3 library is too old`, so that should be good.

If we remove all mention of SDL3 from CMakeLists.txt, is BundleUtilities able to deduce sdl2-compat's dependency on SDL3 from metadata & copy it without any help?
Is there some way we can check if the SDL2 returned by find_package is actually an sdl2-compat?

CMakeLists.txt
Nopey: Did fixup_bundle(APP SDL3.lib SEARCHPATH) copy any files at all? Presently (and as mentioned in another conversation) the scripts creates both a  `libSDL3.0.dylib` and a `libSDL3.dylib`, perhaps the sdl2-compat is looking for a certain filename?

CMakeLists.txt
mrannanj: This is correct, I reduced it to one line.

CMakeLists.txt
```diff
@@ -498,3 +516,36 @@ else()
...
+    set(SDL3_BUNDLE_SCRIPT "set(SDL3_LIBRARY_PATH [=[${SDL3_LIBRARY_PATH}]=])\n")
+
+    string(APPEND SDL3_BUNDLE_SCRIPT [[
+        include(BundleUtilities)
+
+        set(APP_PATH "${CMAKE_INSTALL_PREFIX}/openomf.app")
+        set(CONTENTS_DIR "${APP_PATH}/Contents")
+
+        fixup_bundle("${APP_PATH}" "" "/opt/homebrew/lib;/usr/local/lib")
+        file(INSTALL "${SDL3_LIBRARY_PATH}" DESTINATION "${CONTENTS_DIR}/Frameworks" RENAME "libSDL3.dylib")
```
Nopey, line 538: `file(INSTALL ... RENAME` doesn't exist.
https://cmake.org/cmake/help/latest/command/file.html#install

Perhaps use `file(COPY_FILE` (new in cmake 3.21) or `configure_file(<input> <output> COPYONLY)` instead? file(INSTALL followed by a rename defeats some of INSTALL's features, anyway.

Nopey: I suspect we may run into trouble `dlopen`'ing SDL3 after using the bare filecopy, as `fixup_bundle` will not have tracked down the dylib dependencies of SDL3.

If this trouble does not occur, the manual file copy is good enough to stay-- considering an SDL3 port is inevitable, and will introduce a direct link dependency on SDL3, enabling fixup_bundle to grab SDL3 for us without any of this dancing around.

Dumb idea: What if we linked SDL3 into the openomf target on macos? I wouldn't expect symbol conflicts between sdl2-compat and SDL3, and if the dep survives the linker, we've just successfully told `fixup_bundle` to copy over the SDL3 that our sdl2-compat needs.

mrannanj: Yeah, I'll try out the options for this rename stuff. The intent is anyway to copy and rename in one command, and I think that part did work at some point. The current version is failing at the signing part.

About linking to sdl3, I'm not sure how that would be done, but have to try and find out.

Vagabond:
Superceeded by #1446

PR #1429, opened 2026-07-22
Fix peculiar corner behavior

LegendaryBlueShirt:
Corner crossup behavior is now accurate to source
Minor: Wallslam is no longer disabled after hp reaches 0


src/game/objects/har.c
```diff
@@ -1899,21 +1899,23 @@ void har_tick(object *obj) {
...
     // Make sure HAR doesn't walk through walls
     // TODO: Roof!
     vec2i pos = object_get_pos(obj);
-    int ab_flag = player_frame_isset(obj, TAG_AB);
-    if((h->state != STATE_DEFEAT && !ab_flag) || player_frame_isset(enemy_obj, TAG_CW)) {
+    if(!player_frame_isset(obj, TAG_AB)) {
+        int left_bound = ARENA_LEFT_WALL;
+        int right_bound = ARENA_RIGHT_WALL;
+        if(pos.y != ARENA_FLOOR) {
+            right_bound = ARENA_RIGHT_WALL - 1;
```
katajakasa, line 1915: Hum. Should we fix this "properly", so that both edges work similarly (for e.g. fair network play) ?

LegendaryBlueShirt: Joneirik and I were entertaining the idea of adding an advanced configuration option to toggle this

katajakasa: That might work. Still, I think the best default here would be to do it "right" -- I think this is a unexpected behaviour / balance problem otherwise :/

Vagabond:
Yeah, this one is weird because I can both see the appeal and the disadvantages

Vagabond (approved):

PR #1433, opened 2026-08-10
Archipelago Randomizer Support

stubobis1:


src/game/utils/settings.h
```diff
@@ -137,6 +145,9 @@ typedef struct {
     settings_tournament tournament;
     settings_keyboard keys;
     settings_network net;
+#if ARCHIPELAGO_ENABLED
+    settings_archipelago archipelago;
+#endif
```
Nopey, line 150: Do not use compile-time guards in settings.c/.h, as binaries built without ARCHIPELAGO_ENABLED will be unaware of these settings & silently remove them from the .conf file.

If we want binaries built without ARCHIPELAGO_ENABLED to be entirely archipelago-unaware, then its settings will have to go in a separate archipelago.conf file.

resources/archipelago/Useful 8bit.mp3
Nopey, line 1: Where are the files under resources/archipelago from?
Are they under a compatible license (e.g. MIT or CC0), or are they your own work?

src/archipelago/submodule/wswrap
Nopey, line 1: I don't like the prospect of adding submodules for dependencies, especially conditionally-enabled ones such as these.

If someone wants to check the various package managers we're using for our other deps (apt, brew, vcpkg) and see if these packages are all available (or, in the case of vcpkg, easily added via our vcpkg-ports folder), we could start looking at moving these deps into Dependencies.cmake.

CMakeLists.txt
```diff
@@ -29,6 +30,15 @@ endif()
 
 project(OpenOMF C)
 
+if(USE_ARCHIPELAGO)
+    enable_language(CXX)
+    set(CMAKE_CXX_STANDARD 17)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
Nopey, line 36: This seems unnecessary.
```suggestion
```

Dockerfile.build
Nopey, line 1: These Dockerfiles seem like build infrastructure that is entirely independent of the rest of the PR.
If you'd like to introduce these, I would suggest creating a separate PR for them.

src/game/gui/textinput.c
```diff
@@ -386,9 +386,16 @@ static void textinput_init(component *c, const gui_theme *theme) {
     }
     refresh(c);
     if(c->h_hint < 0) {
+        if(str_size(&ti->buf) == 0) {
+            str tmp;
+            str_from_c(&tmp, " ");
+            text_set_from_str(ti->text, &tmp);
+            str_free(&tmp);
```
Nopey, line 393: You don't need to construct a `str` here.
```suggestion
            text_set_from_c(ti->text, " ");
```

src/game/gui/textinput.c
Nopey, line 1: Could you explain or motivate the changes to this file?

Nopey (commented):
I haven't read the rest of the .c changes, but these changes in textinput.c stood out to me.

Nopey:
feel free to undraft when this is ready for review 👍

PR #1444, opened 2026-09-07
Use Integers For Damage Calculation

LegendaryBlueShirt:
Cherry picked 689940b7c1f31753f49455030c5c899003561773 from PR 1044


Vagabond:
Is there a better way to do this? wasn't there some fixed point macros/utility functions we had at one point?

LegendaryBlueShirt:
That doesn't apply here because damage/health can never be fractional.  I suspect this code change won't actually affect any damage values because since last year a lot of the damage calculation has already been moved to integer math. The fewer floats we have the better.

Vagabond:
But is it possible to do something cleaner than

```
*damage = *damage * leg_power / APPENDAGE_POWER_FIXP;
```

where APPENDAGE_POWER_FIXP is some random constant?

